PR #93, opened 2023-02-15
Add architecture docs

misaka:
This document is based on the arc42 (https://arc42.org/), which provides a template to providing architectural documentation. The goal of this approach is to be lightweight and is designed to work well with agile delivery.

The plan is to keep this documentation with the project and to use it as a basis of other architectural documentation, as required.


benilovj (approved):
Looks great for a first pass!

PR #119, opened 2023-02-28
Test POSTing a child vaccination with Playwright

tvararu:
WIP


tests/offline.spec.ts
```diff
@@ -1,11 +1,22 @@
...
+const swWaitForRequest = (context: BrowserContext, url: string) => {
+  return new Promise<Request>((resolve) => {
+    context.route(url, (route: Route) => {
+      if (route.request().serviceWorker()) {
+        resolve(route.request());
+      }
+      return route.fulfill();
+    });
+  });
+};
```
tvararu, line 18: The native `page.waitForRequest` doesn't listen to requests in the ServiceWorker, so I had to write my own wrapper.

Based on Playwright's documentation for `PW_EXPERIMENTAL_SERVICE_WORKER_NETWORK_EVENTS`: https://playwright.dev/docs/service-workers-experimental#network-events-and-routing

package.json
```diff
@@ -15,7 +15,7 @@
     "build:css": "sass ./app/assets/stylesheets/application.sass.scss:./app/assets/builds/application.css --no-source-map --load-path=node_modules --quiet-deps",
     "build": "esbuild app/javascript/*.[jt]s app/javascript/controllers/*.[jt]s --bundle --sourcemap --outdir=app/assets/builds --public-path=assets",
     "build:serviceworker": "esbuild app/javascript/serviceworker/main.js --bundle --sourcemap --outfile=public/sw.js",
-    "test:e2e": "playwright test",
+    "test:e2e": "PW_EXPERIMENTAL_SERVICE_WORKER_NETWORK_EVENTS=1 playwright test",
```
tvararu, line 18: This is necessary to listen to requests originating from the ServiceWorker, but it's an experimental flag: https://playwright.dev/docs/service-workers-experimental

tests/offline.spec.ts
```diff
@@ -26,8 +37,20 @@ test("Works offline", async ({ page }) => {
...
+  await expect(page.getByTestId("child-status").nth(0)).toContainText(
+    "Not yet"
+  );
 
   await page.getByTestId("child-link").nth(0).click();
   await expect(page.getByTestId("full-name")).toContainText("Aaron Pfeffer");
+
+  await page.getByTestId("confirm-button").click();
+
+  await goOnline();
```
misaka, line 49: At this point I'm not really familiar with this part of the system and I'm left wondering whether the "Vaccinated" field should be `Vaccinated` or `Not Yet`. Especially as the page gets reloaded further down. Would it make sense to add a test here to make that clear?

misaka: And by test, I mean an expectation. e.g.:

```
  await expect(page.getByTestId("child-status").nth(0)).toContainText(
    "Vaccinated"
  );
```

tvararu: It will be `Vaccinated`, shimmed by the client-side JS, but we don't implement that just yet; or the current implementation needs some updating. Can we add this as a follow-up? I'm expecting to wrap things up on the spiking and then card out stories to implement more of the prototype.

misaka (approved):

PR #120, opened 2023-03-01
Add and use /record-async route

tvararu:
This adds the CSRF protection back to the `/record` route and creates a new `/record-async` route. We still have to add another layer of auth to this, which is coming in a follow-up PR.

Review commit by commit.

### DIFF
--- a/app/javascript/serviceworker/record-route.js
+++ b/app/javascript/serviceworker/record-route.js
@@ -22,7 +22,9 @@ export const recordRouteHandler = async ({ request }) => {
     const campaignId = getCampaignIdFromURL(request.url);
     const campaignUrl = recordTemplateURL(campaignId);
 
-    saveRequest(clonedRequest);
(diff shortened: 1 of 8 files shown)


app/javascript/serviceworker/record-route.js
```diff
@@ -22,7 +22,9 @@ export const recordRouteHandler = async ({ request }) => {
     const campaignId = getCampaignIdFromURL(request.url);
     const campaignUrl = recordTemplateURL(campaignId);
 
-    saveRequest(clonedRequest);
+    const url = clonedRequest.url + "-async";
```
tvararu, line 25: End to end tests indirectly check that this is hooked up correctly, but though it's covered by unit tests, we don't assert against it. Should we?

tvararu:
This is a better implementation: https://github.com/nhsuk/record-childrens-vaccinations/pull/122

PR #133, opened 2023-03-09
Enable inline rendering for asciidoc

misaka:
Note: Before merging the URLs for the images need to be adjusted to point to the main branch. See below for why.

This enables inline rendering of Asciidoc files by using the `.adoc` extension and using plantuml.com's proxy renderer to render diagrams.

One downside of this approach is that the URLs for the diagram source files has to be encoded into the document. This means, for example, that for this PR I had to hardcode the name of the branch in the `architecture.adoc` file.  Before merging we'll need to change these to point to the main branch, and in the future to make any changes we'll have to do some jiggery pokery.


tvararu (approved):

misaka:
@tvararu updated this PR to render inline and also be dual-mode ... so it works both locally and remotely.

tvararu (approved):

PR #158, opened 2023-03-29
Refactor store methods

tvararu:
Rename them and implement a new `getByUrl` method.

Also rename the POST requests table to `delayedRequests`, and add a table that will be used for `cachedResponses`. The schema can currently be shared between the two.

This is more prep-work beore we can use IndexedDB instead of the cache API in the `NetworkFirst` strategy, followed by encrypting content as it goes in/out of the cache.

Review commit by commit.

### DIFF
--- a/app/javascript/serviceworker/store.ts
+++ b/app/javascript/serviceworker/store.ts
@@ -1,5 +1,7 @@
 import { openDB, DBSchema } from "idb";
 
+type StoreName = "delayedRequests";
+
 interface RequestObject {
   id?: number;
   url: string;
@@ -23,7 +25,7 @@ interface Db {
 const DB_NAME = "offline";
 const DB_VERSION = 1;
 
-const openTx = async (mode: "readwrite" | "readonly") => {
+const openTx = async (storeName: StoreName, mode: "readwrite" | "readonly") => {
   const db = await openDB<OfflineDatabase>(DB_NAME, DB_VERSION, {
     upgrade(db: Db) {
       db.createObjectStore("delayedRequests", {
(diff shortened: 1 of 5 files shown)


app/javascript/serviceworker/store.ts
```diff
@@ -1,58 +1,78 @@
...
-
-interface Db {
-  createObjectStore: (
-    arg0: string,
-    arg1: { keyPath: string; autoIncrement: boolean }
-  ) => void;
 }
 
 const DB_NAME = "offline";
 const DB_VERSION = 1;
```
misaka, line 25: If we're upgrading the DB should we also increment the version here?

misaka: Oh, I see, I was confused ... we aren't upgrading, we are just changing how we _would_ upgrade. 🤦

app/javascript/serviceworker/store.ts
```diff
@@ -1,58 +1,78 @@
...
+  return db.transaction(storeName, mode);
 };
 
-export const saveRequest = async (request: Request) => {
-  const requestData = await request.formData();
-  const data = Object.fromEntries(requestData);
-
-  const tx = await openTx("readwrite");
-  await tx.store.add({ url: request.url, data });
+export const add = async (storeName: StoreName, url: string, body: any) => {
```
misaka, line 49: It feels odd to me here, and for the funcs below, that the caller would need to know the name of the db being used ... can we have a chat to hash this out?

misaka: Works. Nice!

misaka (approved):
:shipit:

PR #164, opened 2023-04-05
Add offline docs

misaka:
Add documentation of how we implement offline working, with relevant security information.

### DIFF
--- a/docs/offline-support.adoc
+++ b/docs/offline-support.adoc
@@ -6,7 +6,7 @@ ifdef::env-github[]
 // When PRing changes to the diagrams you can change this attributes
 // temporarily to the name of the branch you're working on. But don't forget
 // to change it back to main before merging!!
-:github-branch: update-architecture-docs
+:github-branch: add-offline-docs
 
 :github-repo: nhsuk/record-childrens-vaccinations
 
--- a/docs/offline-support.adoc
+++ b/docs/offline-support.adoc
@@ -33,13 +33,16 @@ records requires a more advanced approach.
 
 == Feature overview
 
-The current design of the service expects the SAIS team to save a campaign for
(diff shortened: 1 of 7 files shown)


tvararu (approved):
Really clean. The attack scenarios section in particular is really well written. Good work!

docs/offline-support.adoc
```diff
@@ -0,0 +1,218 @@
...
+example in schools where there may be no mobile signal or WiFi that the SAIS
+team can use.
+
+To accomplish this the standard patterns provided by server-rendered apps are
+insufficient. Aggressive caching can help accommodate viewing data, but updating
+records requires a more advanced approach.
+
+== Feature overview
+
+The current design of the service expects the SAIS team to save a campaign for
```
benilovj, line 36: It might be good to explicitly mention here that the data would be stored encrypted.

docs/offline-support.adoc
```diff
@@ -0,0 +1,218 @@
...
+== Technical approach
+
+IMPORTANT: The design of this service feature is still being designed and tested
+           with users. The exact approach described here may change, e.g. the
+           user may need to manually initiate the synchronisation back to the
+           server, etc.
+
+=== Service Worker
+
+The service worker, a feature widely available in modern browsers, is used to
```
benilovj, line 53: > feature widely available in modern browsers

worth linking to https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorker#browser_compatibility here? I was wondering which browsers were supported so I went here myself...

benilovj (approved):
Apart from the couple of nits, looks great! Good work...

docs/offline-support.adoc
misaka: Good call. Added link.

PR #165, opened 2023-04-06
Use cache match and put in network strategies

tvararu:
Defer to our module instead of calling `caches.open` directly. This will allow us to change the caching mechanism behind the scenes while not having to update all the call sites.

Refactor some other things on the way there, see commits.

This is part of shifting from using the Cache API to IDB as the backing store, and will be done in roughly 3 parts:

0: This PR, prepwork
1: Adding `store` calls to the `put` and `addAll` methods in the cache module, to start storing things in `idb` but not yet reading them
2: Adding `store` calls to the `match` function, completely shifting the implementation from using the Cache API to IDB
3: Removing Cache API related code

Following this, we can move on to actually encrypting/decrypting requests.

### DIFF
--- a/app/javascript/serviceworker/cache.spec.ts
+++ b/app/javascript/serviceworker/cache.spec.ts
@@ -1,3 +1,4 @@
+require("jest-fetch-mock").enableMocks();
 import { addAll, match, put } from "./cache";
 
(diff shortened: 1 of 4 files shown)


misaka (approved):
LBTM! 👍

PR #167, opened 2023-04-06
Add requests to idb through NetworkFirst strategy

tvararu:
The NetworkFirst strategy will continue to use the CacheAPI mainly, but this adds initial support to additionally save the requests to IDB.

Few other refactors done on the way, see commits.

### DIFF
--- a/app/javascript/serviceworker/network.spec.ts
+++ b/app/javascript/serviceworker/network.spec.ts
@@ -1,24 +1,16 @@
 require("jest-fetch-mock").enableMocks();
 import { cacheOnly, networkFirst } from "./network";
+import { match, put } from "./cache";
 
-const mockCache = {
-  match: jest.fn(() => "test"),
-  put: jest.fn(),
-};
-
-const mockCaches = {
-  open: jest.fn().mockResolvedValue(mockCache),
-};
-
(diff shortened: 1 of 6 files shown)


app/javascript/serviceworker/network.spec.ts
```diff
@@ -1,24 +1,16 @@
...
-const mockCaches = {
-  open: jest.fn().mockResolvedValue(mockCache),
-};
-
-global.caches = mockCaches as any;
+jest.mock("./cache");
 
 describe("cacheOnly", () => {
   test("returns cached response if available", async () => {
+    (match as jest.Mock).mockReturnValueOnce("test");
```
misaka, line 9: Is there a way to cast "mock" at the top of the file, just once, instead of in each spec like this?

tvararu: Discussed offline; two options:

`jest.mocked` to create a separate variable https://jestjs.io/docs/mock-function-api#jestmockedsource-options

Or `jest.Mocked` to use as a custom type https://jestjs.io/docs/mock-function-api#jestmockedsource

Good to keep in mind, but not necessarily a lot less code in this case.

misaka (approved):

PR #174, opened 2023-04-12
Fetch cached requests from idb

tvararu:
This migrates the `cache` methods to use `idb` as the backing store instead of the Cache API.

There are a few bugs left over, but functionality is good enough to the point that tests pass. We can tackle these in follow-up PRs:

- [ ] Fix redirects, save status codes
- [ ] Fix regexp matching, registerRoute calls
- [ ] Serialise/deserialise responses properly
- [ ] Make key lookup a URL and list of headers

See commits.

### DIFF
--- a/app/javascript/serviceworker/cache.spec.ts
+++ b/app/javascript/serviceworker/cache.spec.ts
@@ -5,7 +5,7 @@ import { add, getByUrl } from "./store";
 jest.mock("./store");
 
 describe("addAll", () => {
-  test("works", async () => {
+  test("fetches and adds all the passed in requests", async () => {
     const urls = ["https://example.com/test", "https://example.com/test2"];
     const firstResponse = new Response("foo");
     const firstBlob = await firstResponse.clone().blob();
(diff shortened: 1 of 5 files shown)


app/javascript/serviceworker/cache.spec.ts
```diff
@@ -1,49 +1,57 @@
...
+      .mockResolvedValueOnce(secondResponse);
+    await addAll(urls);
+
+    expect(add).toHaveBeenCalledWith("cachedResponses", urls[0], firstBlob);
+    expect(add).toHaveBeenCalledWith("cachedResponses", urls[1], secondBlob);
   });
 });
 
 describe("match", () => {
   test("works", async () => {
```
misaka, line 26: If this were rspec, I would be pickier about the use of "works" here. Is now a good time to improve this?

app/javascript/serviceworker/cache.spec.ts
```diff
@@ -1,49 +1,57 @@
...
+  test("returns undefined if no match", async () => {
+    (getByUrl as jest.Mock).mockResolvedValueOnce(undefined);
+
+    const result = await match("test");
+    expect(result).toBeUndefined();
   });
+});
 
+describe("put", () => {
   test("caches to the store", async () => {
```
misaka, line 44: Similar to the above comment about spec naming, I'd consider this test to be testing two things ... one that it accepts a Request object and the other that it accepts a URL string.

tvararu: I've refactored some things and it isn't necessary to test that the function works with either argument anymore, pushed up a new commit.

app/javascript/serviceworker/cache.ts
```diff
@@ -1,28 +1,28 @@
-import { add } from "./store";
+import { add, getByUrl } from "./store";
 
-const cacheName = "offline-v1";
+const url = (request: RequestInfo): string => new Request(request).url;
```
misaka, line 3: Naming suggestion.

```suggestion
const urlOf = (request: RequestInfo): string => new Request(request).url;
```

misaka (approved):

PR #176, opened 2023-04-17
Create `secret` module

tvararu:
This defines a `secret` module that's a wrapper around the browser [SubtleCrypto API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto).

The test shows how to use it, but the idea is to take a user-provided passphrase and a payload and encrypt/decrypt that easily, without having to know about all the crypto gubbins.

A few libraries were explored before this was implemented but nothing suitable was found, as most of them are unfinished, or don't provide a simple enough API. Here's a list:

- https://github.com/copious-world/crypto-wraps (looks good but similarly isn't simple enough to use)
- https://github.com/digitalbazaar/forge (mature, but doesn't actually use SubtleCrypto)
- https://github.com/is-it-ayush/cryption
- https://github.com/AKASHAorg/easy-web-crypto
- https://github.com/jo/webcryptobox-js


tvararu:
Some more investigation into alternatives, some found via https://www.npmjs.com/search?q=keywords:subtlecrypto:

- https://github.com/174n/tiny-enc: This one actually is minimal and has a friendly API, but it's not widely used and is missing a `deriveKey` step after the `importKey` step
- https://github.com/junkurihara/jscu/tree/develop/packages/js-crypto-aes: More robust than tiny-enc, but also not in wide use
- https://github.com/ethereum/js-ethereum-cryptography: Battle hardened and audited, but doesn't use SubtleCrypto, it's a JS-native implementation of all the primitives. Would be a lot of extra code down the wire for only a few simple AES calls

Also @misaka dug up this reference demo from Mozilla: https://mdn.github.io/dom-examples/web-crypto/derive-key/index.html

Source code is at https://github.com/mdn/dom-examples/blob/main/web-crypto/derive-key/pbkdf2.js, their implementation is 99% identical to ours in how they import the key, derive it, and use it to encrypt/decrypt.

tvararu:
Pushed up an optional commit that adds an `.init()` method to memoise the key. This is optional, but I think it makes sense.

app/javascript/serviceworker/simple-crypto.ts
```diff
@@ -0,0 +1,119 @@
...
+      },
+      this.key,
+      encoder.encode(plaintext)
+    );
+
+    // Prepend the iv to the ciphertext, so that it can be retrieved later.
+    return self.btoa(
+      [...new Uint8Array(iv), ...new Uint8Array(ciphertext)]
+        .map((b) => String.fromCharCode(b))
+        .join("")
```
misaka, line 88: Trying to understand this bit ... is it concating the iv and the encrypted data, and then turning it into a string? Maybe worth a chat since you were also talking about how we'll get a blob in the request response (?) while here we need a string ...

misaka (approved):

PR #177, opened 2023-04-17
Disable browser caching for children.json

tvararu:
This prevents the browser from saving a request for this file to the Cache_Data folder.

This is just the bare minimum for now, but we'll probably expand this to other files or perhaps create a separate controller to inherit from for sensitive endpoints and data.

There might be downsides to setting aggressive no-cache headers; it potentially breaks the back button in situations of flaky connectivity.

### DIFF
--- a/app/controllers/vaccinations_controller.rb
+++ b/app/controllers/vaccinations_controller.rb
@@ -4,6 +4,7 @@ class VaccinationsController < ApplicationController
   before_action :set_children, only: %i[index record_template]
 
   def index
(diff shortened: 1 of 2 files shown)


misaka:
Would it be worth having some behaviour tests around this?

tvararu:
@misaka added a test to this. Ended up doing it in Rails because it isn't easy to intercept requests and check their response headers in PlayWright when you're using a ServiceWorker in the middle...

misaka (approved):
:shipit:

PR #190, opened 2023-04-25
Set all pages to Cache-Control no-store

tvararu:
Without this, the browser can choose to temporarily store pages in the user's navigation in a client-side plaintext cache. Because our pages can contain things like NHS numbers or user data, this is a potential security risk.

There's no tests in this commit because `spec/requests/pages_spec.rb` covered testing the `/children.json` endpoint, and it should still pass.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,6 @@
 class ApplicationController < ActionController::Base
+  before_action :set_disable_cache_headers
+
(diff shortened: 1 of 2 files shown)


misaka (approved):

misaka:
I wanted to say "Tests!" but ... ¯\_(ツ)_/¯ ... we still have the test on the index, right?

PR #198, opened 2023-05-03
Ask the user for an offline password

tvararu:


app/javascript/serviceworker/messages.spec.js
```diff
@@ -31,16 +31,32 @@ describe("messageHandler", () => {
...
+      expect(initCache).toHaveBeenCalledWith("test");
       expect(addAll.mock.calls[0][0]).toMatchInlineSnapshot(`
         [
-          "/campaigns/undefined/children",
-          "/campaigns/undefined/children.json",
-          "/campaigns/undefined/children/record-template",
-          "/campaigns/undefined/children/show-template",
+          "/assets/application.js",
+          "/favicon.ico",
+          "/dashboard",
```
misaka, line 51: This part concerns me a little ... I think the access codes are per-campaign. It's possible a SAIS nurse will be working on more than one offline campaign at a time, esp if they're part of a big team. I think for security purposes, seeing as each campaign is a different data set we should have a different access code for each campaign. 

That being the case, I don't think assets that are used across assets should be encrypted. For example, which password would they enter to view `/dashboard`?

In any case, probably for another PR, I'll create an issue.

app/views/campaigns/show.html.erb
```diff
@@ -20,14 +20,7 @@
...
 
-      <form method="post">
-        <button class="nhsuk-button nhsuk-u-margin-bottom-3"
-                type="submit"
-                data-action="offline-campaign#saveOffline:prevent"
-                data-testid="save-offline">
-          Save offline
-        </button>
-      </form>
+      <a href="<%= setup_offline_path %>" class="nhsuk-button nhsuk-u-margin-bottom-3" data-testid="save-offline">Save offline</a>
```
misaka, line 23: With the action connection to `offline-campaign#saveOffline` removed, is there any functionality still being used from `offline-campaign` Stimulus controller? I get the impression that `offline-campaign` is superceded by the `setup-offline` controller, so we can get rid of that link-up here and the Stimulus controller.

app/javascript/serviceworker/messages.js
```diff
@@ -10,15 +10,24 @@ const messageHandlers = {
...
+    const campaignId = event.data.payload["campaignId"];
+    const password = event.data.payload["password"];
 
-    addAll([
+    await initCache(password);
+
+    await addAll([
+      ...event.data.payload["additionalItems"],
+      `/favicon.ico`,
+      `/dashboard`,
```
misaka, line 22: This part concerns me a little ... I think the access codes are per-campaign. It's possible a SAIS nurse will be working on more than one offline campaign at a time, esp if they're part of a big team. I think for security purposes, seeing as each campaign is a different data set we should have a different access code for each campaign.

That being the case, I don't think assets that are used across assets should be encrypted. For example, which password would they enter to view /dashboard?

In any case, probably for another PR, I'll create an issue.

app/javascript/controllers/setup_offline_controller.ts
```diff
@@ -0,0 +1,36 @@
...
+    await wb.messageSW({
+      type: "SAVE_CAMPAIGN_FOR_OFFLINE",
+      payload: {
+        password: this.passwordTarget.value,
+        campaignId: 1,
+        additionalItems: [css, js],
+      },
+    });
+
+    this.formTarget.submit();
```
misaka, line 34: We're submitting the password after we initialise the cache with it, and download all the assets. If a password validation fails, on one hand it shouldn't matter much since we'll, in theory, download the data again when they try again to provide a proper password, BUT ... if they decide not to try the password again, for whatever reason, we're in a weird state. Personal data is downloaded to this device with a potentially weak password, which also won't be recoverable since the server has rejected it.

I think we should be more cautious with this. We could roll a fix forward, but we could also forget in which case we'll be susceptible to weird bugs, at the very least.

misaka: In it's current incarnation, if the user submits an empty password it caches the data using that as an encryption key.

misaka (changes requested):
I feel that a couple issues should be addressed, like removing the old `offline-campaign` Stimulus controller and ensuring we don't download data before we know the password is secure. In addition, I think we should create issues for any cans we've kicked down the road, or find some way to document them somewhere. The ones I can think of:

- Only encrypt the campaign-specific data. ([see comment](https://github.com/nhsuk/record-childrens-vaccinations/pull/198/files#r1192369275))
- Update the key derivation, mainly just how we get the salt
- Also, I propose we rename "passwords" to "access codes", or whatever the new term for these things are, at least in the code. Otherwise we risk mixing this up with "user passwords", which I've already done a few times

Would it be worth merging the Rails-side stuff separately? It's a smaller bit of work but I don't think there are issues there that need to address immediately, so could help to make the PR smaller.

PR #199, opened 2023-05-05
Add ADR about Offline security

tvararu:
Details the architecture that we're using to maintain security for patient data.


adr/00005-offline-security.md
```diff
@@ -0,0 +1,40 @@
+# 5. Offline security
+
+Date: 2023-05-04
+
+## Status
+
+Accepted
```
misaka, line 7: Would it be worth saying that this is in-progress as we're still deciding on the exact mechanism for the encryption key? Or is that an addendum, or follow-up ADR?

tvararu: We totally can do this.

misaka (approved):
🚀

PR #200, opened 2023-05-05
Add wait to offline test

tvararu:
When clicking "Save offline," 5 or so requests are queued from the server. If we immediately switch the server to offline, those requests might still be in flight. We could use the `swWaitForRequest` function with a loop of all the possible requests to wait for them, or we could just add a 100ms timeout, which seems simpler.

This is a temporary measure to fix flaky tests on the `main` branch and will be replaced by a different offline e2e test, once we switch to the password-based implementation.

### DIFF
--- a/tests/offline.spec.ts
+++ b/tests/offline.spec.ts
@@ -34,6 +34,9 @@ test("Works offline", async ({ page, context }) => {
 
   await page.getByTestId("save-offline").click();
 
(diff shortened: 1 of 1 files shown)


tvararu:
On second thought, it looks like all the PRs are passing even without this. Closing it as it seems unnecessary, and will be replaced by https://github.com/nhsuk/record-childrens-vaccinations/pull/198.

PR #203, opened 2023-05-05
Disable implicit offline caching

tvararu:
Prerequisite for: https://github.com/nhsuk/record-childrens-vaccinations/pull/198

Depends on https://github.com/nhsuk/record-childrens-vaccinations/pull/200 (roughly... it might be mergeable without, but because this piles on additional requests in the `addAll`, the test just becomes even flakier without it).

The key bit in this PR is removing the `await put(request, networkResponse.clone());` call in the `NetworkFirst` strategy.

By doing this, pages are no longer cached implicitly, but instead have to be cached *explicitly*, via the "Save offline" button. The button is also updated to cache the JS and CSS alongside some other pages.

There's also a random "annotate" commit thrown in, I didn't want to split it off as its own PR, but it's harmless.

Disabling implicit offline caching is a necessary prerequisite to switching to using a user-provided password as the encryption key for the offline cache. We can't cache things before the user gives us that key.

### DIFF
--- a/app/javascript/controllers/offline_campaign_controller.js
+++ b/app/javascript/controllers/offline_campaign_controller.js
@@ -6,9 +6,14 @@ export default class extends Controller {
   connect() {}
 
   saveOffline() {
+    // Pass in paths to CSS and JS to save for offline use, because document
+    // methods are not available in serviceWorker
+    const css = document.querySelector("link[rel=stylesheet]").href;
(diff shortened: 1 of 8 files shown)


misaka:
#200 closed. Was that the wrong PR or is it irrelevant?

tvararu:
Yes, it's irrelevant in the end. The tests seem to pass without that and it should be replaced soon...

PR #205, opened 2023-05-05
Fix "Vaccinated" tags when offline

tvararu:
These were broken a while ago. When a user would vaccinate a patient while offline, they wouldn't see the "Vaccinated" tag update on their client, which is jarring.

This fixes them by populating the content of the tag by querying the content of the `delayedRequests` offline database and finding matching requests.


tvararu:
As we're imminently redesigning this page, I'm going to close this for now.

PR #211, opened 2023-05-17
Terraform

misaka:
Early implementation of terraform scripts to deploy to AWS.

- Uses [terraform-aws-module](https://github.com/terraform-aws-modules) as building blocks to reduce the code we need.
- Uses AWS Fargate with ECS, which means we provide a Docker container and Fargate manages running it for us, including starting/restarting, auto scaling, etc.
- AWS Fargate, however, does not manage the VBC, LB or DB for us. These are setup in Terraform.
- Uses AWS Secret Manager to access secrets required for deployment such as DB credentials, support user credentials, Rails master key, etc.
- DB is setup using AWS RDS.

Currently the terraform scripts create a working setup, but some things remain to be done ... some for this PR and some as cans to be kicked 👞🥫:

- 🥫 Remove ECS containers from public subnet, if possible. These only need to be reachable via the LB.
- 🥫 Ensure DB subnet is only reachable by our private subnets.
- 🥫 Check we're encrypting the DB and how the key is managed. Is it useful and easy to manage the key ourselves?
- 🥫 Setup an HTTPS endpoint.
- 🥫 Save tfstate to Terraform cloud, if possible.
- 🥫 Add instructions to project RAEDME and/or Terraform RAEDME.
- 🥫 Change how the DB hostname is discovered for use by the `DATABASE_URL` env var. Currently it relies on existing state, which means the DB has to be created before that env var can be set. In a new environment the steps are:
  1. disable setting the `DATABASE_ENV` env var
  2. run the scripts so that the db is created
  3. re-enable `DATABASE_ENV`
  4. run scripts again to set the env var on the ECS task
- 🥫 Split out `main.tf` to separate files.


docs/deploying.md
```diff
@@ -0,0 +1,95 @@
...
+### Pushing Docker image to AWS ECR
+
+We use a public AWS Elastic Container Registry for our Docker images.
+
+```sh
+# Use AWS to authenticate with Docker. You'll need the aws cli tools installed
+# (`awscli` in Homebrew) and to be logged into the AWS account that owns the
+# Elastic Container Repository (ECR). All public repositories are in us-west-1
+# region.
+aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/z8i3v8n4
```
tvararu, line 65: `us-east-1`?

misaka: Ya. Oops, I should've said `us-east-1` on line 63.

terraform/main.tf
```diff
@@ -0,0 +1,284 @@
...
+module "db" {
+  source  = "terraform-aws-modules/rds/aws"
+
+  identifier = "${local.abbreviated_app_name}-${local.env}-db"
+  apply_immediately = true
+
+  engine            = "postgres"
+  engine_version    = "13.7"
+  family            = "postgres13"
+  instance_class    = "db.t3.micro"
```
tvararu, line 99: Will be interesting to see if we need to revisit this for prod, but I assume even a micro is plenty? Struggling to find exact specs.

terraform/main.tf
```diff
@@ -0,0 +1,284 @@
...
+      }
+    }
+  }
+
+  tags = local.tags
+}
+
+module "alb_sg" {
+  source  = "terraform-aws-modules/security-group/aws"
+  # version = "~> 4.0"
```
tvararu, line 234: What was this for?

terraform/main.tf
```diff
@@ -0,0 +1,284 @@
...
+
+  egress_rules       = ["all-all"]
+  egress_cidr_blocks = module.vpc.private_subnets_cidr_blocks
+
+  tags = local.tags
+}
+
+module "alb" {
+  source  = "terraform-aws-modules/alb/aws"
+  # version = "~> 8.0"
```
tvararu, line 251: Is this still needed? Same as above

tvararu (approved):
LGTM!

I know this builds and runs, but did we run a `terraform validate` on it?

Perhaps we can add that to `bin/lint` in the future and create a separate `terraform.yml` GitHub action to check it.

PR #221, opened 2023-05-31
Remove URL from staging database config

tvararu:
This isn't necessary and breaks when running `bin/setup` locally unless DATABASE_URL is defined.


misaka:
When would this be necessary? Any deployed environment should set `DATABASE_URL`, and the database will never be on `localhost`, so I'm not sure where this would make sense.

misaka (approved):

PR #222, opened 2023-06-01
Remove URL from config/database.yml

tvararu:
This was a fix to run the app locally on Linux when the Postgres database is installed via asdf. However, it breaks on macOS and potentially other setups.


README.md
```diff
@@ -58,6 +58,25 @@ will use:
...
+psql -U postgres
+> CREATE ROLE your-user-here LOGIN SUPERUSER;
+```
+
+When running PostgreSQL via `pg_ctl start`, specify the socket:
+
+```sh
+pg_ctl start -o "-k /run/postgresql"
+```
+
```
misaka, line 79: Would it be worth modifying `bin/db` to do this? It'd probably require an OS check, etc.

tvararu: Let's not since I'm not sure everyone would choose to use `asdf` for Postgres anyway. :thinking:

PR #232, opened 2023-06-07
Add triage list

misaka:
We rename `children` to `patients` everywhere but in the vaccinations controllers and the JavaScript that uses it, as the changes in those files will be more significant.

### DIFF
--- a/spec/factories/triage.rb
+++ b/spec/factories/triage.rb
@@ -1,0 +1,23 @@
+# == Schema Information
+#
+# Table name: triage
+#
+#  id          :bigint           not null, primary key
+#  notes       :text
+#  status      :integer
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#  campaign_id :bigint
+#  patient_id  :bigint
+#
+# Indexes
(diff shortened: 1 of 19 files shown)


spec/factories/triage.rb
```diff
@@ -0,0 +1,25 @@
...
+# Indexes
+#
+#  index_triage_on_campaign_id  (campaign_id)
+#  index_triage_on_patient_id   (patient_id)
+#
+FactoryBot.define do
+  factory :triage do
+    status { "Ready for session" }
+    notes { nil }
+    # patient { Patient.first || association(:patient) }
```
benilovj, line 22: maybe worth getting rid

benilovj (approved):

spec/factories/triage.rb
misaka: It's a goner

misaka:
Here's how it looks:

image

And if we stub in the left side of the grid row where the filter would be, just to illustrate spacing:

image

benilovj (approved):

PR #236, opened 2023-06-09
Match design of sessions#index to the prototype

tvararu:
Remove the card wrapper around the table and set some CSS.

### Screenshots

image


misaka (approved):
🚀

PR #238, opened 2023-06-09
Create PatientSession model

tvararu:
We'll need to record multiple vaccination outcomes per patient + session. To model this, we've decided to have a one-to-many relationship between patient_sessions and vaccination_records.

To create that relationship, Rails wants the PatientSession to be its own join model, rather than just a join table. We also have to explicitly add `has_many :patient_sessions` to the patient and session models.

This change does not affect the controller or view code at all, as its just plumbing to update the table name and to associate the models correctly.

Some enum/validations have been shuffled around, as I prefer putting all the relationships at the top of a model definition.

### Data model

image

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -21,11 +21,9 @@
 #  index_patients_on_nhs_number  (nhs_number) UNIQUE
 #
 class Patient < ApplicationRecord
-  enum :sex, %w[Female Male]
-  enum :gp, ["Local GP"]
-  enum :screening, ["Approved for vaccination"]
-  enum :consent, ["Parental consent (digital)"]
-  enum :seen, ["Not yet", "Vaccinated"]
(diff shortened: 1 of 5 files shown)


spec/factories/vaccination_records.rb
```diff
@@ -0,0 +1,24 @@
...
+#  index_vaccination_records_on_patient_session_id  (patient_session_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (patient_session_id => patient_sessions.id)
+#
+FactoryBot.define do
+  factory :vaccination_record do
+    patient_session { nil }
+    administered_at { "2023-06-09" }
```
misaka, line 22: I expect in the (near) future we may want to generate these when we create a session, maybe with a cascading approach. This works for now.

misaka (approved):

PR #245, opened 2023-06-13
Update tags in vaccinations index page

tvararu:
This implements the outcome tags from the prototype, as well as adding some glue code to allow us to index on the `.cohort` (née `.patient_sessions`). This allows us to get the vaccination records.

We're still using the `patient.seen` to populate the outcome tag, but that will change in the next PR.

Also includes the beginning of a `tests/vaccination.spec.ts`.

Review commit by commit.

### After

image


app/controllers/vaccinations_controller.rb
```diff
@@ -73,7 +73,11 @@ def set_patient
     @patient = Patient.find(params[:id])
   end
 
-  def set_patients
-    @patients = @session.patients.order(:first_name, :last_name)
+  def set_cohort
+    @cohort =
```
misaka, line 77: Can we call this `patient_sessions` instead of `cohort`?

misaka (approved):

PR #256, opened 2023-06-21
Update status tag component

misaka:
Refactoring of the status tag component to be less use-case specific. This was done to make it easier to add pages like the triage index.


app/helpers/triage_helper.rb
tvararu, line 1: Nice

tvararu (approved):
Really nice. Test is a lot shorter too!

PR #258, opened 2023-06-22
Update child details card on vaccinations show

tvararu:
Some fields are missing because they aren't part of the data model or are pulled in from consent. We can add them later.

See commits.

image

### DIFF
--- a/app/assets/stylesheets/application.sass.scss
+++ b/app/assets/stylesheets/application.sass.scss
@@ -16,3 +16,15 @@ $color_app-pale-blue: #ccdff1;
 .app-u-frutiger * {
   font-family: "Frutiger W01", Arial, sans-serif !important;
 }
+
+.app-summary-list--no-bottom-border {
+  .govuk-summary-list,
+  .nhsuk-summary-list {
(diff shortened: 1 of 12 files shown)


app/assets/stylesheets/application.sass.scss
misaka, line 1: 👍

app/javascript/controllers/child_vaccination_controller.js
```diff
@@ -27,12 +27,9 @@ export default class extends Controller {
     const child = json[childId];
 
     this.fullNameTarget.textContent = child["full_name"];
-    this.sexTarget.textContent = child["sex"];
     this.dobTarget.textContent = child["dob"];
-    this.consentTarget.textContent = child["consent"];
     this.gpTarget.textContent = child["gp"];
     this.nhsNumberTarget.textContent = child["nhs_number"];
-    this.screeningTarget.textContent = child["screening"];
```
misaka, line 35: Interesting. I was removing fields also in the testing env PR, we should line these up.

misaka (approved):
Looking good, once we sort out that test.

PR #259, opened 2023-06-23
Tests in test env

misaka:
This changes the e2e tests to use the test Rails env, isolating the tests from any dev data we have. It also loads up a smaller example campaign specific to testing in order to try to make it easier to work with.


misaka:
Decided we won't pull in the example test campaign into the e2e tests to base expectations on ... at least, not yet.

tvararu (approved):

PR #262, opened 2023-06-26
Update vaccinations show page

tvararu:
See commits.

### After

image


app/controllers/triage_controller.rb
misaka, line 1: 🙏

misaka (approved):
🚀

PR #263, opened 2023-06-26
Change 'Patient#gp' field from an enum to a string

benilovj:
This allows something other than 'Local GP' to be recorded in it.

### DIFF
--- a/app/avo/resources/patient_resource.rb
+++ b/app/avo/resources/patient_resource.rb
@@ -14,7 +14,7 @@ class PatientResource < Avo::BaseResource
   field :dob, as: :date
   field :nhs_number, as: :number
   field :sex, as: :select, enum: ::Patient.sexes
-  field :gp, as: :select, enum: ::Patient.gps
+  field :gp, as: :text
   field :screening, as: :select, enum: ::Patient.screenings
   field :consent, as: :select, enum: ::Patient.consents
   field :seen, as: :select, enum: ::Patient.seens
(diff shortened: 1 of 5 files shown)


misaka:
There is GP information in the consent response, which is structured a little differently but more in line with what the consent forms look like.

https://github.com/nhsuk/record-childrens-vaccinations/blob/main/db/schema.rb#L43-L44
```ruby
# schema.rb
    t.integer "gp_response"
    t.text "gp_name"
```

https://github.com/nhsuk/record-childrens-vaccinations/blob/main/app/models/consent_response.rb#L57
```ruby
# consent_response.rb
  enum :gp_response, %w[yes no dont_know]
```

It's not clear (to me) where GP info should sit since in theory there could be multiple consent responses with different info (e.g. separated parents, potentially with one version being erroneous?).

tvararu:
I think this GP field on the patient shouldn't be used anymore. Can we try to remove it for now and model GP on the consent_response as Misha points out?

misaka:
> I think this GP field on the patient shouldn't be used anymore. Can we try to remove it for now and model GP on the consent_response as Misha points out?

I think at the moment in the Child Details component we pull GP info out of that field, right? We need to know whether that's meant to use the most recent consent value, or if it should be removed.

benilovj:
This appears to be a can of worms... I thought it was just a quick fix, but it's not a priority right now. The patient record will look a tiny bit strange because each will have "Local GP" on each one, but that's not a big problem. Closing for now.

PR #269, opened 2023-06-26
Add triage form

misaka:
Adds triage notes to patient triage page.

Note the change to the routes table (commit: [Nest triage route under sessions/patients](https://github.com/nhsuk/record-childrens-vaccinations/commit/ec6322c5d56efaaeaadd45eaf4369637fdc41571)). This change may be controversial and worth a discussion. If accepted it'll mean changes to the vaccinations routes (probably).

image

### DIFF
--- a/app/controllers/triage_controller.rb
+++ b/app/controllers/triage_controller.rb
@@ -33,7 +33,7 @@ class TriageController < ApplicationController
   end
 
   def set_triage
-    @triage = @patient.triage_for_campaign(@session.campaign)
+    @triage = @patient.triage_for_campaign(@session.campaign) || Triage.new
   end
 
   def set_consent_response
--- a/app/controllers/triage_controller.rb
+++ b/app/controllers/triage_controller.rb
@@ -1,6 +1,6 @@
 class TriageController < ApplicationController
-  before_action :set_session, only: %i[index show]
-  before_action :set_patient, only: [:show]
+  before_action :set_session, only: %i[index show create]
+  before_action :set_patient, only: %i[show create]
   before_action :set_triage, only: [:show]
   before_action :set_consent_response, only: [:show]
 
@@ -22,6 +22,12 @@ class TriageController < ApplicationController
   def show
   end
 
+  def create
+    @triage = Triage.new(campaign: @session.campaign)
+    @triage.update!(triage_params)
+    redirect_to session_triage_index_path(@session)
+  end
+
   private
 
   def set_session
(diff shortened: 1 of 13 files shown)


app/views/triage/_patient_triage_card.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="nhsuk-card">
+  <div class="nhsuk-card__content">
+    <h2 class="nhsuk-card__heading nhsuk-heading-m">
+      Triage
+    </h2>
+
+    <%= form_for [session, triage], builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
```
tvararu: Do you need to specify the builder? It should be the default, I didn't have to on my branch...

app/controllers/triage_controller.rb
```diff
@@ -21,6 +21,12 @@ def index
   def show
   end
 
+  def create
+    @triage = Triage.new(campaign: @session.campaign)
+    @triage.update!(triage_params)
+    redirect_to session_triage_index_path(@session)
```
tvararu: Should we handle potential errors here at this stage, or add it in later?

app/views/triage/_patient_triage_card.html.erb
```diff
@@ -5,6 +5,7 @@
     </h2>
 
     <%= form_for [session, triage], builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= hidden_field(:triage, :patient_id) %>
```
tvararu: Does `f.hidden_field` work, is it preferable?

tvararu (approved):

app/views/triage/_patient_triage_card.html.erb
misaka: I got rid of this in the last change to accommodate update, found a way around it.

app/views/triage/_patient_triage_card.html.erb
misaka: Nope, you're right I don't. Removed.

app/controllers/triage_controller.rb
misaka: You're right. Have you got an example of how to do this to hand?

app/controllers/triage_controller.rb
```diff
@@ -24,14 +24,20 @@ def show
 
   def create
     @triage = Triage.new(campaign: @session.campaign, patient: @patient)
-    @triage.update!(triage_params)
-    redirect_to triage_session_path(@session)
+    if @triage.update(triage_params)
+      redirect_to triage_session_path(@session)
+    else
+      render :show
```
tvararu, line 33: We should add `status: :unprocessable_entity` here

app/controllers/triage_controller.rb
```diff
@@ -24,14 +24,20 @@ def show
...
   end
 
   def update
     @triage = @patient.triage_for_campaign(@session.campaign)
-    @triage.update!(triage_params)
-    redirect_to triage_session_path(@session)
+    if @triage.update(triage_params)
+      redirect_to triage_session_path(@session)
+    else
+      render :show
```
tvararu, line 42: And here

app/controllers/triage_controller.rb
misaka: Done.

tvararu (approved):

PR #273, opened 2023-06-27
Show empty state when a session has no vaccinations to record

benilovj:
## What's changed

1. Made the "record vaccinations" index page two-thirds width, to bring closer to the designs
2. Introduced an empty state when there are no vaccinations to record. This will come in handy when tabs are introduced.

image

## Questions for reviewers

1. There's some inline CSS – should I extract it into the CSS files? If so, what could I name the class?
2. Does the proposed microcopy for the empty state make sense?

### DIFF
--- a/app/assets/stylesheets/_empty-list.scss
+++ b/app/assets/stylesheets/_empty-list.scss
@@ -1,0 +1,5 @@
+.app-empty-list {
+  margin-top: nhsuk-spacing(6);
+  border: 2px solid $nhsuk-border-color;
+  min-height: 300px;
+}
--- a/app/views/layouts/_empty_list.html.erb
+++ b/app/views/layouts/_empty_list.html.erb
@@ -1,4 +1,4 @@
-<div style="margin-top: 32px; border: 2px solid #b1b4b6; min-height: 300px">
+<div class="app-empty-list">
   <div class="nhsuk-card__content">
     <h2 class="nhsuk-heading-l">No results</h2>
     <p><%= message %></p>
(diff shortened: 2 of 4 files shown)


app/views/layouts/_empty_list.html.erb
```diff
@@ -0,0 +1,6 @@
+<div style="margin-top: 32px; border: 2px solid #b1b4b6; min-height: 300px">
```
tvararu, line 1: Could be `.app-empty-list`. Also instead of 32px, you can use the nhsuk-spacing helper

benilovj: @tvararu I've made the suggested changes in [a328455](https://github.com/nhsuk/record-childrens-vaccinations/pull/273/commits/a328455b389d8051ffb4072351a3a5fc808d3017)

app/assets/stylesheets/_empty-list.scss
```diff
@@ -0,0 +1,5 @@
+.app-empty-list {
+  margin-top: nhsuk-spacing(6);
+  border: 2px solid #b1b4b6;
```
tvararu, line 3: Does this look ok? Would be more consistent with the NHSUK Design System.

```suggestion
  border: 2px solid $nhsuk-border-color;
```

tvararu (approved):

app/assets/stylesheets/_empty-list.scss
benilovj: 🤷 but I've changed to the design system default until we have a good reason to deviate...

PR #279, opened 2023-06-28
Record HPV vaccination outcome

tvararu:
This wires up the check your answers page and successfully records vaccinations using the `VaccinationRecord` model.

It adds new fields and necessary validations.

Missing from this is the Notes feature, but this PR is already big enough. The field is currently a no-op.

#### Vaccinations index; new "Could not vaccinate" state now working

image

#### Validations

image
image

#### Site now populated on confirm page

image


app/controllers/vaccinations_controller.rb
```diff
@@ -86,4 +98,13 @@ def set_patient_outcomes
...
         .map { |ps| [ps.patient, ps.outcome] }
   end
+
+  def set_vaccination_record
+    @vaccination_record =
+      @patient
+        .patient_sessions
+        .find_by_session_id(@session.id)
+        .vaccination_records
+        .find_or_initialize_by(administered_at: nil)
```
misaka: What do yo think about adding this to `Patient`?

```ruby
  def vaccination_records_for_session(session)
    patient_sessions.find_by_session_id(@session.id).vaccination_records
  end
```

This code would then become:

```suggestion
    @vaccination_record =
      @patient
        .vaccination_records_for_session(@session)
        .find_or_initialize_by(administered_at: nil)
```

app/controllers/vaccinations_controller.rb
```diff
@@ -86,4 +98,13 @@ def set_patient_outcomes
...
         .map { |ps| [ps.patient, ps.outcome] }
   end
+
+  def set_vaccination_record
+    @vaccination_record =
+      @patient
+        .patient_sessions
+        .find_by_session_id(@session.id)
+        .vaccination_records
+        .find_or_initialize_by(recorded_at: nil)
```
misaka, line 108: What do yo think about adding this to `Patient`?

```ruby
  def vaccination_records_for_session(session)
    patient_sessions.find_by_session_id(@session.id).vaccination_records
  end
```

This code would then become:

```suggestion
    @vaccination_record =
      @patient
        .vaccination_records_for_session(@session)
        .find_or_initialize_by(recorded_at: nil)
```

app/models/vaccination_record.rb
```diff
@@ -18,4 +20,18 @@
...
+
+  validates :administered, inclusion: [true, false]
+  validates :site,
+            presence: true,
+            inclusion: {
+              in: sites.keys
+            },
+            if: -> { administered }
+
+  def self.site_options
```
misaka, line 34: I'd be tempted to put this into a helper, since I think this is only useful in for views.

misaka (approved):
Looks good! Couple of minor suggestions, but shippable as-is.

PR #281, opened 2023-06-28
Add health questions card

misaka:
image

### DIFF
--- a/tests/example_data.ts
+++ b/tests/example_data.ts
@@ -1,0 +1,22 @@
+// Read example campaign file and extract patients by name
+
+import * as fs from "fs";
+import * as path from "path";
+
+const example_campaign_path = path.join(
+  __dirname,
+  "../db/sample_data/example-test-campaign.json"
+);
+const example_campaign = JSON.parse(
+  fs.readFileSync(example_campaign_path, "utf8")
+);
+
+export function example_patient(name: string): Record<string, any> | null {
+  for (const patient of example_campaign.patients) {
+    const fullName = patient.firstName + " " + patient.lastName;
+    if (fullName === name) {
+      return patient;
+    }
+  }
+  return null;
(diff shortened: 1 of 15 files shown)


app/components/app_patient_medical_history_card_component.html.erb
```diff
@@ -0,0 +1,26 @@
+<% if has_medical_history? %>
```
tvararu, line 1: Should the component be the judge of whether it should show or not? I'm thinking this could live at the call-site.

tests/example_data.ts
```diff
@@ -0,0 +1,22 @@
...
+  "../db/sample_data/example-test-campaign.json"
+);
+const example_campaign = JSON.parse(
+  fs.readFileSync(example_campaign_path, "utf8")
+);
+
+export function example_patient(name: string): Record<string, any> | null {
+  for (let patient of example_campaign["patients"]) {
+    let fullName = patient["firstName"] + " " + patient["lastName"];
+    if (fullName == name) {
```
tvararu, line 17: Optional nit:

```suggestion
  for (const patient of example_campaign["patients"]) {
    const fullName = patient.firstName + " " + patient.lastName;
    if (fullName === name) {
```

tvararu (approved):
Nice component! :ship:

app/components/app_patient_medical_history_card_component.html.erb
misaka: Ya, I think I was thinking to accommodate displaying an alternative if the health questions aren't there, but that would be better placed with the caller. Will change.

tvararu (approved):

misaka:
@tvararu Updated the triage show page to match the latest designs.

tvararu (approved):

PR #285, opened 2023-06-29
Add ERD diagram generation

misaka:
image


Gemfile
```diff
@@ -36,6 +36,7 @@ group :development do
   gem "asciidoctor-diagram"
   gem "dockerfile-rails", ">= 1.0.0"
   gem "prettier_print", require: false
+  gem "rails-erd"
```
tvararu, line 39: Any config to get it to automatically run on `db:migrate` or similar?

Can you/GPT add a little section to the Readme with `bundle exec erd` command?

tvararu (approved):

Gemfile
misaka: Oops, I missed something in this commit! There's a rake task that takes care of that.

misaka: Added. This updates the diagram when `db:migrate` is run. Looking at it, I'm not sure how it works exactly, but we can roll-forward any fixes.

PR #286, opened 2023-06-30
Quick snags

tvararu:
See commits!


benilovj (approved):
Looks good.

Optionally, I'd go as far as dropping the `Patient#gp` field completely.

benilovj (approved):

PR #290, opened 2023-06-30
Display an empty state when there is nothing to triage.

benilovj:
image

### DIFF
--- a/app/controllers/triage_controller.rb
+++ b/app/controllers/triage_controller.rb
@@ -4,7 +4,7 @@ class TriageController < ApplicationController
   before_action :set_triage, only: %i[show]
   before_action :set_consent_response, only: %i[show]
 
-  layout "two_thirds", except: %i[index]
+  layout "two_thirds"
 
   def index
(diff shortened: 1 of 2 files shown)


app/controllers/triage_controller.rb
```diff
@@ -4,7 +4,7 @@ class TriageController < ApplicationController
   before_action :set_triage, only: %i[show]
   before_action :set_consent_response, only: %i[show]
 
-  layout "two_thirds", except: %i[index]
```
misaka, line 7: @theo explicitly said the index page shouldn't be in this layout. I don't recall the reasoning why though.

Ship it and sort it out later?

misaka (approved):

app/controllers/triage_controller.rb
benilovj: I'm changing the layout of the index immediately after this PR, so this is deliberate.

The new designs make the table narrower to leave room for the filters in future:

image

PR #301, opened 2023-06-30
Record vaccs: orange 'could not vaccinate' banner on child page

benilovj:
This implementation deviates from the prototype designs because the app currently doesn't record the reason for vaccination not occurring.

image


misaka (approved):
🚀

PR #302, opened 2023-06-30
Add medical history component

misaka:
This component, primarily for the vaccination history page, encapsulates the logic for displaying triage information. The logic is intricate enough to warrant it's own component and a couple helper methods in models; the tests attempt to be thorough at the cost of increasing spec runtime.

This component has to address specific cases, here are a few examples of the changes:

#### Before 

image

#### After

image

#### Before

image

#### After

image

#### Before

image

#### After

image

### DIFF
--- a/app/components/app_patient_medical_history_card_component.html.erb
+++ b/app/components/app_patient_medical_history_card_component.html.erb
@@ -1,0 +1,48 @@
+<div class="nhsuk-card">
+  <div class="nhsuk-card__content">
+    <h2 class="nhsuk-card__heading nhsuk-heading-m">
+      Medical history
+    </h2>
+
+    <% if @consent_response.triage_needed? %>
+      <% if triage_present? %>
+        <p>Triage needed</p>
+
+        <ul class="nhsuk-list nhsuk-list--bullet">
+          <% reasons_triage_needed.each do |reason| %>
+            <li><%= reason %></li>
+          <% end %>
+        </ul>
+      <% else %>
+        <% if @triage.notes.present? %>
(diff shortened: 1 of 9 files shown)


app/components/app_patient_medical_history_card_component.html.erb
```diff
@@ -0,0 +1,48 @@
+<div class="nhsuk-card" id="consent">
```
tvararu, line 1: Is id="consent" correct? Is it used for anything?

app/components/app_patient_medical_history_card_component.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  <div class="nhsuk-card__content">
+    <h2 class="nhsuk-card__heading nhsuk-heading-m">
+      Medical history
+    </h2>
+
+    <% if @consent_response.triage_needed? %>
+      <% if triage_present? %>
+        <p>Triage needed</p>
+
+        <ul class="njsuk-list njsuk-list--bullet">
```
tvararu, line 11:
```suggestion
        <ul class="nhsuk-list nhsuk-list--bullet">
```

tvararu (approved):

app/components/app_patient_medical_history_card_component.html.erb
misaka: Pretty sure it isn't. I think it was being used for testing before but I switched it up. Kheel it?

app/components/app_patient_medical_history_card_component.html.erb
misaka: I know I'm still struggling occasionally with this kbd, but ... did I really type those out by hand?? 🤦

PR #304, opened 2023-07-01
More improvements to child page banners

benilovj:
* Add a banner for patients where triage determined "do not vaccinate"

image

* pull all banner code into a partial
* display the same banners consistently between record and triage child pages


tvararu (approved):
Nice :ship:

PR #307, opened 2023-07-02
Triage banner

misaka:
Adds a banner for the triage page explaining why triage is needed. This does not show up on the vaccination page (not in the prototype), but can be added if it turns out it's needed.

## Banner for "Triage needed"

image

## Banner for "Triage follow-up needed"

image


benilovj:
> This does not show up on the vaccination page (not in the prototype), but can be added if it turns out it's needed.
I checked with the designers, let's make the banners consistent between the triage and record sections.

misaka:
These banners have been merged in with the vaccinations ones and now show up on the vaccinations page.

benilovj (approved):

PR #311, opened 2023-07-03
Add tabs to vaccinations page

misaka:
Initial go at adding tabs. Things left to do:

- [x] Fix playwright tests.
- [x] Change "Action" to "Outcome" on "Vaccinated" and "Not vaccinated" tabs
- [x] Background should be white.
- [ ] Fix tab label background
- [ ] More snagging.

image

With the flash:

image


app/controllers/vaccinations_controller.rb
```diff
@@ -7,7 +7,7 @@ class VaccinationsController < ApplicationController
   before_action :set_consent_response, only: %i[show confirm]
   before_action :set_triage, only: %i[show confirm]
 
-  layout "two_thirds"
+  layout "full_width"
```
tvararu, line 10: Isn't this equivalent to no special layout?

```suggestion
```

misaka: I thought we needed this to add the flash ... ?

benilovj: > I thought we needed this to add the flash ... ?

we shouldn't need to.

i also think we should leave this at two-thirds

misaka: Back to two-thirds it is!

tvararu (approved):

benilovj (approved):

PR #318, opened 2023-07-03
Add some docs of the states a patient can be in

misaka:
Could use some confirmation that these make sense.


tvararu (approved):
The state diagram makes sense to me

PR #320, opened 2023-07-03
Record vaccinations: Rudimentary client-side patient name search

benilovj:
This doesn't work without client-side JavaScript. On a mobile, all three text boxes will be visible.

The code leaves room for improvement 😬 (shouldn't be generating JS dynamically).

https://github.com/nhsuk/record-childrens-vaccinations/assets/23801/b7302a9e-e52f-4969-bf6b-ffc2491f5fc6


misaka (approved):
Checked in review apps, looks good.

PR #326, opened 2023-07-07
Replace CSS govuk classes with nhsuk ones

tvararu:
This brings the interface closer in line to the prototype.

### DIFF
--- a/app/views/errors/internal_server_error.html.erb
+++ b/app/views/errors/internal_server_error.html.erb
@@ -1,17 +1,17 @@
 <%= content_for :page_title, "Sorry, there’s a problem with the service" %>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">Sorry, there’s a problem with the service</h1>
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-two-thirds">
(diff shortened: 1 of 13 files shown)


config/initializers/govuk_formbuilder.rb
```diff
@@ -3,8 +3,8 @@
   #
   # https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder
 
-  # config.brand: 'govuk'
-  #
+  config.brand = "nhsuk"
```
benilovj, line 6: @tvararu does this potentially open the door to using the form builder in lots more places?

tvararu: @benilovj I think so... I haven't tested this in anger so I don't know if it works for every single component or whether some form elements are different. We'll have to see.

tvararu: Unfortunately, I've been looking, but the same can't be easily done for `govuk-components`. Some of the NHS components have a different HTML structure anyway, so simply changing the CSS class wouldn't make much of a difference.

tvararu:
Unfortunately the second commit breaks the JS for the nested checkboxes, as found by the Playwright tests. I'm removing it from this PR.

image

PR #329, opened 2023-07-10
URA state machine

misaka:
This PR supersedes, and is base on, the state machine diagrams PR #318.

It introduces a state machine for `PatientSession` to replace the existing system to derive state. It uses a concern to keep  the state machine logic wrapped up in it's own module. This also allows for snappy specs by removing the need to have a `PatientSession` constructed for each spec.

This PR is currently in draft to get feedback on the implementation and the design of the states and transitions. The approach I landed on was to have states that represented all the possible states a patient could have within a session, with transitions that performed the logic to determine the correct state to transition to.

I've also left transitionary commits in to help illustrate the different approaches I was trying, feel free to tell me if you prefer one of the other approaches or have a suggestion for a better one.

Remaining work:
- [ ] Add missing transitions from "Check Refusal" for gillick competence or verbal consent
- [ ] Add "not vaccinated" states to represent whether patient cannot be vaccinated for the session or the campaign
- [ ] Integrate with the existing system


app/models/concerns/patient_session_state_machine_concern.rb
```diff
@@ -0,0 +1,92 @@
...
+        transitions from: :consent_given_triage_needed,
+                    to: :triaged_ready_to_vaccinate,
+                    if: [:triage_ready_to_vaccinate?]
+
+        transitions from: :consent_given_triage_needed,
+                    to: :triaged_do_not_vaccinate,
+                    if: [:triage_do_not_vaccinate?]
+      end
+
+      event :process_vaccination_result do
```
benilovj, line 41: Not loving this event name. Maybe:

```suggestion
      event :record_vaccination_result do
```

app/models/concerns/patient_session_state_machine_concern.rb
```diff
@@ -0,0 +1,92 @@
...
+    def triage_do_not_vaccinate?
+      triage&.do_not_vaccinate?
+    end
+
+    def vaccination_administered
+      vaccination_record&.administered?
+    end
+
+    def vaccination_not_administered
+      vaccination_record.administered == false
```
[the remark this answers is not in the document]
misaka: 💯  ... I lost that when playing around with another way of doing it, you're right it should've been put back.

app/models/concerns/patient_session_state_machine_concern.rb
misaka: I actually originally had it as `record_vaccination_result`, but changed it because:

- Technically the result is already recorded in the `VaccinationRecord`, so it didn't quite seem right to use "record" here after it's been recorded.
- The other events also started with `process_` (if I ignore your other comments about using "process" 😄). This actually makes some sense (to me) since the actual state is recorded in other objects, and we're processing them to understand what the next state is.
- The `aasm` gem creates predicate methods by prefixing the event names with `may_`, e.g.: `may_record_vaccination_result`. Again, the result is already recorded, so this doesn't sound right, whereas `may_process_vaccination_result` sounds better.

Not to say that `process_` is correct, but these are the reasons I ended up with it. Happy to give it more thought or discussion.

misaka:
Ugh. I could've swore I clicked on the option to make this PR a draft. 🤦

app/models/concerns/patient_session_state_machine_concern.rb
tvararu: That reasoning is sound to me. I'm also thinking if there's a better word. Alternative could be:

```
event :consent
event :triage
event :vaccinate
```

Then `may_consent` etc all are really intuitive.

db/migrate/20230706141503_add_state_to_patient_sessions.rb
```diff
@@ -0,0 +1,5 @@
+class AddStateToPatientSessions < ActiveRecord::Migration[7.0]
+  def change
+    add_column :patient_sessions, :state, :string
```
tvararu, line 3: Could this be non nullable? Since we set an initial value.

```suggestion
    add_column :patient_sessions, :state, :string, null: false
```

tvararu (commented):
This system looks good! Needs the "not vaccinated" and Gillick transition mentioned in the PR review, but otherwise, I think we could merge this and try to work with it and see how it goes. :+1:

db/migrate/20230706141503_add_state_to_patient_sessions.rb
misaka: Good point, let me try that.

app/models/concerns/patient_session_state_machine_concern.rb
misaka: I think the events `consent`, `triage` and `vaccinate` could work, ya, if we can sort out the clash.

db/migrate/20230706141503_add_state_to_patient_sessions.rb
tvararu: I think we can roll forward, so I'd rather have this merged and ready to use. Then we can start integrating and iterating

misaka:
Here's how this looks on the console, by the way:

```
[3] pry(main)> ps = PatientSession.all.sample
  PatientSession Load (2.4ms)  SELECT "patient_sessions".* FROM "patient_sessions"
=> #<PatientSession:0x00000001287d5ab8 id: 2, session_id: 1, patient_id: 2, created_at: Tue, 11 Jul 2023 10:02:45.831124000 BST +01:00, updated_at: Tue, 11 Jul 2023 10:02:45.831124000 BST +01:00, state: "added_to_session">
[4] pry(main)> ps.may_process_triage
NoMethodError: undefined method `may_process_triage' for #<PatientSession id: 2, session_id: 1, patient_id: 2, created_at: "2023-07-11 10:02:45.831124000 +0100", updated_at: "2023-07-11 10:02:45.831124000 +0100", state: "added_to_session">
from /Users/misaka/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/activemodel-7.0.6/lib/active_model/attribute_methods.rb:450:in `method_missing'
[5] pry(main)> ps.may_process_triage?
=> false
[6] pry(main)> ps.may_process_consent?
  Patient Load (0.2ms)  SELECT "patients".* FROM "patients" WHERE "patients"."id" = $1 LIMIT $2  [["id", 2], ["LIMIT", 1]]
  Session Load (0.2ms)  SELECT "sessions".* FROM "sessions" WHERE "sessions"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  Campaign Load (0.2ms)  SELECT "campaigns".* FROM "campaigns" WHERE "campaigns"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  ConsentResponse Load (0.3ms)  SELECT "consent_responses".* FROM "consent_responses" WHERE "consent_responses"."patient_id" = $1  [["patient_id", 2]]
=> true
[7] pry(main)> ps.process_consent
=> true
[8] pry(main)> ps.state
=> "consent_given_triage_not_needed"
[9] pry(main)> ps.may_process_triage?
=> false
[10] pry(main)> ps.consent_given?
=> true
[11] pry(main)> ps.triage_needed?
=> false
[12] pry(main)> ps.vaccination_administered
  VaccinationRecord Load (0.8ms)  SELECT "vaccination_records".* FROM "vaccination_records" WHERE "vaccination_records"."patient_session_id" = $1 ORDER BY "vaccination_records"."id" DESC LIMIT $2  [["patient_session_id", 2], ["LIMIT", 1]]
=> nil
```

I know this isn't our primary use case, but it does seem to show decent ergonomics.

db/migrate/20230706141503_add_state_to_patient_sessions.rb
misaka: > Could this be non nullable? Since we set an initial value.

I tried this, and the fly in the ointment is that Rubocop complains, it wants us to set a default when `null: false`, which means we have do define the default in two places ... which I'm not a fan of. We then have two sources of truth.

Options:

1) Set the default column value in SQL and have two sources of truth. In fact it seems the SQL default will override the aasm default, and if they don't match aasm raises a model validation error "State is invalid". This is better than them being out of sync, but this still smells a little to me.
2) Tell Rubocop to ignore the check for a default value and not bother with it.
3) Go back to `null: true` since aasm handles this quite nicely and generally, with a few other `null:false` exceptions, we rely on in-app validations.

misaka:
@benilovj Theo and I chatted about the action names and we agreed it's nice to have action names that are more clearly verbs/actions. We tried to find some other synonym to `process` since that word is a little overloaded, and came up with `do_*`. WDYT?

benilovj:
> @benilovj Theo and I chatted about the action names and we agreed it's nice to have action names that are more clearly verbs/actions. We tried to find some other synonym to `process` since that word is a little overloaded, and came up with `do_*`. WDYT?

:shipit:

misaka:
Ok, ready to merge once approval is given.

PR #338, opened 2023-07-13
Update e2e tests

misaka:
Many things changed in the rush to get the model office working, this brings the test data to parity with those changes and updates the tests to use this new data. It also updates tests for new features, primarily tabs.

As a last touch the triage specs are now split up into two files, one per user-facing feature. Some functions have been moved into a shared steps file as an experiment to see if it's worth pursuing.


tests/example_data.ts
```diff
@@ -3,6 +3,57 @@
 import * as fs from "fs";
 import * as path from "path";
 
+export const patient_expectations = {
```
tvararu, line 6:
```suggestion
export const patientExpectations = {
```

tvararu: Rest of the file should be camelCased variables too as that's idiomatic in JS/TS.

tests/triaging_patients.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+  ).toHaveClass(new RegExp(colourClass));
+}
+
+export async function then_i_should_see_the_triage_page_for_the_patient(name) {
+  await expect(p.locator("h1"), `[${name}] Triage page title`).toContainText(
+    name,
+  );
+}
+
+export async function and_i_should_see_the_consent_section_for_the_patient(
```
tvararu, line 110: Functions like these feel to me like they're doing too much.

I think we should find some time, perhaps after the second model office session, to scale back our end to end tests and remove looping/conditionals/passed in arguments. I feel like we can lift these checks into unit/component tests and only rely on system/e2e tests to test core user journeys at a high level.

For example, in the method above, I don't think we should need to test for the colour of a tag in a system spec. That should be covered elsewhere.

tvararu (approved):
Approving as comments are non-blocking. Perhaps I can do a bit of refactoring on this after merging.

tests/example_data.ts
misaka: Thanks. I'd meant to switch that!

tests/triaging_patients.spec.ts
misaka: Ya, the viewing triage test is definitely a bit heavy. It did help catch out some things, though, so not all waste.

tvararu (approved):

PR #341, opened 2023-07-17
Move to using state machine

misaka:
The main change here is to use the state machine instead of services that derived the state from the existing objects. The changes include:

**Triage and vaccinations filtering**

The controllers now pass a list of `PatientSession` objects instead `Patient` with their `action` and `outcome`. The views have a simpler mechanism for filtering by simply looking at the state on `PatientSession`.

**Patient row views**

Refactored to use new helper methods `status_text_for_state` and `status_colour_for_state`, and to rely on `patient_session` passed in from the tabs to get `state`, `patient` and `session`.

**PatientSessionHelper** and **VaccinationsHelper**

Moved action/outcome colours to `PatientSessionHelper` and put them together with action/outcome (now called status for brevity) text, banner title and banner explanations. The goal is to keep the status data together for ease of maintenance.

**Banners**

Refactored to use `PatientSessionHelper` instead of the now defunct services. This refactoring is incomplete and we should move the banners into their own components `AppTriageBannerComponent` and `AppVaccinationBannerComponent`.

**PatientSessionStateMachineConcern**

Added `triaged_kept_in_triage` state.


misaka:
@tvararu and I chatted about the `PatientActionOrOutcomeService` specs and initially I thought it wasn't worth carrying over, but now I'm having second thoughts. I'm going to give it a go.

PR #353, opened 2023-07-19
Consent

tvararu:
Review commit by commit.

Fleshes out the "Check your answers" page for adding consent, adding a `recorded_at` field to track when this was done and whether a ConsentResponse is "draft." Also triggers the state machine ".do_consent" transition when recording the consent.

 ### Patient needs consent (nothing changed)

image

### No consent given (previous PR)

image

### Check your answers page (now with content)

image

### Consent saved, patient now shows up in No triage needed

image
image

### Consent shown in show page

image


config/routes.rb
```diff
@@ -30,7 +30,8 @@
         get "record-template", on: :collection
       end
 
-      get "/consent/confirm", to: "consent_responses#confirm", on: :member
+      post "/consent/confirm", to: "consent_responses#confirm", on: :member
+      put "/consent/record", to: "consent_responses#record", on: :member
```
tvararu, line 34: These could probably be replaced with a `resource` call or similar.

app/controllers/consent_responses_controller.rb
```diff
@@ -1,8 +1,54 @@
...
   before_action :set_session
   before_action :set_patient
+  before_action :set_patient_session
+
+  layout "two_thirds"
 
   def confirm
+    @draft_consent_response = @patient
+      .consent_responses
+      .find_or_initialize_by(recorded_at: nil)
```
tvararu, line 11: This could go into a `before_action` since it's duplicated.

app/models/patient.rb
```diff
@@ -61,7 +61,9 @@ def triage_for_campaign(campaign)
   end
 
   def consent_response_for_campaign(campaign)
-    consent_responses.to_a.find { |t| t.campaign_id == campaign.id }
+    consent_responses.to_a.find do |t|
+      t.campaign_id == campaign.id && t.recorded_at.present?
+    end
```
tvararu, line 66: This breaks the state machine spec: `bin/bundle exec rspec ./spec/models/concerns/patient_session_state_machine_concern_spec.rb`

app/views/consent_responses/confirm.html.erb
```diff
@@ -6,3 +6,50 @@
...
+              <br />
+              Telephone: <%= @draft_consent_response.parent_phone %>
+            <% end %>
+            <% if @draft_consent_response.parent_email.present? %>
+              <br />
+              Email: <%= @draft_consent_response.parent_email %>
+            <% end %>
+          </div>
+        <% end
+      end
```
misaka, line 43: Would it be worth using the consent card component? It looks like it should mostly work here.

misaka: Just saw the commit message 😇  ... seems a pain to have the above section repeated, but guess we can 👞 🥫

config/routes.rb
```diff
@@ -30,7 +30,10 @@
         get "record-template", on: :collection
       end
 
-      get "/consent/confirm", to: "consent_responses#confirm", on: :member
+      resource :consent_responses, path: "consent" do
+        post "confirm"
```
misaka, line 34: I'm guessing `/confirm` will be a GET eventually, and it's a POST for now only because it's actually modifying data, is that right?

spec/factories/consent_responses.rb
```diff
@@ -39,22 +40,6 @@
...
-    "Is there anything else we should know?"
-  ],
-  hpv: [
-    "Does the child have any severe allergies that have led to an anaphylactic reaction?",
-    "Does the child have any existing medical conditions?",
-    "Does the child take any regular medication?",
-    "Is there anything else we should know?"
-  ]
-}.freeze
-
```
misaka, line 57: I think the reason these made sense here was that we think we'll eventually have to put these into the database in their own table. I'm not sure it matters much if they're here or in `ConsentResponse`.

config/routes.rb
```diff
@@ -30,7 +30,10 @@
...
+        put "record"
```
misaka, line 35: I get why this is `record` and not `create` ... this action will only update `recorded_at`, but it still feels odd.

misaka (approved):
Nice work, looks good!

config/routes.rb
tvararu: Yes

PR #355, opened 2023-07-20
Consent: Health questions page

tvararu:
See commits.

### After

image

image

### DIFF
--- a/app/views/consent_responses/edit_questions.html.erb
+++ b/app/views/consent_responses/edit_questions.html.erb
@@ -7,6 +7,22 @@
 
 <h1 class="nhsuk-heading-l">Health questions</h1>
 
-<%= govuk_button_to "Continue",
-  session_patient_consent_responses_path(@session, @patient),
-  method: :put %>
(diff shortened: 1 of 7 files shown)


app/views/consent_responses/edit_questions.html.erb
```diff
@@ -0,0 +1,28 @@
+<% content_for :before_main do %>
+  <%= render AppBacklinkComponent.new(
+               href: session_patient_triage_path(@session, @patient),
+               name: "triage page",
+             ) %>
+<% end %>
+
+<h1 class="nhsuk-heading-l">Health questions</h1>
+
+<%= form_for @draft_consent_response,
```
misaka, line 10: Just thinking ... for now this is a draft `consent_response` but this should with a non-draft `consent_response` object too, right? So at some point this will just be `@consent_response` I guess.

misaka (approved):
🚂 Nice. Like the last commit says, this will benefit from a little more refinement, but going in the right direction!

PR #358, opened 2023-07-21
Vaccine batch flow

misaka:
This adds the option to select the batch when recording a vaccination:

image

It also adds the batch info to the confirmation page:

image

And also to the Vaccination details of the patient vaccination page:

image

This PR also removes the Vaccination details section if the vaccine wasn't administered, as per the prototype.


app/controllers/vaccinations_controller.rb
```diff
@@ -95,12 +88,40 @@ def record_template
...
+    end
+  end
+
   private
 
+  # def needs_reason?(vaccination_record)
+  #   not_administered = vaccination_record.administered?
+  #   reason_not_specified = vaccination_record_params[:reason].blank?
+  #   not_administered && reason_not_specified
+  # end
```
tvararu, line 119: Do we want to leave this in?

tvararu (approved):

app/controllers/vaccinations_controller.rb
misaka: Oops. Thanks, removed.

PR #359, opened 2023-07-24
Automatically run db:migrate when releasing to Heroku

tvararu:
There's no reason we shouldn't use this on Heroku, since it isn't production data anyway.


misaka (approved):

tvararu:
Review app appears to not be deploying correctly. Closing this for now as it's a rabbit hole.

misaka (approved):
Let's give it a go?

tvararu:
@misaka had to modify this to get it to work on review apps. Check it out now.

db/migrate/20230113195033_update_children_to_latest.rb
```diff
@@ -1,8 +1,15 @@
...
+
     # Do not run this migration on production or on existing data. It is a
     # destructive migration. Remove child records before running.
-    raise "Child records exist, cannot run migration" if Child.any?
+    if ActiveRecord::Base
+      .connection
+      .execute("SELECT EXISTS (SELECT 1 FROM children)")
+      .first["exists"]
+      raise "Child records exist, cannot run migration"
+    end
```
misaka, line 12: Can we just remove this migration? Or do we actually still need it?

tvararu: I've removed it!

PR #361, opened 2023-07-24
Allow users to modify consent response `consent`

tvararu:
For the "No response" journey, this ends up skipping the transaction that updates the `recorded_at`, effectively leaving the record in draft. This matches the prototype, but feels "unfinished." We'll be revisiting consent more thoroughly at a later date.

See commits for context on some of the tangentially related bits that were included in this PR.

TODO:

- [x] Consent refusal reason screen
- [x] Update state machine to deal with `no_response`
- [x] Validations for `reason_for_refusal`
- [x] Display `reason_for_refusal_other` field
- [x] Fix error summary JS -- separate PR https://github.com/nhsuk/record-childrens-vaccinations/pull/363

Review notes:

- [ ] Consider naming the new enum value `not_provided` so it reads as `consent_not_provided` instead of `consent_no_response`
- [ ] Consider naming action as `edit_consent` instead of `edit_agree`

### Consent: Validations

image

### Reason for refusal

image

### Consent given

image

### Consent refused

image

### Reason for refusal: Validations

image

image

### No response (Confirm is effectively a no-op)

image

### Reason for refusal: Other text field (dev-design)

image


app/controllers/consent_responses_controller.rb
```diff
@@ -38,8 +38,10 @@ def edit_confirm
   end
 
   def record
-    @draft_consent_response.update!(recorded_at: Time.zone.now)
-    @patient_session.do_consent!
+    ActiveRecord::Base.transaction do
+      @draft_consent_response.update!(recorded_at: Time.zone.now)
+      @patient_session.do_consent!
+    end
```
misaka, line 85: Good point, we should be doing this elsewhere too.

app/controllers/consent_responses_controller.rb
```diff
@@ -13,22 +13,61 @@ def create
...
+          redirect_to action: :edit_reason
+        else
+          redirect_to action: :edit_confirm
+        end
+      else
+        render :edit_agree
+      end
+    end
+
+    if consent_response_reason_params.present?
```
misaka, line 45: My expectation here, knowing that we have multi-page editing, is that this method would choose one of the pages to respond to, either "Do they agree ...", the health questions or the reason for refusal, but because it's not using `elsif` I'm suddenly not sure ... is that intentional? Could updating these params overlap, and is that why that's done.

I can see cases where we might do that ... perhaps to facilitate an API call or if we have a form that updates more things at once ... but when mixed in with these renders and redirects I think it's a bit confusing. Also doesn't help that this method is getting long.

This is tricky ... happy to roll-forward with this and address it later.

app/controllers/consent_responses_controller.rb
```diff
@@ -13,22 +13,61 @@ def create
...
+      @draft_consent_response.assign_attributes(consent_response_reason_params)
+      if @draft_consent_response.save(context: :edit_reason)
+        redirect_to action: :edit_confirm
+      else
+        render :edit_reason
+      end
+    end
```
misaka, line 52: Testing on the deployed review apps server, it seems the validations for `consent_response_agree_params` aren't working, it doesn't render anything.

misaka: https://github.com/nhsuk/record-childrens-vaccinations/assets/15608/e679d1ce-2eb5-4145-8f98-b09b77c15827

app/controllers/consent_responses_controller.rb
tvararu: Yeah, it should probably be elsif, or we should break this up into separate methods.

misaka (approved):
Seems validation isn't working ... but this is a change we need and this implementation is looking really good. We've agreed we'll roll forward with any other changes.

PR #363, opened 2023-07-24
Fix nhsuk-error-summary

tvararu:
The summary isn't getting focused because the GOVUK Error Summary JS doesn't match the NHSUK Frontend HTML.

### After

image


tvararu:
This now works and is ready for review.

misaka (approved):

PR #371, opened 2023-07-25
Add assessing gillick competence journey

tvararu:
This adds the form to the vaccinations show page that routes users to either:

- The first step in the consent_responses#new journey, when obtaining consent via phone
- The new assessing_gillick, or pre-gillick page, that displays some informative text about this route
- A no-op `head :no_content` result, if there's no consent, that we'll add in a follow-up PR

The button on the assessing gillick page currently leads to a no-op.

There are no validations for the form; submitting without selecting something is a no-op.

### Vaccination page: No consent, new form

image

### Assessing gillick page

image


app/controllers/vaccinations_controller.rb
```diff
@@ -8,6 +8,7 @@ class VaccinationsController < ApplicationController
   before_action :set_vaccination_record, only: %i[show confirm record]
   before_action :set_consent_response, only: %i[show confirm]
   before_action :set_triage, only: %i[show confirm]
+  before_action :set_draft_consent_response, only: %i[show consent]
```
misaka, line 11: Does `consent` need this? I don't see it using the `@draft_consent_response` in the controller and it appears to always redirect?

app/views/vaccinations/show.html.erb
```diff
@@ -23,7 +25,30 @@
   ) %>
 <% end %>
 
-<% if @vaccination_record&.administered? %>
+<% if @consent_response.nil? %>
+  <div class="nhsuk-card">
+    <div class="nhsuk-card__content">
+      <%= form_with model: @draft_consent_response,
```
misaka, line 31: I wonder if it's correct that this should ever be pre-populated ... same issue with the other places where we have radio buttons on a patient page.

config/routes.rb
```diff
@@ -26,16 +26,22 @@
         get "confirm", on: :member
         put "record", on: :member
 
+        post "consent", on: :member
```
misaka, line 29: Does `VaccinationsController#consent` create anything? Or is this going to be changed in the future?

app/views/vaccinations/show.html.erb
tvararu: Yeah, you're right that it could be prepopulated here which is probably unnecessary. :thinking: 

Let's kick this down the road

config/routes.rb
tvararu: No, I wasn't sure what to use between `POST/PATCH/PUT` since nothing is being mutated. Thoughts?

tvararu: `POST` is probably closer in intent to "this will eventually create a new consent, just not on this specific action" :thinking:

tvararu: Posted two comments before I remembered; yes, the "No" route should create a consent.

misaka (approved):

PR #374, opened 2023-07-25
Add edit gillick competence page

tvararu:
Adds the `gillick_competence_details` text field to the database and creates the form to add the Gillick competence details.

This doesn't complete the gillick competence journey, but does wire up another page. Follow-up PRs will add:

- The "Not gillick competent" journey, which skips the "Check answers" section, and comes with a new banner
- Updating the consent response cards to show Gillick competence instead
- Updating the check your answers section

### Edit gillick page

image

### Validations

image


app/controllers/consent_responses_controller.rb
```diff
@@ -11,21 +11,33 @@ def new
...
       if @draft_consent_response.save(context: :edit_who)
         redirect_to action: :edit_consent
       else
         render :edit_who
       end
+    else
+      # If the params are missing, assume this is the Gillick competence route.
+      # This feels like it could be more explicit.
+      @draft_consent_response.update!(
+        campaign: @session.campaign,
```
misaka, line 35: Should this be set in `set_draft_consent_response`? Would there be any situation we don't want it set?

misaka: Come to think of it, when finding the draft record in `set_draft_consent_response` it doesn't specify the campaign ... it probably should.

misaka (commented):
I feel like `set_draft_consent_response` should be fixed so that it creates a `consent_response` with the correct campaign, as this feels like it could lead to a subtle bug later.

misaka (approved):
🚀

PR #379, opened 2023-07-25
Add Gillick cards and other tweaks

tvararu:
Review commit by commit.

### Don't display "Not provided" option in Gillick route

image

### Gillick card on confirm page

image

### Redirect to vaccination page with flash

image


app/models/consent_response.rb
```diff
@@ -111,6 +111,10 @@ class ConsentResponse < ApplicationRecord
     ]
   }.freeze
 
+  def gillick_competent?
+    via_self_consent? && !consent_not_provided?
+  end
```
tvararu, line 116: Wondering if we should make this an explicit boolean instead.

misaka: If we keep this as part of the `ConsentResponse` model, then would it make sense to make it part of the `consent` enum? So the possible values would be:

```
  enum :consent, %w[given refused not_provided gillick_competent], prefix: true
```

benilovj: I'm not sure I agree with the modelling / naming here, as it could lead to confusion / bugs.

We have two separate concepts here:

* is the child Gillick competent? Y/N. This is true at a point in time, for a given session, as assessed by the nurse. It, in itself, isn't enough to vaccinate. Gillick competence is independent of consent.
* **if** the child is Gillick competent, only then can they self-consent, or override their parent's consent. The child still needs to give or refuse consent as an explicit step (as far as I understand).

So it makes most sense to me for the Gillick competence data to live on `PatientSession`:

* a database boolean (`gillick_competent`)
* assessment details (`gillick_competence_details`)

As for `enum :consent`, I think it ought to be left as is (`%w[given refused not_provided`).

tvararu: Agree that the consent enum should stay the same. And I agree, it makes sense to put the gillick competence fields on the patient session. I'll get on that :+1:

tvararu: Done, see latest commits!

tvararu:
https://github.com/nhsuk/record-childrens-vaccinations/pull/384 fixes the CI build on this.

misaka (approved):

PR #380, opened 2023-07-25
Banner bugfix when patient ready to vaccinate without triage

benilovj:
## Context

There is consent and no contraindications, so the patient is ready to vaccinate:

image

## What happens before the fix

According to the banner, a nurse has deemed the patient can be vaccinated. This isn't true as this record hasn't gone through triage:

image

## What happens after the fix

The description has been removed.

image

## Notes for reviewers

* Best reviewed commit-by-commit
* I've added extra spec coverage to `app_status_banner_component_spec`


misaka (approved):
Nice one!

tvararu (approved):

PR #382, opened 2023-07-25
Handle journey where patient is not Gillick competent

tvararu:
Builds on https://github.com/nhsuk/record-childrens-vaccinations/pull/379.

This adds a new transition from the `added_to_session` state to the `unable_to_vaccinate_not_gillick_competent` state. It updates the controller when the consent_response is being updated so that it shortcuts to this state.

### After

image

### DIFF
--- a/app/controllers/consent_responses_controller.rb
+++ b/app/controllers/consent_responses_controller.rb
@@ -83,7 +83,24 @@ class ConsentResponsesController < ApplicationController
   def update_gillick
     @patient_session.assign_attributes(patient_session_gillick_params)
     if @patient_session.save(context: :edit_gillick)
-      redirect_to action: :edit_consent
+      if @patient_session.gillick_competent?
+        redirect_to action: :edit_consent
+      else
+        ActiveRecord::Base.transaction do
+          @draft_consent_response.update!(
+            recorded_at: Time.zone.now,
+            consent: "not_provided"
+          )
+          @patient_session.do_gillick_assessment!
+        end
+
+        redirect_to session_patient_vaccinations_path(@session, @patient),
+          flash: {
+            success: {
+              body: "Gillick assessment saved for #{@patient.full_name}",
+            },
+          }
+      end
     else
       render :edit_gillick
     end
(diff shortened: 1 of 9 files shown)


app/controllers/consent_responses_controller.rb
```diff
@@ -83,7 +83,24 @@ def update
...
-      redirect_to action: :edit_consent
+      if @patient_session.gillick_competent?
+        redirect_to action: :edit_consent
+      else
+        ActiveRecord::Base.transaction do
+          @draft_consent_response.update!(
+            recorded_at: Time.zone.now,
+            consent: "not_provided"
+          )
+          @patient_session.do_consent!
```
tvararu, line 94: We should add a triage here.

app/models/concerns/patient_session_state_machine_concern.rb
```diff
@@ -97,5 +104,9 @@ def vaccination_administered?
     def vaccination_not_administered?
       vaccination_record&.administered == false
     end
+
+    def not_gillick_competent?
+      !gillick_competent?
+    end
```
misaka, line 110: Feels like this would be nice to have on `PatientSession`

app/models/concerns/patient_session_state_machine_concern.rb
```diff
@@ -29,6 +30,12 @@ module PatientSessionStateMachineConcern
                     if: :consent_refused?
       end
 
+      event :do_gillick_assessment do
+        transitions from: :added_to_session,
+                    to: :unable_to_vaccinate_not_gillick_competent,
+                    if: :not_gillick_competent?
```
misaka, line 36: Are we planning to add a transition for when `gillick_competent?` is true?

app/controllers/consent_responses_controller.rb
```diff
@@ -83,7 +83,24 @@ def update
...
-      redirect_to action: :edit_consent
+      if @patient_session.gillick_competent?
+        redirect_to action: :edit_consent
+      else
+        ActiveRecord::Base.transaction do
+          @draft_consent_response.update!(
+            recorded_at: Time.zone.now,
+            consent: "not_provided"
+          )
+          @patient_session.do_gillick_assessment!
```
misaka, line 94: It would be nice if we continued the pattern of calling this event regardless of the outcome of gillick competence test, so that when successful the patient session transitions to an appropriate state.

misaka (approved):

PR #385, opened 2023-07-26
Handle "No, I am not vaccinating" journey

tvararu:
Add the controller action that triggers the state transition, and hide the vaccination CTAs unless the user is able to be vaccinated.

### Ticking the "No" radio

image

### Record saved, now showing up in a different tab

image

### Vaccination show page

image


misaka (approved):
🚀

PR #391, opened 2023-07-26
Tab counts in record area

benilovj:
## What it looks like

image

## Comments for reviewers

This PR moves the tab partitioning logic out of the view and into the controller.

### DIFF
--- a/tests/default_batch.spec.ts
+++ b/tests/default_batch.spec.ts
@@ -18,7 +18,7 @@ test("Sets default batch for today", async ({ page }) => {
   await when_i_press_confirm();
   await then_i_should_see_a_success_message();
 
-  await when_i_click_on_the_tab("Action needed");
+  await when_i_click_on_the_tab("Action needed (7)");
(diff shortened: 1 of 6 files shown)


tests/default_batch.spec.ts
```diff
@@ -18,7 +18,7 @@ test("Sets default batch for today", async ({ page }) => {
   await when_i_press_confirm();
   await then_i_should_see_a_success_message();
 
-  await when_i_click_on_the_tab("Action needed");
+  await when_i_click_on_the_tab("Action needed (7)");
```
misaka, line 21: Wonder if it would be worth changing these tests to not need the count.

misaka (approved):
:shipit:

tests/default_batch.spec.ts
benilovj: Yes... 
One **really** annoying thing is that the tab IDs also change if the label changes. I'm also not sure that the `c.with_tab` helper allows overriding the `id` attribute.

PR #393, opened 2023-07-27
Other vaccination site

misaka:
I think this is done ... with validations etc. Burning the midnight oil a bit here, so worth testing out in the review app.

image

With validations:

image


benilovj (approved):
Great work @misaka – I really hope you managed to get some 💤 afterwards!

Non-blocking thing, prob for a follow-up PR (and probably involves a design decision)... We ought to surface the delivery method on the patient page, after the vaccination has been given:

image

Maybe it should only be shown if the vaccination was subcutaneous? Or in every case? Not sure but gut feeling says to always show it.

benilovj:
Functionality looks good on the review app. I will merge this in so we can test in model office.

PR #397, opened 2023-07-27
Allow triage when records in consent_refused state

misaka:
When consent is refused nurses will still do triage and could transition to other states.

I think this is the fix, if so I will add tests.


misaka:
This can partly fix the issue, but exposes other issues. Better to roll together a bigger fix for the triage & consent response flows I think.

PR #399, opened 2023-07-27
Allow triage transition when consent refused

misaka:


misaka:
After some more manual testing, this fix actually seems reasonable. The state transitions make sense, and the guard against triggering the `do_triage` event when there's not triage object also works. This'll also a reminder/hint to consider this when refactoring.

misaka:
While looking at the tests, I realised this is actually all that needs doing for the moment.

tvararu (approved):

misaka:
I think this is another example of why we would benefit from some more testing that maybe isn't Playwright e2e tests, but some kind of journey testing that covers more edge-cases. Essentially BDD tests with more coverage than our e2e tests.

tbh, I think Playwright tests would be acceptable for this.

PR #403, opened 2023-07-31
Add sentry integration

misaka:
Generic Sentry integration added. Relies on `SENTRY_DSN` being set in the deployed environment.

This change has been tested, however we don't have a team account on Sentry. At the moment I'm just using a DSN for the organisation "Good Machine" setup on my account; to invite other members to it will require a team, business or open source account, so far as I can see.

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,0 +1,12 @@
+require 'active_support/parameter_filter'
+
+Sentry.init do |config|
+  config.breadcrumbs_logger = [:active_support_logger, :http_logger]
+
+  config.traces_sample_rate = 1.0
+
+  filter = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters)
+  config.before_send = lambda do |event, _hint|
+    filter.filter(event.to_hash)
+  end
+end
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -5,8 +5,30 @@ Sentry.init do |config|
 
   config.traces_sample_rate = 1.0
 
-  filter = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters)
+  rails_filter_parameters =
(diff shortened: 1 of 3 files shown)


misaka:
I've added a `SENTRY_DSN` for the Heroku staging environment so in theory this should work once deployed. It might be worth adding to the review apps to check it's functionality there.

config/initializers/sentry.rb
```diff
@@ -0,0 +1,12 @@
...
+
+Sentry.init do |config|
+  config.breadcrumbs_logger = [:active_support_logger, :http_logger]
+
+  config.traces_sample_rate = 1.0
+
+  filter = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters)
+  config.before_send = lambda do |event, _hint|
+    filter.filter(event.to_hash)
+  end
```
tvararu, line 11: I recommend that we prevent Sentry from logging for 422 and "user errors," which can be done with an excluded exceptions array like this:

https://github.com/DFE-Digital/find-a-lost-trn/blob/main/config/initializers/sentry.rb#L26-L34

misaka: Would it maybe be worth keeping these while we're in the model office & pilot phase? These could be interesting, and it's hidden behind auth so we shouldn't get spurious script kiddies and vulnerability scanning.

tvararu: Hmm, yeah, we can revisit this later :+1:

tvararu (approved):

benilovj:
Do we need to blank out PII fields? I think we should...

misaka:
Line 9-11 is setting up the filtering out of parameters. This depends on [/config/initializers/filter_parameter_logging.rb](https://github.com/nhsuk/record-childrens-vaccinations/blob/main/config/initializers/filter_parameter_logging.rb) being setup.

We can add that to this PR.

PR #412, opened 2023-08-01
Display current short git sha at /sha

tvararu:
Adds a new config variable and exposes it using a route.

```
$ curl http://localhost:4000/sha 
b9f1d67
```

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -38,5 +38,7 @@ module ManageVaccinations
     config.exceptions_app = routes
 
     config.time_zone = "London"
+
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -38,5 +38,7 @@ class Application < Rails::Application
     config.exceptions_app = routes
 
     config.time_zone = "London"
+
+    config.commit_sha = `git rev-parse --short HEAD`.strip
```
misaka, line 42: Calling out to the shell like this gives me the heeby-jeebies ... ¯\_(ツ)_/¯ ... I guess anything more complicated may just be a case of YAGNI.

PR #413, opened 2023-08-02
Fix Ruby formatter

tvararu:
The new Prettier version requires us to explicitly call out plugins using a .prettierrc.

### DIFF
--- a/db/migrate/20230725135855_add_gillick_competence_details_to_consent_response.rb
+++ b/db/migrate/20230725135855_add_gillick_competence_details_to_consent_response.rb
@@ -1,4 +1,6 @@
-class AddGillickCompetenceDetailsToConsentResponse < ActiveRecord::Migration[7.0]
+class AddGillickCompetenceDetailsToConsentResponse < ActiveRecord::Migration[
+  7.0
+]
   def change
     add_column :consent_responses, :gillick_competence_details, :text
   end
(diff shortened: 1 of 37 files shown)


benilovj:
So were we not actually linting lots of stuff? is that why there are so many changes?

tvararu:
@benilovj only for the last couple of weeks since Prettier 3.0.0 came out. All the changes are related to changes in that period

db/migrate/20230725135855_add_gillick_competence_details_to_consent_response.rb
```diff
@@ -1,4 +1,6 @@
-class AddGillickCompetenceDetailsToConsentResponse < ActiveRecord::Migration[7.0]
+class AddGillickCompetenceDetailsToConsentResponse < ActiveRecord::Migration[
```
benilovj, line 1: 🤔 this feels like a weird change...

benilovj (approved):

db/migrate/20230725135855_add_gillick_competence_details_to_consent_response.rb
tvararu: It's trying to wrap the lines to 80 characters where it can; that's 78 long

benilovj: aha, gotcha

PR #415, opened 2023-08-02
Add intitializer for production checks

tvararu:
We want to ensure that when we release the application, we don't forget to configure certain things, like Active Record Encryption. We can codify these to reduce the chance of this occuring in production mode.


misaka (approved):
🚀

PR #417, opened 2023-08-07
Go to vaccination flow after consenting + triaging from vaccinations page

tvararu:
See commits.

TODO:

- [x] Write an end to end test
- [x] Use a partial or component for the form instead of duplicating the code

### New "New" page

image

### DIFF
--- a/app/controllers/consent_responses_controller.rb
+++ b/app/controllers/consent_responses_controller.rb
@@ -4,6 +4,7 @@ class ConsentResponsesController < ApplicationController
   before_action :set_patient_session
   before_action :set_draft_consent_response
   before_action :set_draft_triage, only: %i[edit_questions edit_confirm update]
+  before_action :keep_consent_return_path, except: %i[record]
 
   layout "two_thirds"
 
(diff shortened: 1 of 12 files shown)


misaka:
I agree that `session` makes sense of this. I would be tempted to try using `flash` for this, and then using `keep` in the actions that are part of the consent journey. I think this is more idiomatic, maybe? The one problem, perhaps, is that if we forget to add a `keep` somewhere we could introduce a subtle bug. This might be mitigated by using a `before_action` hook ... maybe something like this:

```
class ConsentResponsesController < ApplicationController
  before_action :keep_consent_return_path, only: %i[new edit_confirm edit_questions edit_refusal update]

  def keep_consent_return_path
    flash.keep[:consent_return_path]
  end

  def record
    # redirect to flash[:consent_return_path]
  end
end

class TriageController < ApplicationController
  def show
    flash[:consent_return_path] = triage_session_path(@session)
  end
end
```

tests/consent_and_vaccinate.spec.ts
misaka, line 1: Just a minor naming nit, this name suggests vaccination is happening in this journey. Should it be "consent_during_vaccination" or maybe "consent_journey-during_vaccination" or some permutation of that like the double-dashed one of BEM?

misaka (approved):
💯  Minor naming nit, but :shipit:  either way.

PR #429, opened 2023-08-09
Simplify tests

tvararu:
Remove reliance on shared_steps in all tests.

See commits.


tvararu:
I decided to push up an optional commit that removes `triage_viewing_patients.spec.ts` and the shared modules too.

tests/vaccination_default_batch.spec.ts
```diff
@@ -32,8 +34,8 @@ async function when_i_go_to_the_vaccinations_page() {
   await p.goto("/sessions/1/vaccinations");
 }
 
-async function when_i_click_on_the_tab(name: string) {
-  await p.getByRole("tab", { name: name, exact: true }).click();
+async function when_i_click_on_the_action_needed_tab() {
+  await p.getByRole("tab", { name: "Action needed (7)", exact: true }).click();
```
misaka, line 38: How about this so that we don't tie ourselves too closely too the example data?

```suggestion
  await p.getByRole("tab", { name: /Action needed$/ }).click();
```

misaka (approved):
🚀

PR #431, opened 2023-08-10
Allow resubmitting a consent after "No response" selected

tvararu:
This fixes a bug where users are not able to submit a consent multiple times, in scenarios such as:

- Nurse attempts to contact parent
- Parent doesn't reply
- Time passes
- Nurse attempts again, parent replies and agrees

### DIFF
--- a/app/controllers/consents_controller.rb
+++ b/app/controllers/consents_controller.rb
@@ -35,6 +35,15 @@ class ConsentsController < ApplicationController
   end
 
   def update
+    if consent_who_params.present?
+      @draft_consent.assign_attributes(consent_agree_params)
+      if @draft_consent.save(context: :edit_consent)
+        redirect_to action: :edit_consent
+      else
+        render :edit_who
+      end
+    end
+
     if consent_agree_params.present?
       @draft_consent.assign_attributes(consent_agree_params)
       if @draft_consent.save(context: :edit_consent)
--- a/tests/consent_no_response.spec.ts
+++ b/tests/consent_no_response.spec.ts
@@ -1,0 +1,113 @@
+import { test, expect, Page } from "@playwright/test";
+
+let p: Page;
+
+test("Consent - No response", async ({ page }) => {
+  p = page;
+  await given_the_app_is_setup();
+
+  await when_i_select_a_child_with_no_consent_response();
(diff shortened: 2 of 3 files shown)


app/controllers/consents_controller.rb
```diff
@@ -35,6 +35,15 @@ def create
...
   def update
+    if consent_agree_params.present?
+      @draft_consent.assign_attributes(consent_agree_params)
+      if @draft_consent.save(context: :edit_consent)
+        redirect_to action: :edit_consent
+      else
+        render :edit_who
+      end
+    end
+
```
misaka, line 46: Why have to `if consent_agree_params.present?` sections?

tvararu: Rebase mistake!

misaka: Aahh, yes. This controller action really is giving me hives, though, with the checks multiple branches of logic that could be followed if more than one of these `consent_*_params.present?` come back true.

Not to suggest we do it in this PR, but how do you feel about splitting them into sub-controllers like [Vaccinations::BatchesController](https://github.com/nhsuk/record-childrens-vaccinations/blob/5a9cd92b3e67de6873196abb1558014b9b37251d/app/controllers/vaccinations/batches_controller.rb)?

tests/consent_no_response.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+let p: Page;
+
+test("Consent - No response", async ({ page }) => {
+  p = page;
+  await given_the_app_is_setup();
+
+  await when_i_go_to_the_triage_page();
+  await and_i_click_the_consent_tab();
+  await and_i_click_on_a_patient();
```
misaka, line 11: What do you think about rolling steps like these up into a single step? For example: https://github.com/nhsuk/record-childrens-vaccinations/blob/add-moar-e2e-tests/tests/consent_gillick_refused.spec.ts#L9-L11

```suggestion
  await when_i_select_a_child_with_no_consent_response();
```

misaka (approved):
👍 

There's a couple comments about stuff I've been thinking about addressing, happy to take them out of the PR if you agree about them in general.

PR #432, opened 2023-08-11
Rename consent responses to consents

tvararu:
See commits, but I recommend just lightly skimming. Anything not caught by our tests should be caught soon enough and then we should fill in that gap. This PR shouldn't stay open longer than it needs to because it touches a lot of things.


misaka (approved):
😻 

I think the only thing I might've added to this is to change the `consent` enum in `Consent` to `response`.

https://github.com/nhsuk/record-childrens-vaccinations/blob/0bad42e0afde5e97b9b58ed23518da3e505a2e26/app/models/consent.rb#L51

But perhaps that should be a separate PR.

PR #435, opened 2023-08-11
Nest consent routes

tvararu:
This allows us to pass in vaccinations or triage into the URL, which then can be used to decide where to send the user in case they use the back button, or if they finish the consent journey.

See commits.


tvararu:
Tried the concerns route, but it seems like it could be a lot more effort. It creates new bespoke route paths for the consent route, one for triage and one for vaccinations: 

image

Where this becomes an issue is in the views. We'd have to use branching logic for every POST form in the consent journey; to split up and use either one route_path or the other. Unless we create separate views and controllers... at which point, we're not reusing the consent journey.

image

misaka:
Could we use `url_for(controller: "consents", action: "new")` here?

I did a basic setup of routing concerns  locally and tried this out in the view:

```
>> request.path
=> "/sessions/1/patients/1/triage"
>> params
=> #<ActionController::Parameters {"controller"=>"triage", "action"=>"show", "session_id"=>"1", "patient_id"=>"1"} permitted: false>
>> url_for(action: "new", controller: "consents")
=> "/sessions/1/patients/1/triage/consents/new"
```

misaka:
> @misaka that works and is actually quite clean, now I'm just trying to figure out what we can do for the back buttons that also need to be aware of whether they have to go back to the triage or vaccinations section:
> 
> ```
> <% content_for :before_main do %>
>   <%= render AppBacklinkComponent.new(
>     href: session_patient_triage_path(@session, @patient),
>     name: "triage page",
>   ) %>
> <% end %>
> ```

I don't think there'll be something quite as clean for this, unfortunately, but we can still avoid branching logic maybe. In this example we still need `"triage"`:

```ruby
>> url_for(params.permit(:patient_id, :session_id).to_h.merge(controller: "triage", action: "show"))
=> "/sessions/1/patients/1/triage"
```

We could just create a helper for this:

```ruby
# routes_helper.rb

...
def triage_or_vaccination_url(url)
  if url.include? "/triage/"
    :triage
  else
    :vaccination
  end
end

def back_link_for_consent_page_path
  url_for(
    params
      .permit(:patient_id, :session_id)
      .to_h
      .merge(controller: triage_or_vaccination_consent(request.url),
             action: "show"))
end
...
```

I think it could work well if the called to the back link view component got to chose whether to use a traditional `*_path` route helper, or one of ours.

misaka (approved):

PR #440, opened 2023-08-15
Add validations specs

misaka:
These tests that validations errors are displayed on forms. With the complexity of the various form journeys we have, and given the setting that this tool will be used, these are useful to ensure that the service communicates errors in the correct way to nurses.

### DIFF
--- a/tests/vaccination_validations.spec.ts
+++ b/tests/vaccination_validations.spec.ts
@@ -1,0 +1,83 @@
+import { test, expect, Page } from "@playwright/test";
+
+let p: Page;
+
+test("Vaccination validations", async ({ page }) => {
+  p = page;
+  await given_the_app_is_setup();
+
+  await given_i_am_on_the_vaccination_page_for_a_child();
+  await when_i_continue_without_entering_anything();
+  await then_the_vaccination_validation_errors_are_displayed();
+
+  await given_i_select_that_they_got_the_vaccine();
+  await when_i_continue_without_entering_anything();
+  await then_i_see_the_vaccination_site_validation_errors();
+
+  await given_i_move_on_to_the_how_the_vaccine_was_given_page();
+  await when_i_continue_without_entering_anything();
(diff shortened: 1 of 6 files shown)


misaka:
I was originally concerned that having these tests in separate files would start to really ramp up our test run times. However, on my local system, we seem to be averaging 1s per test file ... here's a comparison of our specs before and after this change:

**Before:**
- 12 specs
- ~12s runtime

**After:**
- 18 specs
- ~17s runtime

This doesn't seem too bad a price to pay if this translates to a similar runtime in GitHub actions ... although at the moment the entire Playwright test, including installation, spinning-up, etc, is 2m ... so this doesn't seem like it'll be a big factor in slowing these down.

tests/vaccination_validations.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+async function when_i_continue_without_entering_the_vaccination_status() {
+  await p.getByRole("button", { name: "Continue" }).click();
+}
+const when_i_continue_without_specifying_a_vaccination_site =
+  when_i_continue_without_entering_the_vaccination_status;
+const when_i_continue_without_entering_it_was_given =
+  when_i_continue_without_entering_the_vaccination_status;
+const when_i_continue_without_entering_a_batch =
+  when_i_continue_without_entering_the_vaccination_status;
```
tvararu, line 44: Not sure if we need these... they're effectively comments but I feel like "when I continue without entering anything" would work just as well without the need for aliases

tvararu (approved):
Nit not blocking :shipit:

tests/vaccination_validations.spec.ts
misaka: Ya, I think I agree. I think I did this earlier when I didn't have as good names for the steps that came before and after, but I improved those and, ya, that ugliness isn't needed anymore.

PR #442, opened 2023-08-16
Error when revisiting old forms

misaka:
Users can use the browser back button to return to old forms that are part of the consent journey, resulting in undefined behaviour in the app. This change makes it clearer that what they are doing isn't supported by rendering errors when form pages are requested in an inconsistent state.

This change now renders the error `unprocessable entity` page instead of the form when there no `draft_consent` object on pages that require it. Only the `create` and `new` pages should be able to create these objects. However, a 500 error can still easily be triggered if the user creates another consent object by going back to the `assessing_gillick` page and restarts the journey there.

https://github.com/nhsuk/record-childrens-vaccinations/assets/15608/450e2d74-c19e-4624-9ada-1adbf83d51d6

Unless we want to allow them to change the Gillick competence result, maybe we should raise the same `unprocessable entity` status when the revisit the `assessing_gillick` page.

This issues also still exists in the vaccinations controller, and probably the triage controller.

https://github.com/nhsuk/record-childrens-vaccinations/assets/15608/3ebd7511-a16e-4dd8-ba67-a31728397710

What still needs doing:

- [ ] Raise this error on `assessing_gillick` page.
- [ ] Raise these errors in vaccinations controllers.
- [ ] ~~Raise these errors in triage controllers.~~

### DIFF
--- a/app/views/errors/unprocessable_entity.html.erb
+++ b/app/views/errors/unprocessable_entity.html.erb
@@ -1,15 +1,11 @@
 <%= content_for :page_title, 'Sorry, there’s a problem with the service' %>
 
-<div class="nhsuk-grid-row">
-  <div class="nhsuk-grid-column-two-thirds">
-    <h1 class="nhsuk-heading-l">Sorry, there’s a problem with the service</h1>
+<h1 class="nhsuk-heading-l">Sorry, there’s a problem with the service</h1>
 
-    <p class="nhsuk-body">Try again later.</p>
+<p class="nhsuk-body">Try again later.</p>
 
-    <p class="nhsuk-body">
-      If you continue to see this error contact the <%= t('service.name')
-      %> team: <a class="nhsuk-link" href="mailto:<%= t('service.email') %>"><%=
-      t('service.email') %></a>.
-    </p>
-  </div>
-</div>
+<p class="nhsuk-body">
+  If you continue to see this error contact the <%= t('service.name') %>
+  team: <a class="nhsuk-link" href="mailto:<%= t('service.email') %>">
+  <%= t('service.email') %></a>.
+</p>
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,13 @@
 class ApplicationController < ActionController::Base
   before_action :set_disable_cache_headers
 
+  class UnprocessableEntity < StandardError
+  end
+
+  unless Rails.configuration.consider_all_requests_local
+    rescue_from UnprocessableEntity, with: :handle_unprocessable_entity
+  end
+
   if Rails.env.staging? || Rails.env.production?
     http_basic_authenticate_with name: Settings.support_username,
                                  password: Settings.support_password,
@@ -16,5 +23,9 @@ class ApplicationController < ActionController::Base
     response.headers["Cache-Control"] = "no-store"
     response.headers["Pragma"] = "no-cache"
     response.headers["Expires"] = "0"
+  end
+
+  def handle_unprocessable_entity(_exception)
+    render "errors/unprocessable_entity", status: :unprocessable_entity
   end
 end
--- a/app/controllers/consents_controller.rb
+++ b/app/controllers/consents_controller.rb
@@ -2,7 +2,8 @@ class ConsentsController < ApplicationController
   before_action :set_session
   before_action :set_patient
   before_action :set_patient_session
-  before_action :set_draft_consent
+  before_action :set_draft_consent, only: %i[create new]
+  before_action :get_draft_consent, except: %i[assessing_gillick create new]
   before_action :set_draft_triage, only: %i[edit_questions edit_confirm update]
   before_action :keep_consent_return_path, except: %i[record]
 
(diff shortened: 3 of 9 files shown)


app/views/errors/unprocessable_entity.html.erb
```diff
@@ -1,15 +1,11 @@
 <%= content_for :page_title, 'Sorry, there’s a problem with the service' %>
 
-<div class="nhsuk-grid-row">
-  <div class="nhsuk-grid-column-two-thirds">
-    <h1 class="nhsuk-heading-l">Sorry, there’s a problem with the service</h1>
```
tvararu, line 5: I think removing two-thirds from here means that the regular error pages at /422 and /404 don't render correctly. We probably need to add the `layout :two_thirds` to the `ErrorsController`.

tvararu (commented):
Approach seems sensible, carry on with the TODOs :+1:

misaka:
@tvararu I've done a bit of a deeper dive into the consents ... there's an outstanding issue with the `who` page still showing up.

Currently the way the flow is structured, the "Get consent" button on the triage page does not create the consent object, it loads up a "new" page, and relies on the subsequent POST to create the draft version of the object. This draft version of the object doesn't need to be valid, it only has part of the required fields populated.

The sequence of requests looks like this:

```mermaid
sequenceDiagram
    participant Client
    participant TriageController
    participant ConsentController
    participant Consent

	rect lightcyan
        Client ->>+ TriageController: GET consent
        TriageController -->>- Client: 200 child page
    end

    rect lightcyan
        Note right of Client: Get consent
        Client ->>+ ConsentController: GET new
        ConsentController ->>+ Consent: find_or_initialize
        Consent -->>- ConsentController: new consent object
        ConsentController -->>- Client: 200 "Who are you trying to get consent from" form
    end

    rect lightcyan
        Note right of Client: Continue
        Client ->>+ ConsentController: POST consent -> create
        ConsentController ->>+ Consent: find_or_initialize
        Consent -->>- ConsentController: new consent object
        ConsentController ->>+ Consent: save
        Note left of Consent: Creates new<br/>consent object
        Consent -->>- ConsentController: true
        ConsentController -->>- Client: 302 edit/agree
        Client ->>+ ConsentController: GET edit/agree
        ConsentController -->>- Client: 200 "Do they agree..." form
    end
    
    rect lightcyan
        Note right of Client: Continue
        Client ->>+ ConsentController: PUT consent -> update
        ConsentController ->>+ Consent: find_or_initialize
        Consent -->>- ConsentController: persisted consent object
        ConsentController ->>+ Consent: update
        Consent -->>- ConsentController: true
        ConsentController -->>- Client: 302 edit/questions
        Client ->>+ ConsentController: GET edit/questions
        ConsentController -->>- Client: 200 "Health questions" form
    end
```

The problem here is if users use the back button, or find some other way to get to the `new` page. Assuming they've already gone through the consent process and a consent object has been created, then they can create another draft consent object, and this causes errors in other parts of the system that aren't setup to handle multiple consent objects.
 
I can see two fixes to this.

1. We return an error when they try to create a second consent object. The system can only handle one, at the moment, so we don't let them create another.
2. We change the request sequence so that the "Get consent" button creates a draft object immediately, and then the rest of the forms only update the object. This means that on the `edit_who` form, if a consent object already exists we'll return `unprocessable entity` instead of the form. Technically they shouldn't ever expect this form to work, but this seems like a more consistent approach to doing these multi-form journeys.

Here's what the request sequence for option 2 would look like:

```mermaid
sequenceDiagram
    participant Client
    participant TriageController
    participant ConsentController
    participant Consent

    rect lightcyan
	    Client ->>+ TriageController: GET consent
	    TriageController -->>- Client: 200 child page
	end

    rect lightcyan
        Note right of Client: Get consent
        Client ->>+ ConsentController: POST consent -> create
        ConsentController ->>+ Consent: create
        Note right of ConsentController: Creates new<br/>consent object
        Consent -->>- ConsentController: consent object
        ConsentController -->>- Client: 302 edit/who
        Client ->>+ ConsentController: GET edit/who
        ConsentController -->>- Client: 200 "Who are you trying to get consent from" form
    end

    rect lightcyan
        Client ->>+ ConsentController: PUT consent -> update
        Note right of Client: Continue
        ConsentController ->>+ Consent: find
        Consent -->>- ConsentController: persisted consent object
        ConsentController ->>+ Consent: update
        Consent -->>- ConsentController: true
        ConsentController -->>- Client: 302 edit/agree
        Client ->>+ ConsentController: GET edit/agree
        ConsentController -->>- Client: 200 "Do they agree..." form
    end
    rect lightcyan
        Note right of Client: Continue
        Client ->>+ ConsentController: PUT consent -> update
        ConsentController ->>+ Consent: find
        Consent -->>- ConsentController: persisted consent object
        ConsentController ->>+ Consent: update
        Consent -->>- ConsentController: true
        ConsentController -->>- Client: 302 edit/questions
        Client ->>+ ConsentController: GET edit/questions
        ConsentController -->>- Client: 200 "Health questions" form
    end
```

misaka:
I also think this approach is good, as it vastly simplifies the `ConstentController` ... at the expense of creating more controllers, but still, each with it's own little domain or responsibility. Skinnyarise all the things!

tvararu (approved):
:ship:

PR #444, opened 2023-08-21
Raise UnprocessableEntity when assessing gillick

misaka:
This plugs the whole left from a previous PR which the user to revisit gillick assessment when they were already assessed as not being Gillick competent, due to the existing code only checking for the presence of a consent record.


misaka:
@tvararu missed a case when the asesssing gillick competence page would show. I considered if it's worth adding tests, but these seem like such outliers that I'm not sure it's worth it.

tvararu (approved):

PR #452, opened 2023-08-22
Add devise and users

misaka:
Adds the devise gem and users.

This doesn't, however, enforce authentication anywhere, or add dummy users. That will be separate work for later, this is just the foundation work and leaves the service mostly untouched aside from a couple of extra routes for signing in and out that aren't yet being used.


misaka:
re: the GitGuardian comment, the secrets in question were not used and commented out (auto-generated by Devise's initialisation process). I've completely removed those secrets so that they won't be used in the future, and don't trigger any more issues.

tvararu (approved):

PR #457, opened 2023-08-23
Consent form start and confirm page

misaka:
Add `start` and `confirm` pages for consent form journey.

This adds the consent form table, model and controller. Decisions taken here:

- Create a new `ConsentForm` model. Even though it seems like it would be nice to re-use the existing one, a key difference as pointed out by @tvararu is that this new links to a `session`, and not a `patient` or `campaign`. To change the existing `Consent` to accommodate this would be a distraction.
- Create a `ConsentFormController` with `start`, `confirm`, `create` and `record` actions. We'll use sub-controllers for the remaining steps of the journey, which will follow similar patterns with actions such as `update`, `edit`, `*_params`, etc.
- Use the route `/sessions/:session_id/consent_forms/` as a starting the base for this controller, since it makes sense to have the session identifier.
- Use `@consent_form` instead of `@draft_consent_form`. For (slightly more) brevity and because, really, most of these controller actions are agnostic as to whether consent form is draft or not.


app/models/consent_form.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+#  fk_rails_...  (session_id => sessions.id)
+#
+
+class ConsentForm < ApplicationRecord
+  audited
+
+  belongs_to :session
+
+  enum :route, %i[website phone paper in_person self_consent], prefix: "via"
```
tvararu, line 26: These can't be `self_consent` or `in_person`, I think?

app/views/consent_forms/confirm.html.erb
```diff
@@ -0,0 +1,12 @@
+<% content_for :before_main do %>
+  <%= render AppBacklinkComponent.new(
+    href: url_for(action: :start),
+    name: "start page",
+  ) %>
+<% end %>
```
tvararu, line 6: One thing we've been bad at doing is adding different page titles, which is good for accessibility:

```suggestion
<% end %>

<%= content_for :page_title, "Check and confirm answers" %>
```

tvararu (approved):

app/models/consent_form.rb
tvararu: Arguably `website` is the only valid option here.

misaka: As discussed, `route` is removed.

app/views/consent_forms/confirm.html.erb
misaka: As discussed, this will be addressed in a separate PR.

PR #458, opened 2023-08-24
Populate page titles based on h1 content

tvararu:
This introduces a h1 helper that automatically sets `content_for(:page_title)`. Currently none of our pages have custom titles, which is an accessibility issue. Every page must also have one (and preferably only one) h1 element, and in most cases the content of the custom title prefix and the h1 will coincide.

TODO:

- [x] Investigate automatically setting an `Error:` prefix when the page status is a 422
- [x] Update every template to use the h1 helper
  - [x] For the pages with simple forms that use an h1 as the legend, set the page_title using the manual way

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,0 +1,30 @@
+module ApplicationHelper
+  def h1(text = nil, **options, &block)
+    title_text = options[:page_title] || text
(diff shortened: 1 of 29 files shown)


misaka (approved):
Asked elsewhere ... why do some H1s have `nhsuk-heading-l` class and some don't?

PR #460, opened 2023-08-24
Add child name form

misaka:
This adds the edit page for the child's name to the parental consent journey.

image

With the corresponding confirm page and working "Change" button:

image

Still missing:
- [x] Back button should go to the confirm page when editing the name.
- [ ] There is a line under the "Change" link on the confirm page, dunno where it's coming from, but keeeeil eit!
- [ ] Maybe e2e testing for validations and the back button? Not sure how to tests these outside of e2e tests though.


misaka:
I think this is ready for review. I might be tempted to add tests to check validations and the back button ...

tvararu (approved):
:+1: We can fix the weird underline and the bold label in a follow-up

PR #464, opened 2023-08-25
Add dob step to consent form

misaka:
Includes validations and specs.

image

Things to do from the review just now:
- [x] Rename to `date_of_birth`
- [x] Fix the `update` action in this step and the `name` step to use `respond_to` if available. This means that if editing from the confirm page after updating the dob/name we'll be sent back to the confirm page.
- [ ] Address deprecation warnings caused by custom default format used in errors.
- [ ] [Optional] When an missing date field is entered keep the other ones that are present and only report that one is blank.
- [ ] [Optional] Implement date validation that only checks the year and not the specific date.


db/migrate/20230825085229_add_dob_to_consent_form.rb
```diff
@@ -0,0 +1,7 @@
+class AddDobToConsentForm < ActiveRecord::Migration[7.0]
+  def change
+    change_table :consent_forms, bulk: true do |t|
+      t.date :dob
```
tvararu, line 4: Should we name it `date_of_birth`?

tvararu (approved):
Code LGTM, just testing it in review app but pre-approving

tvararu:
:bug: Submitting a form with partially filled in fields does not persist the content of the fields that have been filled in

image

image

tvararu:
This date format isn't the most intuitive, and we probably don't need to be so specific about the date to the exact month and day...

image

db/migrate/20230825085229_add_dob_to_consent_form.rb
misaka: Ya, actually, I kind of prefer that.

misaka:
> 🐛 Bug: Submitting a form with partially filled in fields does not persist the content of the fields that have been filled in
> 
> image
> 
> image

I believe your response to this when I mentioned it was "That path leads to a world of pain", or a reasonable facsimile of such. :P

tvararu:
Currently if you reach the confirm page, and go back to change the name, after submitting the name form you go to the date of birth page. Even though it's already been filled out. That's probably something we should change with some sort of wizard

misaka:
> This date format isn't the most intuitive, and we probably don't need to be so specific about the date to the exact month and day...
> 
> image

We could tweak it to only check the years, ya ... it'll be a bit more vague, but do we really gain a lot by that? I agree that the date isn't the most intuitive ... but again, is it worth the effort?

misaka:
> Currently if you reach the confirm page, and go back to change the name, after submitting the name form you go to the date of birth page. Even though it's already been filled out. That's probably something we should change with some sort of wizard

👍  There's the `return_to` parameter that could be included via a hidden field.

misaka:
I fixed the formatting of the date by setting the default date format, but that's causing deprecation errors. We could leave it but tbh, we'll have to deal with it and if we adopt the validation module from Apply for teacher training that may sort this as well as other errors.

image

image

misaka:
I'm going to back-out the default date format change because of the deprecation warnings, I'm not sure what the behaviour will be in production and we'll address, and the other issues, in follow-up PRs.

tvararu (approved):

PR #465, opened 2023-08-25
Override service name in consent journeys

tvararu:
This changes from using the `t('service.name')` translation string to a `@service_name` instance variable that's set by `ApplicationController` and can be overridden by other controllers that inherit from it.

It introduces a new `ConsentForms::BaseController` that can be used for this purpose.

This PR also includes a couple of other small tweaks, see commits.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,6 +1,5 @@
 en:
   service:
-    name: Manage vaccinations for school-aged children
     email: my-service@email
   activerecord:
     attributes:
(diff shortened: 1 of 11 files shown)


config/locales/en.yml
```diff
@@ -86,7 +85,7 @@ en:
             last_name:
               blank: Enter a last name
             use_common_name:
-              inclusion: Tell us whether they use a different name
```
misaka, line 89: Emma recommended that we use "Tell us whether ..." here.

misaka (approved):
🚀

PR #473, opened 2023-09-01
Add DateParamsValidator for date_of_birth

misaka:
The approach is taken partially from DfE's FindALostTRN, slightly tweaked here as we don't have form objects and want to reuse model validations.

In this approach we only validate the form inputs, but not the actual parameters required by the date of birth attribute on consent forms. This should make this library usable on other models.


tvararu (approved):
:motor_boat:

PR #482, opened 2023-09-05
Define steps in ConsentForm

tvararu:
Replace the `on:` context based validation with conditional validation that relies on specifying a `form_step`. This also makes the validations fire off for previous steps, and at the end for the final `record` step.

Unifies the `*_params` methods in the EditController into one `update_params` method.

`required_for_step?` takes an optional `exact` parameter, for transient steps like the `school` one that ask a question but don't actually save a field to the database (and shouldn't be re-validated on other steps).

Overall approach based on:
https://www.joshmcarthur.com/2014/12/23/rails-multistep-forms.html

### DIFF
--- a/app/controllers/consent_forms_controller.rb
+++ b/app/controllers/consent_forms_controller.rb
@@ -8,7 +8,8 @@ class ConsentFormsController < ConsentForms::BaseController
   end
 
   def create
-    consent_form = @session.consent_forms.create!
+    consent_form = @session.consent_forms.new
+    consent_form.save!(validate: false)
     redirect_to session_consent_form_edit_path(@session, consent_form, :name)
   end
 
--- a/app/controllers/consent_forms_controller.rb
+++ b/app/controllers/consent_forms_controller.rb
@@ -8,8 +8,7 @@ class ConsentFormsController < ConsentForms::BaseController
   end
 
   def create
-    consent_form = @session.consent_forms.new
-    consent_form.save!(validate: false)
+    consent_form = @session.consent_forms.create!
     redirect_to session_consent_form_edit_path(@session, consent_form, :name)
   end
 
--- a/app/models/consent_form.rb
+++ b/app/models/consent_form.rb
@@ -33,29 +33,31 @@ class ConsentForm < ApplicationRecord
 
   belongs_to :session
 
-  with_options if: -> { required_for_step?(:name) } do
-    validates :first_name, presence: true
-    validates :last_name, presence: true
-    validates :use_common_name, inclusion: { in: [true, false] }
-    validates :common_name, presence: true, if: :use_common_name?
-  end
+  with_options on: :update do
+    with_options if: -> { required_for_step?(:name) } do
+      validates :first_name, presence: true
+      validates :last_name, presence: true
+      validates :use_common_name, inclusion: { in: [true, false] }
+      validates :common_name, presence: true, if: :use_common_name?
+    end
 
-  with_options if: -> { required_for_step?(:date_of_birth) } do
-    validates :date_of_birth,
-              presence: true,
-              comparison: {
-                less_than: Time.zone.today,
-                greater_than_or_equal_to: 22.years.ago.to_date,
-                less_than_or_equal_to: 3.years.ago.to_date
-              }
-  end
+    with_options if: -> { required_for_step?(:date_of_birth) } do
+      validates :date_of_birth,
+                presence: true,
+                comparison: {
+                  less_than: Time.zone.today,
+                  greater_than_or_equal_to: 22.years.ago.to_date,
+                  less_than_or_equal_to: 3.years.ago.to_date
+                }
+    end
 
-  with_options if: -> { required_for_step?(:school, exact: true) } do
-    validates :is_this_their_school,
-              presence: true,
-              inclusion: {
-                in: %w[yes no]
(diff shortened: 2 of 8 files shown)


app/models/consent_form.rb
```diff
@@ -23,34 +23,57 @@
 #
 
 class ConsentForm < ApplicationRecord
-  attr_accessor :is_this_their_school
+  cattr_accessor :form_steps do
+    %i[name date_of_birth school]
+  end
```
tvararu, line 28: This is the new source of truth for consent form steps; I like that it's in the same place as the validations.

app/controllers/consent_forms_controller.rb
```diff
@@ -8,7 +8,8 @@ def start
   end
 
   def create
-    consent_form = @session.consent_forms.create!
+    consent_form = @session.consent_forms.new
+    consent_form.save!(validate: false)
```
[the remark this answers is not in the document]
misaka: Have you tried adding `on: :update` to the validations? It sounds like that should work.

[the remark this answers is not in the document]
tvararu: Will give that a go

app/models/consent_form.rb
```diff
@@ -23,34 +23,57 @@
...
+    return false if exact && form_step != step
+
+    # All fields are required if no form_step is set
+    return true if form_step.nil?
+
+    # Otherwise, all fields from previous and current steps are required
+    return true if form_steps.index(step) <= form_steps.index(form_step)
+
+    false
+  end
```
misaka, line 78: I found this a little difficult to read, sorry to say, although maybe that was just the fog left-over from the meeting earlier. 

1. I think `required_for_step?(step, ...)` gives the false impression that `step` is what's being tested, when in fact it's the form step that's being tested, hence the name change. 
2. With the return statements written the way the are, I was having a hard time understanding what `exact` did, er ... exactly. Testing it separately makes it clearer, although I'm not sure if Rubocop might not try to golf that down for us.

```suggestion
  def form_step_matches?(step, exact: false)
    # All fields are required if no form_step is set
    return true if form_step.nil?

    if exact
      # Exact means that the form_step must match the step
      return form_step == step
    else
      # Otherwise, all fields from previous and current steps are required
      return form_steps.index(step) <= form_steps.index(form_step)
  end
```

misaka (commented):
I feel like the methods in the model. at least, should have a spec. Can we (and should we) also test that the right validations are triggered at the right time, to test stuff like the `school` step having `exact` set?

misaka:
Sorry, I should say that those comments aside, I really like this change ... stuff like sticking the form steps into the model is good, even if at first glance the form step s are a controller concern ... the fact that they are here with the step validations is good.

spec/models/consent_form_spec.rb
```diff
@@ -0,0 +1,67 @@
+require "rails_helper"
+
+RSpec.describe ConsentForm, type: :model do
+  describe "Validations" do
+    let(:cf) { build(:consent_form) }
+
+    context "when form_step is :name" do
+      before { cf.form_step = :name }
+
+      it do
```
misaka, line 10: Missing description here. Or are these expressible as as a couple one-liners?

misaka: Actually, these would work well as one-liners ...

```
    subject(:cf) { build(:consent_form) }

    context "when form_step is :name" do
      before { cf.form_step = :name }

      it { should validate_presence_of(:first_name).on(:update) }
      it { should validate_presence_of(:last_name).on(:update) }
    end
```

tvararu: I wrote it like that but one of the linters re-wrote it, lol...

spec/models/consent_form_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        expect(cf).to validate_presence_of(:last_name).on(:update)
+      end
+
+      it do
+        expect(cf).to validate_presence_of(:date_of_birth).on(:update)
+        # expect(cf).to validate_comparison_of(:date_of_birth)
+        #   .is_less_than(Time.zone.today)
+        #   .is_greater_than_or_equal_to(22.years.ago.to_date)
+        #   .is_less_than_or_equal_to(3.years.ago.to_date)
+        #   .on(:update)
```
tvararu, line 36: I'm not sure why this doesn't work

```
NoMethodError:
       undefined method `validate_comparison_of' for #<RSpec::ExampleGroups::ConsentForm::Validations::WhenFormStepIsDateOfBirth "example at ./spec/models/consent_form_spec.rb:30">
```

spec/models/consent_form_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        it { should validate_presence_of(:first_name).on(:update) }
+        it { should validate_presence_of(:date_of_birth).on(:update) }
+      end
+
+      it { should validate_presence_of(:is_this_their_school).on(:update) }
+      it do
+        should validate_inclusion_of(:is_this_their_school).in_array(
+                 %w[yes no]
+               ).on(:update)
+      end
```
tvararu, line 56: This is the linter's idea, not mine

misaka (approved):
Nice. `required_for_step` still makes me itch, but :shipit:

PR #483, opened 2023-09-05
Add parent page to Consent Form journey 

tvararu:
Depends on https://github.com/nhsuk/record-childrens-vaccinations/pull/482.

This adds the "About you" parent details page from the prototype. Also updates the Avo resource with the newly added fields.

### Screenshot with validation errors

image

### Confirmation page

image


app/views/consent_forms/confirm.html.erb
```diff
@@ -53,4 +53,46 @@
...
+      end
+
+      summary_list.with_row do |row|
+        row.with_key { "Phone number" }
+        row.with_value { @consent_form.parent_phone }
+        row.with_action(text: "Change",
+          href: session_consent_form_edit_path(@session, @consent_form, :parent),
+          visually_hidden_text: "your phone")
+      end
+    end %>
```
tvararu, line 94: This isn't the exhaustive list; notably the `_other` stuff isn't visible. I can add it now, or we can come back to it later, the prototype isn't up to date on how this should look in every permutation.

misaka: Is this something that we'll easily catch in a snagging session, or should we card it up?

misaka (approved):

app/views/consent_forms/confirm.html.erb
tvararu: Let's leave it for later...

PR #484, opened 2023-09-06
[WIP] Add consent page to Consent Form journey

tvararu:


app/models/consent_form.rb
```diff
@@ -42,6 +43,7 @@ class ConsentForm < ApplicationRecord
 
   enum :parent_relationship, %w[mother father guardian other], prefix: true
   enum :contact_method, %w[text voice other], prefix: true
+  enum :response, %w[given refused not_provided], prefix: true
```
misaka, line 46: `response_given`, `response_refused` ... these don't quite sound right. Even with a consent_form instance: `consent_form.response_given` kind of sounds like the responded. What do you think of the `consent` prefix: `consent_form.consent_given`, `consent_form.consent_refused` ... `consent_form.consent_not_provided`?

I can't decide it this is going back to the `consent_request.consent_response_given`, or whatever it was, situation we had.

misaka (approved):

PR #485, opened 2023-09-07
Add consent refused reason page

tvararu:
Depends on https://github.com/nhsuk/record-childrens-vaccinations/pull/484.

Review commit by commit.

### Screenshot

image


app/controllers/consent_forms/edit_controller.rb
```diff
@@ -73,6 +77,15 @@ def set_consent_form
     @consent_form = ConsentForm.find(params.fetch(:consent_form_id))
   end
 
+  def set_steps
+    # Translated steps are cached after running setup_wizard_translated.
+    # To allow us to run this method multiple times during a single action
+    # lifecycle, we need to clear the cache.
+    @wizard_translations = nil
```
tvararu, line 84: The reason we need this is because the library is setting `@wizard_translation ||=`: https://github.com/zombocom/wicked/blob/7037c81780cac6eae0ed2d51560eafa91651e5e4/lib/wicked/wizard/translated.rb#L28

During an `update` action, the `before_action`s get triggered, setting the form steps to an array of `form_steps` that don't contain `reason`.

After `@consent_form.assign_attributes(update_params)` in the update action, the `form_steps` are updated on the instance, but the internal state of the Wizard is still operating on the old `form_steps`. Setting `self.steps` isn't enough, we also need to trigger `setup_wizard_translated` again and in addition force the updating of the `@wizard_translations` array by setting it to `nil` before.

There shouldn't be much downside in re-running these operations twice for update actions, other than the fact that it's not super clean.

app/models/consent_form.rb
```diff
@@ -101,6 +121,9 @@ def required_for_step?(step, exact: false)
     # All fields are required if no form_step is set
     return true if form_step.nil?
 
+    # Step can't be required if it's not in the current form_steps list
+    return false unless step.in?(form_steps)
+
     # Otherwise, all fields from previous and current steps are required
     return true if form_steps.index(step) <= form_steps.index(form_step)
```
tvararu, line 128: I'd like to find a better way to express the logic in this function. :thinking:

app/controllers/consent_forms/edit_controller.rb
```diff
@@ -4,10 +4,10 @@ class ConsentForms::EditController < ConsentForms::BaseController
 
   layout "two_thirds"
 
-  steps(*ConsentForm.form_steps)
-
   before_action :set_session
   before_action :set_consent_form
+  before_action :set_steps # Uses @consent_form, needs set_consent_form
```
misaka, line 9: Could we exclude `update` from this action since we have to run `set_steps` manually there anyway? And would that obviate the need for `@wizard_translations = nil`?

```suggestion
  before_action :set_steps, except: :update # Uses @consent_form, needs set_consent_form
```

misaka (approved):
This is good, it's not terribly hacky although if we can add the `except: :update` and as a result remove `@wizard_translations = nil` that would be even better.

app/controllers/consent_forms/edit_controller.rb
tvararu: Great idea, I'll try it and report back.

PR #492, opened 2023-09-11
Sign in and out

misaka:
Adds a sign-in page that (mostly) matches the prototype. The "I've forgotten my password" link isn't included because we don't have email integration yet and that doesn't work.

image

It includes fixes up to the banners to support the login success notification.

image

It also adds a nav menu with the email of the currently sign-in user and a link to "Sign-out".

image

The behaviour of the "Sign-out" link might need a little bit of work, interested in suggestions for improvement.

Also, here's how the menu looks on mobile displays, showing the "Sign-out" item with the full width highlighting:

image

Which matches the background highlighting on the "Home" menu item:

image


app/assets/stylesheets/application.sass.scss
```diff
@@ -81,3 +93,40 @@ $color_app-pale-blue: #ccdff1;
...
+    right: nhsuk-spacing(1);
+    top: 11px;
+  }
+}
+
+.nhsuk-header__menu-toggle {
+  @include mq($until: desktop) {
+    right: 0px;
+  }
+}
```
tvararu, line 123: Instead of an override, we could make this new modifier class specific to our app:

```suggestion
.app-header__menu-toggle--no-search {
  @include mq($until: desktop) {
    right: 0px;
  }
}
```

Then we can add it to the menu element in the header. Might have to tweak the rule if it [isn't specific enough](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity).

tvararu (approved):
:shipit:

app/assets/stylesheets/application.sass.scss
misaka: Done

PR #497, opened 2023-09-13
Add GP page to parental consent journey

tvararu:
Commits are broken into 3 steps I'll probably follow for the next few pages that should all be formulaic:

- Commit 1: Migrations, enum declaration, avo fields
- Commit 2: Validations, tests, `form_steps` update, validation messages
- Commit 3: Views, controller param, and e2e spec update

### After

image

image


app/views/consent_forms/edit/contact_method.html.erb
```diff
@@ -5,7 +5,7 @@
   ) %>
 <% end %>
 
-<% title "Phone contact method" %>
+<% title = "Phone contact method" %>
```
tvararu, line 8: Missed this but somehow it still worked :raised_eyebrow:

PR #499, opened 2023-09-13
Add health questions to parental consent journey

tvararu:
See commits.

### After

image

image


db/migrate/20230913152019_add_questions_to_consent_form.rb
```diff
@@ -0,0 +1,10 @@
+class AddQuestionsToConsentForm < ActiveRecord::Migration[7.0]
+  def change
+    add_column :consent_forms,
+               :health_questions,
+               :jsonb,
+               null: false,
+               default: {
+               }
```
misaka, line 8: In the controller you're using a number to access the answer here, e.g. `1`. Should this be a `[]`? Or are you future proofing?

misaka:
```ruby
      health_questions[current_health_question_index].merge! health_question_params
```

https://github.com/nhsuk/record-childrens-vaccinations/pull/499/files#diff-e0e1a4b64eb811d449a4d96f396f2136a5457ddcc8006c8551831caf75ac36fbR25

tvararu: You're right, it should be `[]`

app/models/consent_form.rb
```diff
@@ -61,6 +62,19 @@ class ConsentForm < ApplicationRecord
        prefix: "refused_because"
   enum :gp_response, %w[yes no dont_know], prefix: true
 
+  serialize :health_questions, HealthQuestion::ArraySerializer
+
+  HEALTH_QUESTIONS = [
```
tvararu, line 67: These should come from the database, not a fixture in this model

app/models/consent_form.rb
```diff
@@ -142,7 +156,8 @@ def form_steps
       (:reason if consent_refused?),
       (:injection if consent_refused? && eligible_for_injection?),
       (:gp if consent_given?),
-      (:address if consent_given?)
+      (:address if consent_given?),
+      (:health_1 if consent_given?)
```
tvararu, line 160: Need to extend this up to `:health_8`

app/models/health_question.rb
```diff
@@ -19,4 +19,24 @@
 class HealthQuestion < ApplicationRecord
   attr_accessor :response, :notes
   belongs_to :vaccine
+
```
tvararu, line 22: We should create a separate class that isn't backed by AR, possibly `HealthAnswer`

config/locales/en.yml
```diff
@@ -16,6 +16,7 @@ en:
     contact_method: "contact-method"
     gp: "gp"
     address: "address"
+    health_1: "health-1"
```
tvararu, line 19: Need to extend this to `health_8:`

app/controllers/consent_forms/edit_controller.rb
```diff
@@ -11,25 +11,37 @@ class ConsentForms::EditController < ConsentForms::BaseController
...
-                      )
-        )
-      end
+    if current_step == :school && @consent_form.is_this_their_school == "no"
+      return(
+        redirect_to session_consent_form_cannot_consent_path(
+                      @session,
+                      @consent_form
+                    )
+      )
```
tvararu, line 44: This isn't related, but I don't think we'll end up using a `case` statement here so I simplified it.

app/models/consent_form.rb
tvararu: Descoped for a future PR, to keep this one smaller

app/models/consent_form.rb
```diff
@@ -142,7 +156,15 @@ def form_steps
...
-      (:address if consent_given?)
+      (:address if consent_given?),
+      (:health_1 if consent_given?),
+      (:health_2 if consent_given?),
+      (:health_3 if consent_given?),
+      (:health_4 if consent_given?),
+      (:health_5 if consent_given?),
+      (:health_6 if consent_given?),
+      (:health_7 if consent_given?),
+      (:health_8 if consent_given?)
```
tvararu, line 167: Played with the idea of golfing this down but decided this explicit setup looks fine...

app/views/consent_forms/confirm.html.erb
```diff
@@ -152,6 +154,36 @@
...
         end
+      else
+        summary_list.with_row do |row|
+          row.with_key { "Phone number" }
+          row.with_value { "Not provided" }
+          row.with_action(text: "Change",
+            href: session_consent_form_edit_path(@session, @consent_form, :parent),
+            visually_hidden_text: "your phone")
+        end
+      end
```
tvararu, line 165: This is unrelated to this PR, just aligning to the prototype in another small way.

app/models/consent_form.rb
misaka: I spent about 1 min on it and came to a similar conclusion. :P

config/locales/en.yml
misaka: I'm surprised to see that these, and other translations in this section, are necessary. Is the default behaviour not to `dasherize` these? Where are they used?

PR #500, opened 2023-09-15
Authentication

misaka:
**IMPORTANT**

Merging this change make all environments inaccessible without signing-in, which will require a user be added to the db. This can be done either by reloading the example data (using `/reset`) or by creating a user in the console. Authorisation hasn't been implemented yet so users don't need to be added to a team or to a session.

----

Add authentication requirement to the service. This will require that users sign-in before accessing most pages. These pages will redirect users to the sign-in page, which will then redirect them back to the page that sent them on successful sign-in.

Exempt pages are:

- `/start` (redirected-to from `/`) -- clicking "Start now" on this page will direct users to the dashboard, which will redirect them to sign-in
- `/reset` -- this is required to bootstrap environments with users (it creates `nurse@dev` in `development`, `nurse@test` in `test`, etc)
- `/csrf` -- this action was used to support offline journeys, and although we're disabling that feature I've skipped authentication here just to allow specs to pass (up for debate, but as we don't know how offline will work this seemed the most expedient approach)
- `/sessions/:session_id/consents` and the entire parental consent form journey -- these will use a different authentication mechanism, if any


misaka:
Hrm. The `rspec` tests require that a user be created as part of the test. Will fix.

tests/consent_during_vaccination.spec.ts
```diff
@@ -22,6 +24,11 @@ async function given_the_app_is_setup() {
   await p.goto("/reset");
 }
 
+async function and_i_am_signed_in() {
+  await p.goto("/users/sign_in");
+  signInTestUser(p);
```
tvararu, line 29: I think these should all be awaits:

```suggestion
  await signInTestUser(p);
```

tests/sign_in.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  await then_i_should_see_the_dashboard();
+  await and_i_should_see_a_banner_saying_i_am_signed_in();
+
+  // Sign-out
+  await when_i_sign_out();
+  await then_i_should_see_the_start_page();
+  await and_i_should_see_a_banner_saying_i_am_signed_out();
+
+  // Saves and redirects the user to the page they were trying to visit before
+  // they were rudely imterrupted by the sign-in page.
```
tvararu, line 23:
```suggestion
  // they were rudely interrupted by the sign-in page.
```

tvararu (approved):
Nicely done :ship: 

Keeping basic auth around in addition is good for now.

tvararu (approved):

PR #519, opened 2023-09-21
Skip to confirmation page if param is set

tvararu:
Depends on https://github.com/nhsuk/record-childrens-vaccinations/pull/518.

This sets `skip_to_confirm=true` as a parameter on most of the change links from the confirmation page. If this is set, we then redirect the user to the confirmation page after they submit the step.

See commits.

### DIFF
--- a/app/views/consent_forms/confirm.html.erb
+++ b/app/views/consent_forms/confirm.html.erb
@@ -6,7 +6,7 @@
 <% end %>
 
 <% change_link = Proc.new do |attr|
-  options = {}
+  options = attr.in?(%i[consent reason]) ? {} : { skip_to_confirm: true }
   session_consent_form_edit_path(
     @session,
     @consent_form,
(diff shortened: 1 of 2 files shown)


app/views/consent_forms/confirm.html.erb
```diff
@@ -5,6 +5,16 @@
...
 
+<% change_link = Proc.new do |attr|
+  options = attr.in?(%i[consent reason]) ? {} : { skip_to_confirm: true }
+  session_consent_form_edit_path(
+    @session,
+    @consent_form,
+    attr.to_s.dasherize,
+    options
+  )
+end %>
```
misaka, line 16: Did `def` not work here?

misaka (approved):
Nice, I like this mechanism, just wondering if `change_link` shouldn't be in a helper too, but I would guess there's a reason it's done this way.

PR #527, opened 2023-09-28
Add example campaign generator

misaka:
This adds an example campaign generator rake task based on the existing `generate_model_office_data` one. It's called `generate_example_campaign` and the key changes / features are:

- Allow generation of `flu` data. This isn't quite complete, the flu health questions have to be added still.
- Allow loading campaigns as a new campaign, without overwriting existing ones. This is a change in the `load_campaign_example` rake task.
- Generate a team and user with the campaign.
- Spec to test that the generated example campaign still works. **NB:** This could cause test breakage in the future as it compares the generated example campaign with a static fixture checked into the repo, but if we can keep this up-to-date we'll always have working example campaign generation.
- Options to control the number of patients to generate in different states on the cmdline. Use `rails -D generate_example_campaign` for more info.
- Allow specifying a seed for random number generation for repeatable campaign generation. Useful for testing purposes.
- Presets to replicate the model office data generation or to add others in the future.

Remaining tasks:
- [ ] Add flu health questions. I think these should be part of another PR as this one is already Big Enough. Also, we define the health questions in more than one place in the code so we it may take a little more work. 🥾 🥫


README.md
```diff
@@ -183,6 +168,57 @@ To [generate tests interactively by clicking in a live browser](https://playwrig
...
+# Specify a file to load
+$ rake load_campaign_example[db/sample_data/model-office.json]
+
+# Specify that a new campaign should be created instead of using reusing existing
+$ rake load_campaign_example[db/sample_data/model-office.json] new_campaign=1
+```
+
+The importer will `find_or_create` the records by default, using specific attributes to match records:
+
+- **campains** -- a combination of `type`, `location` and `date`
```
benilovj, line 190:
```suggestion
- **campaigns** -- a combination of `type`, `location` and `date`
```

tvararu (approved):

PR #532, opened 2023-10-04
More authorisationz

misaka:
**NB: This PR depends on #533**

Add authorisation to triage and vaccination controllers. Users will only be able to see the patients that are in campaigns that their team are running.

In addition to adding the policies, a lot of the work was uplifting the testing environment:
- Allow error pages to be rendered in the test environment
- Allow usernames to be specified when generating example campaigns (to support consistent usernames in tests)
- Add example flu campaign and load it in `/reset`
- And, of course, adding tests for the triage and vaccinations pages


tvararu (approved):
:boom:

PR #533, opened 2023-10-04
Load example campaign

misaka:
Add `LoadExampleCampaign` and use it from the `load_example_campaign` Rake task and from `/reset`.

This was motivated by strange behaviour when using `/reset` whereby for each time `/reset` was triggered, the action would trigger the task n+1 times. This wasn't noticeable when we the HPV campaign was being created with `find_or_create`, but with the new tests where we've added another campaign which simply uses `create`, we were getting many copies of the same campaign and related data.

### DIFF
--- a/lib/tasks/load_campaign_example.rake
+++ b/lib/tasks/load_campaign_example.rake
@@ -1,110 +1,18 @@
-require "example_campaign_data"
+require "load_example_campaign"
 
 desc "Load campaign example file into db"
-task :load_campaign_example, [:example_file] => :environment do |_task, args|
-  new_campaign = ENV.fetch("new_campaign", false).in? %w[true 1 yes]
+task :load_campaign_example,
+     %i[example_file new_campaign] => :environment do |_task, args|
+  new_campaign =
+    args.fetch(:new_campaign) { ENV.fetch("new_campaign", false) }.in? [
+            true,
+            "true",
+            "1",
+            "yes"
+          ]
 
   example_file =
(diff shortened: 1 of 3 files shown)


lib/tasks/load_campaign_example.rake
```diff
@@ -1,110 +1,18 @@
-require "example_campaign_data"
+require "load_example_campaign"
 
 desc "Load campaign example file into db"
-task :load_campaign_example, [:example_file] => :environment do |_task, args|
-  new_campaign = ENV.fetch("new_campaign", false).in? %w[true 1 yes]
+task :load_campaign_example,
+     %i[example_file new_campaign] => :environment do |_task, args|
+  new_campaign =
```
misaka, line 6: Strictly speaking this bit is a stowaway from #532, but it's not causing harm and isn't worth the time to be split off, unless it turns out to be controversial.

lib/tasks/load_campaign_example.rake
```diff
@@ -1,110 +1,18 @@
...
+    args.fetch(:new_campaign) { ENV.fetch("new_campaign", false) }.in? [
+            true,
+            "true",
+            "1",
+            "yes"
+          ]
```
misaka, line 12: Strictly speaking this bit is a stowaway from #532 , but it isn't causing harm and isn't worth splitting off unless it turns out to be controversial.

tvararu (approved):

PR #534, opened 2023-10-05
Retrieve health questions from database

tvararu:
Before this, we were hard-coding the array of health questions for the consent form in the consent form model itself.

Also contains 2 small unrelated commits.


tests/parental_consent.spec.ts
```diff
@@ -56,7 +56,7 @@ async function given_the_app_is_setup() {
 }
 
 async function when_i_go_to_the_consent_start_page() {
-  await p.goto("/sessions/1/consents/start");
+  await p.goto("/sessions/2/consents/start");
```
tvararu, line 59: 2nd campaign is flu in the sample data which is what we're testing for here (we check that the first question is what we expect)

misaka (approved):
🚀

PR #535, opened 2023-10-06
Add hint text to health questions

tvararu:
The field needs to live in the database, on the model, so that it can be edited by SAIS teams.

Because the questions are saved when the ConsentForm is created, we also need to persist them in the jsonb field containing all the health question metadata. We could change in the future to persisting just an ID and
then fetching the latest information from the database, but this could backfire if SAIS teams end up reusing the same question and completely changing it instead of adding a new one. We'd have to be careful about how we design the "update" action to disallow editing the question after it's already been used.

image


misaka (changes requested):
I think the newly minted `ExampleCampaignGenerator` should be updated to add the hints to the appropriate questions, since that's what generated the `example-flu-campaign.json` file updated in this PR.

In fact, I think we should change the spec to test the generated output against that file (`example-flu-campaign.json`) to ensure that the generator works.

misaka (approved):
🦑

PR #536, opened 2023-10-06
Configure good_job

tvararu:
good_job is a Postgres-based ActiveJob backend for Rails.

Unlike Sidekiq, it doesn't require a separate Redis instance, and can run in the main web server process instead of a separate worker. This means that a good_job deployment is easier to think about (1 server instead of 3) and costs less.

I've also set `heroku config:set GOOD_JOB_EXECUTION_MODE=async` to configure the application to run with the async job queue in Production.


config/routes.rb
```diff
@@ -10,6 +10,7 @@
   root to: redirect("/start")
 
   mount Avo::Engine, at: Avo.configuration.root_path
+  mount GoodJob::Engine => "/good-job"
```
tvararu, line 13: This probably needs proper auth, and Avo as well should be accessible only with an admin user.

misaka (approved):

PR #553, opened 2023-10-17
Asthma health questions

misaka:
This converts the health questions to a linked-list so that we can add non-linear / branching journeys through the questions.


app/models/health_question.rb
```diff
@@ -20,5 +20,17 @@
 #
 class HealthQuestion < ApplicationRecord
   attr_accessor :response, :notes
+  jsonb_accessor :metadata,
+                 next_question: :string
+
   belongs_to :vaccine
+
+  def self.first_health_question
+    id_set = ids - all.pluck(Arel.sql("metadata->>'next_question'")).map(&:to_i)
```
tvararu, line 28: :chefs-kiss:

app/controllers/consent_forms_controller.rb
```diff
@@ -8,12 +8,19 @@ def start
...
 
   def create
+    vaccine = @session.campaign.vaccines.first
+    first_health_question = vaccine.first_health_question
     health_answers =
-      @session.campaign.vaccines.first.health_questions.map do |hq|
-        HealthAnswer.new question: hq.question,
+      [
+        first_health_question,
+        *(vaccine.health_questions - [first_health_question])
```
tvararu, line 16: We could extract this as a `rest_of_health_questions` variable, might be a bit clearer what the intent is

Minor, feel free to ignore

tvararu: Alternatively it can be

```ruby
(vaccine.health_questions - [first_health_question]).unshift(first_health_question)
```

Not sure that's necessarily better though

tvararu (approved):

PR #554, opened 2023-10-18
Improve password security

tvararu:
Bump the minimum character requirement to 8 characters as per the [design system guidance](https://design-system.service.gov.uk/patterns/passwords/).

Integrate the [devise-pwned_password](https://github.com/michaelbanfield/devise-pwned_password) gem to check that user-submitted passwords haven't been found in any previous security breaches.

The gem has an implicit timeout of 5 seconds, and is also designed to work as a progressive enhancement, so it [won't block our users if the HaveIBeenPwned API goes down](https://github.com/michaelbanfield/devise-pwned_password#considerations):

> The gem is designed to silently swallows errors if the PwnedPasswords service is unavailable, allowing users to use compromised passwords during the time when it is unavailable.

image

### DIFF
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -27,6 +27,6 @@ FactoryBot.define do
     full_name { |n| "Test User #{n}" }
     email { |n| "test-#{n}@localhost" }
     teams { [Team.first || create(:team)] }
-    password { "rosebud" }
+    password { "rosebud1" }
   end
 end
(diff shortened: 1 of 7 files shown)


spec/factories/users.rb
```diff
@@ -27,6 +27,6 @@
     full_name { |n| "Test User #{n}" }
     email { |n| "test-#{n}@localhost" }
     teams { [Team.first || create(:team)] }
-    password { "rosebud" }
+    password { "rosebud1" }
```
misaka, line 30: security ftw!

PR #555, opened 2023-10-19
Validate postcodes

tvararu:
Add the `uk_postcode` gem and use it in the consent form.


spec/models/consent_form_spec.rb
```diff
@@ -224,7 +226,18 @@
...
+        consent_form =
+          build(
+            :consent_form,
+            address_postcode: "asdf",
+            form_step: :address,
+            response: "given"
+          )
+        expect(consent_form.valid?(:update)).to be_falsey
+        expect(consent_form.errors[:address_postcode]).to be_present
+      end
```
tvararu, line 240: We should add a test for the formatting

misaka (approved):

PR #556, opened 2023-10-20
Implement email and phone validator

tvararu:
This PR contains an email and phone validator that's implemented to a similar spec as the GOVUK Notify one. This reduces the chance that emails entered by our users will be rejected by Notify.

It follows their implementation and implements their test cases:

- https://github.com/alphagov/notifications-utils/blob/9311851c105237dc5d770c9f1c0a5ad4c2acb380/notifications_utils/recipients.py#L622
- https://github.com/alphagov/notifications-utils/blob/9311851c105237dc5d770c9f1c0a5ad4c2acb380/notifications_utils/__init__.py#L14
- https://github.com/alphagov/notifications-utils/blob/9311851c105237dc5d770c9f1c0a5ad4c2acb380/tests/test_recipient_validation.py#L118

For the email validator, International Domain Names (IDNs) are not currently implemented. The two test cases with non-UTF-8 characters are commented out.

image

### DIFF
--- a/spec/models/consent_form_spec.rb
+++ b/spec/models/consent_form_spec.rb
@@ -137,6 +137,9 @@ RSpec.describe ConsentForm, type: :model do
       it { should validate_presence_of(:parent_relationship).on(:update) }
       it { should validate_presence_of(:parent_email).on(:update) }
 
+      it { should_not allow_value("invalid").for(:parent_email).on(:update) }
+      it { should allow_value("foo@foo.com").for(:parent_email).on(:update) }
+
       context "when parent_relationship is 'other'" do
         let(:parent_relationship) { "other" }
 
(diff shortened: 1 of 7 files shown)


spec/models/consent_form_spec.rb
```diff
@@ -226,16 +231,8 @@
...
-          build(
-            :consent_form,
-            address_postcode: "asdf",
-            form_step: :address,
-            response: "given"
-          )
-        expect(consent_form.valid?(:update)).to be_falsey
-        expect(consent_form.errors[:address_postcode]).to be_present
+      it do
+        should_not allow_value("invalid").for(:address_postcode).on(:update)
```
misaka, line 235: I think we want to keep the postcode validations testing here, no?

tvararu: We are, but I changed the syntax, see first commit

PR #570, opened 2023-10-26
Asthma questions for reals

misaka:
Changes to the health questions to add branching for "follow-up" questions if certain questions have a "yes" response.

A couple of things still need considering / doing:

- [x] Add unit tests for `ConsentForm#each_health_answer` and `ConsentForm#health_answers_valid?`
- [ ] ~If a user changes, for example, the response to "Does your child have asthma" to "yes" from the confirmation page, they should have to answer the follow-up questions.~ punted
  - Additional question, should they always do the follow-up questions? What if they answer "yes" and answer the follow-up questions, then change the response to that question "no" (their child does not have asthma), and then back to "yes"? Should they do the follow-up questions again?
- [ ] ~Discussion: should we start counting health questions from 1? They currently start at 0 except for the hidden link texts on the confirmation page, which start them from 1.~ punted


tests/parental_consent_asthma_route.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  await p.getByLabel("Name of GP surgery").fill("Test GP Surgery");
+  await p.getByRole("button", { name: "Continue" }).click();
+
+  await p.getByLabel("Address line 1").fill("1 Test Street");
+  await p.getByLabel("Town or city").fill("Test Town");
+  await p.getByLabel("Postcode").fill("TE1 1ST");
+  await p.getByRole("button", { name: "Continue" }).click();
+
+  // await p.getByRole("radio", { name: "No" }).click();
+  // await p.getByRole("button", { name: "Continue" }).click();
```
tvararu, line 75: Do we need to leave these in?

app/models/consent_form.rb
```diff
@@ -161,6 +165,19 @@ def form_steps_for_health_questions
...
+  def each_health_answer
+    return if health_answers.empty?
+    return to_enum(:each_health_answer) unless block_given?
+
+    ha = health_answers.first
+    loop do
+      yield ha
+      next_health_answer_index = ha.next_health_answer_index
+      break unless next_health_answer_index
+      ha = health_answers[next_health_answer_index]
```
tvararu, line 177: I prefer an explicit variable name here, `ha` looks funny. :drum: 

Non-blocking suggestion:

```suggestion
    health_answer = health_answers.first
    loop do
      yield health_answer
      next_health_answer_index = health_answer.next_health_answer_index
      break unless next_health_answer_index
      health_answer = health_answers[next_health_answer_index]
```

tvararu: Also could this become an infinite loop?

tvararu (approved):
:shipit:

app/models/consent_form.rb
misaka: > Also could this become an infinite loop?

Ya, it could, in the same way a recursive function could infinitely recurse. We could implement something to prevent this, e.g.:

1. run some kind of pre-check, probably when the health answers are created, to check they don't do this
2. track which health answer has already been seen in this loop and bail out if a duplicate is found

These health answers are pretty static ... so the likelihood of this happening seems low, but it's true the impact could be quite high if it does happen. 🤔

PR #582, opened 2023-11-01
Fix test

tvararu:
This is currently broken on `main`.


misaka (approved):

tests/parental_consent_change_answers.spec.ts
```diff
@@ -101,6 +88,6 @@ async function when_i_click_the_confirm_button() {
 
 async function then_i_see_the_confirmation_page() {
   await expect(p.locator("h1")).toContainText(
-    "Joe Test will get their nasal flu vaccination at school",
+    "will get their nasal flu vaccination at school",
```
misaka, line 91: I was worried that this might be a thing, although I didn't expect it so soon ... we could add a param to `/random_consent_form` to specify a name, or other fields, if it makes our tests better.

PR #584, opened 2023-11-02
Fix back buttons

misaka:
This fixes the back buttons which were broken for pretty much all health question steps (since the recent change to how they work) and for the "change answers from the confirmation page" journey for the rest of the questions.

### DIFF
--- a/app/controllers/consent_forms/edit_controller.rb
+++ b/app/controllers/consent_forms/edit_controller.rb
@@ -8,6 +8,7 @@ class ConsentForms::EditController < ConsentForms::BaseController
   before_action :setup_wizard_translated
   before_action :validate_params, only: %i[update]
   before_action :set_health_answer, if: -> { is_health_question_step? }
+  before_action :set_follow_up_changes_start_page, only: %i[show]
 
   def show
     render_wizard
@@ -124,7 +125,9 @@ class ConsentForms::EditController < ConsentForms::BaseController
   end
 
   def skip_to_confirm?
-    request.referer.include?("skip_to_confirm")
+    request.referer.include?("skip_to_confirm") ||
+      params.key?(:skip_to_confirm) ||
+      session.key?(:follow_up_changes_start_page)
   end
 
   def next_health_question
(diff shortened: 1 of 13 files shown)


app/controllers/consent_forms/edit_controller.rb
```diff
@@ -142,14 +145,18 @@ def next_health_answer_missing_response?
...
       else
         jump_to Wicked::FINISH_STEP
       end
     elsif is_health_question_step? && next_health_question
       jump_to "health-question", question_number: next_health_question
     end
   end
+
+  def set_follow_up_changes_start_page
+    if params[:skip_to_confirm] && is_health_question_step?
```
tvararu, line 158: Is `skip_to_confirm?` a better source of truth here, or would this introduce bugs?

```suggestion
    if skip_to_confirm? && is_health_question_step?
```

app/helpers/consent_forms_helper.rb
```diff
@@ -28,4 +28,37 @@ def health_answer_response(health_answer)
...
+    end
+  end
+
+  def backlink_path
+    if params[:skip_to_confirm]
+      wizard_path(Wicked::FINISH_STEP)
+    else
+      previous_wizard_path
+    end
+  end
```
tvararu, line 63: Yeah, I like keeping this separate. It's clean and the complexity is isolated to the other one

tvararu (approved):
:clap:

app/helpers/consent_forms_helper.rb
misaka: 💯

app/controllers/consent_forms/edit_controller.rb
misaka: This won't work because we should only set `session[:follow_up_changes_start_page]` on the first health question that gets visited from the confirmation page. Lemme try to illustrate:

```
GET /sessions/2/consents/edit/confirm
# User clicks on the change link for the asthma question.
# This first GET is where "follow_up_changes_start_page" gets set
GET /sessions/2/consents/edit/health-question?question_number=0&skip_to_confirm=true
# User changes the answer to "Yes", causing follow-up questions to be viewed
PUT /sessions/2/consents/edit/health-question?question_number=0
# "follow_up_changes_start_page" should _not_ get set here
GET /sessions/2/consents/edit/health-question?question_number=1
```

If we make that suggested change above, then we risk setting `follow_up_changes_start_page` to "1" which would break the mechanism, unfortunately.

This does, point out though, that naming may be improved.

PR #592, opened 2023-11-09
Update consent form confirmation page when consent needs triage

tvararu:
When submitting a consent form, there are 4 possible cases:

- Confirmed (done)
- **Confirmed, but needs triage** :point_left: this PR
- Confirmed, using injection
- Refused

Each case has slightly different content on the confirmation page, and a different email gets sent out to the parent.

See commits.

#### Confirmation page

image

#### New email

image

### DIFF
--- a/app/controllers/consent_forms_controller.rb
+++ b/app/controllers/consent_forms_controller.rb
@@ -28,6 +28,10 @@ class ConsentFormsController < ConsentForms::BaseController
 
     session.delete(:consent_form_id)
 
-    ConsentFormMailer.confirmation(@consent_form).deliver_later
+    if @consent_form.any_health_answers_truthy?
+      ConsentFormMailer.confirmation_needs_triage(@consent_form).deliver_later
+    else
+      ConsentFormMailer.confirmation(@consent_form).deliver_later
+    end
   end
 end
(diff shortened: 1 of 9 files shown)


app/views/consent_forms/record.html.erb
```diff
@@ -1,3 +1,7 @@
-<%= render "confirmation_agreed" %>
+<% if @consent_form.any_health_answers_truthy? %>
```
misaka, line 1: This is really just about naming, but ... should the view really be deciding what constitutes needing triage? I think it feels like `@consent_form` should have a method `needs_triage?`, but which will essentially just be `any_health_answers_truthy?`

tvararu: Was thinking about that. We could add another method to the consent form to make this clearer.

app/controllers/consent_forms_controller.rb
```diff
@@ -28,6 +28,10 @@ def record
 
     session.delete(:consent_form_id)
 
-    ConsentFormMailer.confirmation(@consent_form).deliver_later
+    if @consent_form.any_health_answers_truthy?
```
misaka, line 31: Similar comment as the one on the view about tieing `any_health_answers_truthy?` to needing triage in the controller here.

app/views/consent_forms/record.html.erb
misaka: I think it would be ok if we just renamed the method in consent form, tbh, but either approach works for me.

misaka (approved):
Good work! The only issue mentioned isn't blocking.

PR #599, opened 2023-11-14
Add confirmation pages for offering injection and for refused consent

tvararu:
This adds two more mail templates and two more view templates for the remaining consent form final states. This covers the last two scenarios where users refuse consent and agree or disagree to be contacted with the offer of an injection instead.

### Screenshots of all scenarios

| | Confirmation page | Email |
|-|-|-|
| Accepted - All health questions "No" | Screenshot 2023-11-14 at 17 50 33 | Screenshot 2023-11-14 at 17 51 04 |
| Accepted - At least one health question is "Yes" | Screenshot 2023-11-14 at 17 51 37 | Screenshot 2023-11-14 at 17 51 49 |
| Refused - Agreed to be contacted for an injection | Screenshot 2023-11-14 at 17 52 14 | Screenshot 2023-11-14 at 17 52 26 |
| Refused - Disagreed to be contacted | Screenshot 2023-11-14 at 17 56 51 | Screenshot 2023-11-14 at 17 57 02 |

### DIFF
--- a/app/controllers/concerns/consent_form_mailer_concern.rb
+++ b/app/controllers/concerns/consent_form_mailer_concern.rb
@@ -1,0 +1,15 @@
+module ConsentFormMailerConcern
+  extend ActiveSupport::Concern
+
+  def send_record_mail(consent_form)
+    if consent_form.contact_injection?
+      ConsentFormMailer.confirmation_injection(consent_form).deliver_later
+    elsif consent_form.consent_refused?
+      ConsentFormMailer.confirmation_refused(consent_form).deliver_later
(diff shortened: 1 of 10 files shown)


misaka (commented):
The only way we have to test that the controller sends the email is a controller / request spec, is that right? Have you considered doing one just for this case?

misaka:
Or do we setup something that'll capture the API connection to GOVUK and expose it somewhere Playwright can inspect it. 🤔

misaka (approved):
🚀

PR #610, opened 2023-11-22
Display basic tabs in consents index page

tvararu:
This queries the list of patient_sessions and splits them into the respective categories that match the tabs in the design.

The tabs contain just the number of items in each category for now. The tables are coming in a follow-up PR, as componentised versions of the partials used for tables in the other sections.

This moves some of the `consent_given?` and associated methods out of the state machine concern and into the `patient_session` model.

### Screenshot

image

### DIFF
--- a/app/controllers/consents_controller.rb
+++ b/app/controllers/consents_controller.rb
@@ -1,7 +1,16 @@
 class ConsentsController < ApplicationController
   before_action :set_session
+  before_action :set_patient_sessions, only: %i[index]
 
   def index
+    methods = %i[consent_given? consent_refused? consent_conflicts? no_consent?]
+
+    @tabs =
+      @patient_sessions.group_by do |patient_session|
+        methods.find { |m| patient_session.send(m) }
+      end
+
+    methods.each { |m| @tabs[m] ||= [] }
   end
 
   private
@@ -11,5 +20,13 @@ class ConsentsController < ApplicationController
       policy_scope(Session).find(
         params.fetch(:session_id) { params.fetch(:id) }
       )
+  end
+
+  def set_patient_sessions
+    @patient_sessions =
+      @session
+        .patient_sessions
+        .includes(patient: :consents)
+        .order("patients.first_name", "patients.last_name")
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/patient_session.rb
```diff
@@ -49,4 +49,27 @@ def able_to_vaccinate?
     !unable_to_vaccinate? && !unable_to_vaccinate_not_assessed? &&
       !unable_to_vaccinate_not_gillick_competent?
   end
+
+  def consent_given?
+    return false if consent_no_response?
+
+    patient.consents.all?(&:response_given?)
+  end
```
misaka, line 57: Doesn't this and `consent_refused?` override the methods defined in `PatientSessionStateMachineConcern`?

tvararu: You're right!

app/controllers/consents_controller.rb
```diff
@@ -12,4 +26,12 @@ def set_session
...
       )
   end
+
+  def set_patient_sessions
+    @patient_sessions =
+      @session
+        .patient_sessions
+        .includes(patient: :consents)
+        .order("patients.first_name", "patients.last_name")
+  end
```
misaka, line 36: Just a thought ... why is this a `before_action`? It's not being reused, and it's not even using `params` to find a resource, so is it just because we're pulling a resource from the db? Is that enough to qualify it to be a `before_action`? 🤔 

Anyway, not a blocker, it just makes me wish there was a firmer rationale for creating `before_action` hooks.

misaka (approved):

app/controllers/consents_controller.rb
tvararu: Ah, it's just the pattern we use in every other controller because sometimes we reuse these. Though in this case I don't think the other actions will have a need for the full list of `@patient_sessions`.

I do like separating the `@assignments =` that fetch from the DB with this pattern though. It's like once you get some actions doing it (because it makes sense, for example, `set_session`), it's kind of nice to conform all this type of logic to private methods.

PR #615, opened 2023-11-27
Make show pages one

misaka:
Beginning of incremental approach to moving to the new child/patient page design.

This PR simply adds a `AppPatientPageComponent` and moves the common components between the triage and the vaccinations show pages into it. No visual changes and the only changes to journey tests are to add checks for the Back button in the triage journey.


tvararu (approved):
:shipit: it begins!

PR #616, opened 2023-11-27
Add table to consents index

tvararu:
This adds a reusable table component that can be expanded and reused across the consents/triage/vaccinations index pages. 

### After

image

### DIFF
--- a/app/components/app_patient_table_component.rb
+++ b/app/components/app_patient_table_component.rb
@@ -1,0 +1,48 @@
+class AppPatientTableComponent < ViewComponent::Base
+  def call
+    govuk_table do |table|
+      table.with_head do |head|
+        head.with_row do |row|
+          @columns.each { |column| row.with_cell(text: column_name(column)) }
+        end
+      end
+
+      table.with_body do |body|
+        @patient_sessions.each do |patient_session|
+          body.with_row do |row|
+            @columns.each do |column|
+              row.with_cell(text: column_value(patient_session, column))
+            end
+          end
+        end
+      end
+    end
+  end
+
+  def initialize(patient_sessions:, columns: %i[name dob])
+    super
+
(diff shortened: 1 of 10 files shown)


app/components/app_patient_table_component.rb
```diff
@@ -0,0 +1,48 @@
+class AppPatientTableComponent < ViewComponent::Base
+  def call
```
tvararu, line 2: This is another way of defining ViewComponent templates, which I think makes most sense in this context given that it's all Ruby and no HTML currently. https://viewcomponent.org/guide/templates.html#call

app/views/consents/index.html.erb
```diff
@@ -8,24 +8,21 @@
...
-  c.with_tab(label: "Consent refused (#{ @tabs[:consent_refused?].size })",
-             classes: 'nhsuk-tabs__panel') do
-    @tabs[:consent_refused?].size.to_s
-  end
-
-  c.with_tab(label: "Consent conflicts (#{ @tabs[:consent_conflicts?].size })",
-             classes: 'nhsuk-tabs__panel') do
-    @tabs[:consent_conflicts?].size.to_s
+<% def consent_tab(c, label, data)
+  c.with_tab(label: "#{label} (#{data.size })", classes: 'nhsuk-tabs__panel') do
```
misaka, line 12: What is `c` here? Would it ruin your golf game if we gave it a more descriptive name?

misaka:
```suggestion
  c.with_tab(label: "#{label} (#{data.size})", classes: 'nhsuk-tabs__panel') do
```

tvararu: `c` stands for `component` I think https://govuk-components.netlify.app/components/tabs/

misaka (approved):

PR #625, opened 2023-11-28
Add reason for refusal column to refused tab

tvararu:
Also update the example campaign generator to generate campaigns with multiple consents.

TODO:

- [x] Fix the 22 broken tests that all rely on hardcoded patients such as "Alexandra Sipes" that are now gone 😬 

Review commit by commit, ignore generated files.

### Screenshot

image


misaka (approved):
Look great! 💥

PR #626, opened 2023-11-28
Merge consent on triage and vaccinations pages

misaka:
This moves the rendering of the consent section of the patient's page to the patient's page component. Not all of the new designs have been implemented, seeing as this PR is already pretty large they'll come as follow-up PR(s). Known issues:

- "Assess Gillick competence" button is missing
- "No response given" should be "No response yet" with styling changes -- also, the spacing between the text and the button here is different from the new design. However, the new design has a hidden field between these two which is adding this space. Is this space intentional?
- Open health questions for "Triage needed" patients
- Clear green "Consent given" and red "Consent refused"  messages
- Update consent fields in consent response section
- Making consent response sections closable
- Support multiple consent responses
- Hide consent response sections when patient has reached a final "outcome" state

# No consent

### Triage page

image

### Vaccination page

image

## After -- both pages

image

# With consent response

### Triage page

image
 
### Vaccinations page

image

## After -- both pages

image

### DIFF
--- a/app/components/app_details_component.rb
+++ b/app/components/app_details_component.rb
@@ -1,0 +1,28 @@
+class AppDetailsComponent < ViewComponent::Base
+  erb_template <<-ERB
+    <details class="nhsuk-details nhsuk-expander"<%= open_attr %>>
+      <summary class="nhsuk-details__summary">
+        <span class="nhsuk-details__summary-text">
+          <%= @summary %>
+        </span>
+      </summary>
+
+      <div class="nhsuk-details__text">
+        <%= content %>
+      </div>
+    </details>
+  ERB
+
+  def initialize(summary:, open: false)
+    super
+
+    @summary = summary
+    @open = open
+  end
+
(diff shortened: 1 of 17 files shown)


spec/factories/consents.rb
```diff
@@ -45,10 +45,13 @@
     transient do
       random { Random.new }
       health_questions_list { Consent::HEALTH_QUESTIONS.fetch(:flu) }
+      # Allow caller to provide patient_session as a shortcut to prodive
```
tvararu, line 48:
```suggestion
      # Allow caller to provide patient_session as a shortcut to produce
```

app/components/app_details_component.rb
```diff
@@ -0,0 +1,28 @@
+class AppDetailsComponent < ViewComponent::Base
```
tvararu, line 1: Nice!

tvararu (approved):

PR #630, opened 2023-11-29
Update Ruby to 3.2.2

misaka:
I was encountering segfaults caused by syntax errors with 3.2.0.


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # syntax = docker/dockerfile:1.4
 
 # Make sure RUBY_VERSION matches the Ruby version in .ruby-version and Gemfile
-ARG RUBY_VERSION=3.2.0
+ARG RUBY_VERSION=3.2.2
```
tvararu, line 4: Spent 10 minutes trying to golf this to use a `cat .ruby-version` but no luck.

tvararu (approved):

Dockerfile
misaka: 🏌️‍♂️

PR #634, opened 2023-11-30
Update session table display to include town and postcode, make responsive

paulrobertlloyd:
This PR adds `session.location.town` and `session.location.postcode` to each session row. Also, it changes the link to show the `session.location.name`, not `session.name`. Is location information guaranteed to be always available, or does it require checking for the presence of these values, first?

It also updates the table to use the responsive variant provided by the NHS Design System. While not perfect (right aligned values, hmmm), it adapts better to smaller screens than the non-responsive table.

## Mobile

| Before | After |
| - | - |
| Screenshot 2023-11-30 at 13 32 06 | Screenshot 2023-11-30 at 13 37 49 |

## Desktop

| Before | After |
| - | - |
| Screenshot 2023-11-30 at 13 32 23 | Screenshot 2023-11-30 at 13 37 39 |

### DIFF
--- a/app/views/sessions/_session_row.html.erb
+++ b/app/views/sessions/_session_row.html.erb
@@ -1,12 +1,17 @@
 <tr class="nhsuk-table__row">
   <td class="nhsuk-table__cell">
+    <span class="nhsuk-table-responsive__heading">Date</span>
     <%= session.date.to_fs(:nhsuk_date_short_month) %>
   </td>
   <td class="nhsuk-table__cell">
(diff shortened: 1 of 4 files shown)


paulrobertlloyd:
Test failing as expecting session link to include `Flu campaign at` which has been removed. Can we use a known school name, or should we test for the presence of a `data-*` attribute or similar?

tvararu:
Desktop view has weird column widths sometimes:

Screenshot 2023-11-30 at 14 06 55

tvararu (approved):

PR #640, opened 2023-12-04
Fix consent fields in avo

misaka:
Adds better formatting of the JSON health question and answer fields in the `consent` and `consent_form` resources. Without this when editing the a consent record Avo somehow modified the healh questinos JSON, it stripped the topmost wrapping `[]`.

Also fixes complaints about `datetime` not being a field type in these resources.

| Viewing health questions before | Viewing health questions after |
|-|-|
| image | image |

| Editing health questions before | Editing health questions after |
|-|-|
| image | image |


tvararu (approved):
Nice!

PR #647, opened 2023-12-04
Show stats on session page

paulrobertlloyd:
I had a quick go at adding relevant stats to the session page:

- Number of patient in cohort
- Patients with consent given
- Patients with consent refused
- Patients without a response
- Patients needing triage
- Patients ready to vaccinate

To get some of these numbers, it seems as if various states are combined in the consents, triage and vaccination and controllers; it doesn’t seem right to repeat that same or similar logic in the sessions controller too.

Screenshot 2023-12-04 at 15 40 53

All of which is to say is, here’s a page where we’d want to show some information we already show in other places, maybe some broader refactoring is required to make it easier to build out the above?

### DIFF
--- a/app/controllers/triage_controller.rb
+++ b/app/controllers/triage_controller.rb
@@ -19,7 +19,6 @@ class TriageController < ApplicationController
     tabs_to_states = {
       needs_triage: %w[consent_given_triage_needed triaged_kept_in_triage],
       triage_complete: %w[triaged_ready_to_vaccinate triaged_do_not_vaccinate],
-      get_consent: %w[added_to_session],
       no_triage_needed: %w[
         consent_refused
         consent_given_triage_not_needed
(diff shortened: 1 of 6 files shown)


app/controllers/triage_controller.rb
```diff
@@ -13,13 +13,13 @@ def index
...
         .patient_sessions
-        .includes(:vaccination_records, patient: %i[consents triage])
+        .strict_loading
+        .includes(:campaign, :patient)
         .order("patients.first_name", "patients.last_name")
 
     tabs_to_states = {
       needs_triage: %w[consent_given_triage_needed triaged_kept_in_triage],
       triage_complete: %w[triaged_ready_to_vaccinate triaged_do_not_vaccinate],
-      get_consent: %w[added_to_session],
```
tvararu, line 22: This was left in by accident.

tvararu (approved):

PR #648, opened 2023-12-04
Match mum/dad to patient records

tvararu:
This builds on the previous `match_mum_and_dad_info_to_consent` method. The previous way would massage the patient data based on the randomly generated consent data.

This new approach instead purposefully generates a number of consent objects that plausibly match the patient details. It also makes sure that situations where there are 2 consents are uniformly from a mum and a dad, and that situation with 3 consents are from a mum, dad, and granddad.

There's a few more commits that do related work in this area of making the factories more robust/predictable.

See commits.


misaka:
I guess we'll add a way to create conflicting consent examples later? That would be handy.

misaka (approved):

PR #649, opened 2023-12-04
Add link to ‘School sessions’ to signed in header navigation

paulrobertlloyd:
Screenshot 2023-12-04 at 16 08 26

Also fixes the size of the ‘Sign out’ button, which was appearing to small at the mobile breakpoint; removing `@include nhsuk-font(16);` seemingly does the trick.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -40,9 +40,12 @@
         <ul class="nhsuk-header__navigation-list app-header__navigation-list">
           <% if current_user.present? %>
             <li class="nhsuk-header__navigation-item">
-              <div class="nhsuk-header__navigation-link app-u-text-decoration-none">
+              <%= link_to "School sessions", sessions_path, class: "nhsuk-header__navigation-link" %>
+            </li>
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -40,9 +40,14 @@
         <ul class="nhsuk-header__navigation-list app-header__navigation-list">
           <% if current_user.present? %>
             <li class="nhsuk-header__navigation-item">
-              <div class="nhsuk-header__navigation-link app-u-text-decoration-none">
+              <a class="nhsuk-header__navigation-link" href="<%= sessions_path %>">
+                School sessions
+              </a>
```
misaka, line 45: You're so analogue Paul! 😆 

```suggestion
              <%= link_to "School sessions", sessions_path, class: "nhsuk-header__navigation-link" %>
```

There's also a `govuk_link_to`.

But this isn't actually important, just a non-blocking suggestion.

paulrobertlloyd: Ah, of course! Fixed.

PR #654, opened 2023-12-05
Move vaccination card to patient page component.

misaka:
This change doesn't change the journey much, aside from making the vaccination card visible on all the pages. This isn't ideal but I want to get this change in before we go down the rabbit-hole of adding session-awareness and other changes to stuff like the journey (hiding the delivery site on the patient page).

### DIFF
--- a/app/views/vaccinations/show.html.erb
+++ b/app/views/vaccinations/show.html.erb
@@ -8,6 +8,7 @@
 <%= render AppPatientPageComponent.new(
       patient_session: @patient_session,
       consent: @consent,
+      vaccination_record: @draft_vaccination_record,
       route: "vaccinations",
     ) %>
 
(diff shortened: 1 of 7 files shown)


app/views/vaccinations/show.html.erb
```diff
@@ -85,11 +86,5 @@
       </dl>
     </div>
   </div>
-<% elsif @vaccination_record.nil? %>
-  <div class="nhsuk-card">
-    <div class="nhsuk-card__content">
-      <%= render "form", embedded: true %>
```
tvararu, line 91: Is the _form partial still in use?

tvararu (approved):

app/views/vaccinations/show.html.erb
misaka: Oooh, good point. I'll remove it

misaka: 🥾 🥫 ?

PR #656, opened 2023-12-05
Iterate model office data

tvararu:
Depends on https://github.com/nhsuk/manage-childrens-vaccinations/pull/684

- Add patients with conflicting consent to generator
- Add some cases of double consent but triage needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ $ rails generate_example_campaign patients_that_still_need_triage=2 patients_wit
 $ rails generate_example_campaign[db/sample_data/model-office.json] \
   type=hpv \
   seed=42 \
-  username="Nurse Chapel" \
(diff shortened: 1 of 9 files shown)


app/models/concerns/patient_session_state_machine_concern.rb
```diff
@@ -101,7 +101,7 @@ def consent_given?
     def consent_refused?
       return false if no_consent?
 
-      consents.any?(&:response_refused?)
+      consents.all?(&:response_refused?)
```
misaka, line 104: Really? When I was putting together the consent details page I thought that this is correct ... is there going to be a `consent_conflict?` ?

tvararu: Yeah. `consent_refused?` is when all consents were refused. `consent_conflicts?` is when some were refused and some were given.

misaka (approved):
:shipit:

PR #657, opened 2023-12-05
Remove delivery site choice from patient page

misaka:
New designs make users always choose the deliver method and site on the follow-up page instead of the patient page.


paulrobertlloyd:
Eek! We reversed this decision based on feedback from Coventry that the other options are rarely chosen. I’ll review the change and feedback with any changes.

paulrobertlloyd:
This is what we’ve reverted back to:

Screenshot 2023-12-05 at 23 01 41

If you select ‘Other’, we show all the options.

Also: we need to remove the ‘Left buttock’ and ‘Right buttock’ options from the detailed site options.

misaka:
😱  ... alright, thanks for catching this Paul. Sounds like we can just dispose of this PR entirely then.

PR #660, opened 2023-12-06
Conditionally render vaccination forms

misaka:
The patient must be ready to vaccinate and the session must be in progress for the vaccination form to be shown.

This doesn't change the appearance of the vaccination form.


tvararu (approved):
🚀

PR #669, opened 2023-12-07
Design snags (Part 1)

paulrobertlloyd:
Updates to dashboard, session and patient table views. Incudes links to sections that have not been built yet, but we want to include for the model office so that users can see how the current services works within the intended broader context.


tvararu (approved):

tvararu:
Reduced the scope of this to make it easier to merge; I'm packaging up the other changes as a follow-up PR.

PR #679, opened 2023-12-08
Implement CI/CD pipeline for the project

loladzeus:


misaka:
@loladzeus Do you need this merged to test it? Or can you test it out in this branch?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
+name: CI
+
+on:
+  push:
```
misaka, line 4: Should this be `merge`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Build the application
+        run: |
+          echo "Placeholder for actual build commands"
+          # Replace the above line with actual build commands (like npm run build, etc.)
```
misaka, line 17: I know we plan to put stuff in here, but I would be more comfortable if this PR didn't include this build section at all ... I reckon you should keep this on your local system as a stash and you can re-add it when you get around to it, if necessary.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::393416225559:role/GitHubActionsRole
```
tvararu, line 34: How does this work? Did we have to do some prior setup to make it work? Is it this? https://github.com/aws-actions/configure-aws-credentials#overview

.github/workflows/ci.yml
misaka: One of the questions I ask myself when looking at placeholder code in a PR, is if I don't come back to this code for _whatever_ reason, would we want it there? i.e. does it give useful instruction to others? In this case, I would say not.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
...
+          aws-region: eu-west-2
+
+      - name: Install and Verify AWS Copilot
+        run: |
+          curl -Lo /tmp/copilot https://github.com/aws/copilot-cli/releases/latest/download/copilot-linux
+          chmod +x /tmp/copilot
+          mv /tmp/copilot /usr/local/bin/copilot
+          copilot --version
```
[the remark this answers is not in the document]
tvararu: Possibly not a bad idea to leave it in for debugging considering that we're always fetching the latest version of Copilot.

Always fetching the latest version also could backfire if they release a v2 that has breaking changes. But I'm ok with this for now

.github/workflows/ci.yml
loladzeus: Yes, so I went away and created a custom IAM policy, and attached it to a role.
Have a look at IAM->Roles->GitHubActionsRole. If you expand the associated policy, you'll see more granular permissions. The improvements task that's coming up is to slim this down a bit.

.github/workflows/ci.yml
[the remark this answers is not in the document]
loladzeus: I agree with @tvararu. I think one line of output that gives the reassurance that copilot is there and running the expected version is a useful thing to have.

.github/workflows/ci.yml
loladzeus: Don't think so. ref

https://github.com/orgs/community/discussions/26724
https://stackoverflow.com/questions/60710209/trigger-github-actions-only-when-pr-is-merged

.github/workflows/ci.yml
loladzeus: I see what you mean and I agree, a placeholder code is a good idea, but maybe not if it's being productionised

.github/workflows/ci.yml
tvararu: Is it worth documenting this in case we have to re-do it in a new AWS subscription? Or did you just follow the README in which case maybe we just need a pointer to that somewhere in one of our docs

.github/workflows/ci.yml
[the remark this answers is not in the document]
misaka: 👍

PR #682, opened 2023-12-09
Allow editing a consent

tvararu:
Allow nurses to contact a parent and edit their consent if they accept on the phone.

- [x] Separate PR: Correctly fetch HPV health questions from DB instead of legacy hardcoded ones
- [x] Make it work with patients with multiple consents (correctly find the right consent)
- [x] Add test coverage

Stretch:

- [ ] Persist the previous refusal / history

### DIFF
--- a/app/components/app_consent_component.html.erb
+++ b/app/components/app_consent_component.html.erb
@@ -26,17 +26,14 @@
   </p>
 
   <p>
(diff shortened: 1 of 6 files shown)


misaka (approved):
:shipit: -- LFTM, ship it boss!

PR #685, opened 2023-12-09
Update vaccination site values and labels

tvararu:
Use `(upper position)` for the left and right arm sites.

Screenshot 2023-12-09 at 15 43 31

### DIFF
--- a/app/components/app_vaccinate_form_component.rb
+++ b/app/components/app_vaccinate_form_component.rb
@@ -36,6 +36,17 @@ class AppVaccinateFormComponent < ViewComponent::Base
   end
 
   def vaccination_initial_delivery_sites
-    %w[left_arm right_arm other]
+    sites = "activerecord.attributes.vaccination_record.delivery_sites"
+    [
+      OpenStruct.new(
+        value: "left_arm_upper_position",
+        label: t("#{sites}.left_arm_upper_position")
+      ),
+      OpenStruct.new(
+        value: "right_arm_upper_position",
+        label: t("#{sites}.right_arm_upper_position")
+      ),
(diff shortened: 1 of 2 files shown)


app/components/app_vaccinate_form_component.rb
```diff
@@ -36,6 +36,10 @@ def campaign_name
   end
 
   def vaccination_initial_delivery_sites
-    %w[left_arm right_arm other]
+    [
+      OpenStruct.new(value: "left_arm", label: "Left arm (upper position)"),
```
misaka, line 40: In `VaccinationRecord` we have `left_arm` and `left_arm_upper_position` as values for the enum `delivery_site`. Should we be using the latter here?

tvararu: You're right!

app/components/app_vaccinate_form_component.rb
```diff
@@ -36,6 +36,16 @@ def campaign_name
   end
 
   def vaccination_initial_delivery_sites
-    %w[left_arm right_arm other]
+    [
+      OpenStruct.new(
+        value: "left_arm_upper_position",
+        label: "Left arm (upper position)"
```
misaka, line 42: Can we easily use existing entries in `en.yml` for this?

```
en:
  activerecord:
    attributes:
      vaccination_record:
        delivery_sites:
          left_arm_upper_position: Left arm (upper position)
```

tvararu: Hmm, not easily but may as well use the translation strings.

misaka (approved):

PR #690, opened 2023-12-09
Return to consent responses page in consent and vaccination flows

misaka:
When nurses get consent for a patient from the consents page, return them to the consents page when done. Likewise, when getting consent during vaccination, return them to the vaccination pages (only necessary for consent refused as the consent given flow already did this).

### DIFF
--- a/app/views/consents/show.html.erb
+++ b/app/views/consents/show.html.erb
@@ -7,5 +7,5 @@
 
 <%= render AppPatientPageComponent.new(
       patient_session: @patient_session,
-      route: "triage",
+      route: "consents",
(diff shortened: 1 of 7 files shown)


app/views/consents/show.html.erb
```diff
@@ -7,5 +7,5 @@
 
 <%= render AppPatientPageComponent.new(
       patient_session: @patient_session,
-      route: "triage",
+      route: "consents",
```
tvararu, line 10: Was wondering if we forgot to do this, nice

tvararu (approved):

PR #691, opened 2023-12-10
Update the confirmation page for consent

misaka:
> [!Important]
> This PR is based on and depends on #690 

Updates the consent and triage sections of the confirmation page for the consent flow.

image

This is still missing:
- Change links
- Support for recording the nurse recording the consent. This is hard-coded to display the current user at the moment.
- More sophisticated support for the time of recording. This displays the current time for the Response section, but is set to a different time when they actually save. These times should probably be the same.


app/views/nurse_consents/edit_confirm.html.erb
```diff
@@ -26,43 +26,51 @@
...
   <div class="nhsuk-card__content">
     <h2 class="nhsuk-card__heading nhsuk-heading-m">
-      Consent
+      <%= {
+        given: "Consent given by",
+        refused: "Refusal confirmed by",
+        not_provided: "",
+      }[@draft_consent.response.to_sym] %>
+      <%= @draft_consent.parent_name.titleize %>
+      (<%= @draft_consent.human_enum_name(:parent_relationship).capitalize %>)
```
tvararu, line 35: I think this does the same:

```suggestion
      (<%= @draft_consent.who_responded %>)
```

app/views/nurse_consents/edit_confirm.html.erb
```diff
@@ -26,43 +26,51 @@
...
-        row.with_value { @draft_consent.response.humanize }
+        row.with_key { "Name" }
+        row.with_value { @draft_consent.parent_name.titleize }
       end
 
       summary_list.with_row do |row|
-        row.with_key { "For" }
-        row.with_value { @patient.full_name }
+        row.with_key { "Relationship" }
+        row.with_value { @draft_consent.human_enum_name(:parent_relationship).capitalize }
```
tvararu, line 46:
```suggestion
        row.with_value { @draft_consent.who_responded }
```

tvararu (approved):

PR #693, opened 2023-12-11
Ensure model office data contains valid mobile numbers

tvararu:
Update the factory and regenerate the data.

### DIFF
--- a/spec/factories/patients.rb
+++ b/spec/factories/patients.rb
@@ -58,7 +58,8 @@ FactoryBot.define do
     parent_email do
       "#{parent_name.downcase.gsub(" ", ".")}#{random.rand(100)}@gmail.com"
     end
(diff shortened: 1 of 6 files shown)


spec/factories/patients.rb
```diff
@@ -58,7 +58,8 @@
     parent_email do
       "#{parent_name.downcase.gsub(" ", ".")}#{random.rand(100)}@gmail.com"
     end
-    parent_phone { Faker::PhoneNumber.phone_number }
+    # Replace first two digits with 07 to make it a mobile number
+    parent_phone { Faker::PhoneNumber.phone_number.gsub(/\A\d{2}/, "07") }
```
benilovj, line 62: I think this does the same/similar (it's location aware)?
```suggestion
    parent_phone { Faker::PhoneNumber.cell_phone }
```

benilovj (approved):

PR #695, opened 2023-12-11
Prefill consent

tvararu:
This is temporary. The prototype has a better design but it's more effort to implement at this stage.

This prefills the fields when a user attempts to submit consent for a patient whose parent didn't respond to requests for cconsnt.


benilovj (approved):
Temporarily :shipit:

PR #702, opened 2023-12-11
Display attempts to contact a parent for consent

tvararu:
Previously, we would save these as a draft consent record with no recorded_at. Attempting to submit another consent later would surface the details and you'd be able to continue.

Instead, we want to display the consent records in the consent card. The patient session should stay in the same state ("Added to session") when handling consent in this way.

The design isn't final but the functionality is there to test.

Screenshot 2023-12-11 at 17 48 29

### DIFF
--- a/app/components/app_consent_component.html.erb
+++ b/app/components/app_consent_component.html.erb
@@ -1,5 +1,11 @@
 <%= render AppConsentStatusComponent.new(patient_session:) %>
 
+<% if patient_session.no_consent? %>
+  <p class="app-status">
+    No response yet
+  </p>
+<% end %>
+
 <% if patient_session.consents.present? %>
   <% consents_grouped_by_parent.each do |summary, consents| %>
     <%= render AppDetailsComponent.new(summary:, open: open_consents?) do %>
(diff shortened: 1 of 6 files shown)


misaka:
🤔  Looks like some of the generated data doesn't validate. Looking at the model office data there are some invalid phone numbers in it too.

image

Dunno how badly it'll affect the model office, we can fix them in this PR though.

misaka:
I'm pushing a change that manually modifies the model office phone numbers to be valid mobile numbers.

I tried to update the `patients` factory to generate valid mobile numbers and to regenerate the hpv and flu campaigns but the yaks started lining up. For some reason the two campaigns start to randomly pick the same school, causing Nurse Joy to be able to see both sessions.

I think the eventual fix will be to change the factory to generate moblie numbers:

`spec/factories/patients/rb`
```diff
-    parent_phone { Faker::PhoneNumber.phone_number.gsub(/\A\d{2}/, "07") }
+    parent_phone { Faker::PhoneNumber.cell_phone }
```

And then to either generate the hpv and flu example data with different seeds, or to change the loader to not reuse the location between sessions.

misaka (approved):
Your changes look fine, pls review my commit

PR #718, opened 2023-12-14
Add button to make sessions "in-progress"

misaka:
Also adds Flipper feature flagging system.

The button only shows up if the feature flag has been turned on, which can be done from the console from the time being:

```ruby
[1] pry(main)> Flipper.enable :make_session_in_progress
```

image

Also adds success banner:

image


app/controllers/sessions_controller.rb
```diff
@@ -32,6 +32,16 @@ def show
     end
   end
 
+  def make_in_progress
+    @session.update!(date: Time.zone.today)
+    redirect_to session_path,
+                flash: {
+                  success: {
+                    body: "Session is now in progress."
```
tvararu, line 40:
```suggestion
                    heading: "Session is now in progress"
```

app/views/sessions/show.html.erb
```diff
@@ -47,6 +49,10 @@
         </p>
       </div>
     </div>
+
+    <% if Flipper.enabled? :make_session_in_progress_button %>
```
tvararu, line 53:
```suggestion
    <% if !@session.in_progress? && Flipper.enabled? :make_session_in_progress_button %>
```

tvararu (approved):

app/views/sessions/show.html.erb
misaka: I thought about this, but then I thought it would be more consistent to just show it alway, since it's idempotent anyway.

app/views/sessions/show.html.erb
```diff
@@ -47,6 +49,24 @@
...
+        <% if Flipper.enabled? :make_session_in_progress_button %>
+          <p>
+            Nurses can only record vaccinations while a session is "in-progress". You can
+            make this session "in-progress" to test vaccination recording.
+          </p>
+          <%= govuk_button_to(
+            "Make session in-progress",
+            make_in_progress_session_path,
+            method: :put
+          ) %>
```
tvararu, line 66: Maybe something like this, makes it a bit clearer that the in progress state only lasts for a day:

```suggestion
          <p>
            Vaccinations can only be recorded if a session is in progress. If you want
            to test vaccination recording, you can set this session as in progress for today.
          </p>
          <%= govuk_button_to(
            "Set session in progress for today",
            make_in_progress_session_path,
            method: :put
          ) %>
```

tvararu (approved):

PR #721, opened 2023-12-15
Add ManageConsentsController

tvararu:
This is the start of a Wicked rewrite of the NurseConsentsController. The new controller is hidden behind a feature flag and will be built piecemeal, in parallel to the old one.

This first PR implements a stubbed version of the first step (`/consents/who`) as well as the surrounding feature flagging machinery and route. Following PRs will implement the next steps in order.

There are no tests for this, as the intention is to reuse the existing `nurse_consents_*` tests, once the flow is working end to end, to prove that the new implementation is functionally identical to the old.

Review without significant whitespace for the changes to the `routes.rb`.

### DIFF
--- a/app/controllers/dev_controller.rb
+++ b/app/controllers/dev_controller.rb
@@ -8,7 +8,13 @@ class DevController < ApplicationController
     ActiveRecord::Base.connection.transaction do
       data_tables =
         ActiveRecord::Base.connection.tables -
-          %w[users schema_migrations ar_internal_metadata]
+          %w[
+            users
+            schema_migrations
+            ar_internal_metadata
+            flipper_features
+            flipper_gates
+          ]
       data_tables.each do |table|
         ActiveRecord::Base.connection.execute(
           "TRUNCATE #{table} RESTART IDENTITY CASCADE"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -71,19 +71,27 @@ Rails.application.routes.draw do
         get "record-template", on: :collection
       end
 
-      resource :nurse_consents, path: ":route/consent" do
-        get "assessing-gillick", to: "nurse_consents#assessing_gillick"
+      constraints -> { Flipper.enabled?(:new_consents) } do
+        resources :manage_consents,
+                  only: %i[show update],
+                  path: ":route/consents"
+      end
 
-        get "edit/gillick", to: "nurse_consents#edit_gillick"
-        put "update/gillick", to: "nurse_consents#update_gillick"
+      constraints -> { !Flipper.enabled?(:new_consents) } do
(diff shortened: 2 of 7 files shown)


config/routes.rb
```diff
@@ -71,19 +71,27 @@
...
-        get "assessing-gillick", to: "nurse_consents#assessing_gillick"
+      constraints -> { Flipper.enabled?(:new_consents) } do
+        resources :manage_consents,
+                  only: %i[show update],
+                  path: ":route/consents"
+      end
 
-        get "edit/gillick", to: "nurse_consents#edit_gillick"
-        put "update/gillick", to: "nurse_consents#update_gillick"
+      constraints -> { !Flipper.enabled?(:new_consents) } do
```
misaka, line 80: Hrm. I love that we can feature flag this, but at which point do we want to test both paths to ensure the fallback is valid?

I guess we _could_ add this route to expose the new consents on `:route/new_consents`:

```
      constraints -> { !Flipper.enabled?(:new_consents) } do
        resources :manage_consents,
                  only: %i[show update],
                  path: ":route/new_consents"
      end
```

If we think it's worth it. Haven't quite yet decided myself yet.

db/seeds.rb
```diff
@@ -5,3 +5,7 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+
+FEATURE_FLAGS = %i[make_session_in_progress_button new_consents].freeze
+
+FEATURE_FLAGS.each { |flag| Flipper.add(flag) unless Flipper.exist?(flag) }
```
misaka, line 11: Ah, missed this step. Is there a specific reason, though, not to just them auto-create?

app/controllers/dev_controller.rb
```diff
@@ -8,7 +8,13 @@ def reset
...
       data_tables =
         ActiveRecord::Base.connection.tables -
-          %w[users schema_migrations ar_internal_metadata]
+          %w[
+            users
+            schema_migrations
+            ar_internal_metadata
+            flipper_features
+            flipper_gates
+          ]
```
misaka, line 17: Good catch. 👍

misaka (approved):
Comments made, no blockers. 🚀

db/seeds.rb
tvararu: Yeah, you get a warning to STDOUT every time you visit a page with a new feature flag. It makes the test output spammy

config/routes.rb
tvararu: Yeh at some point we can revisit this 👍

PR #737, opened 2023-12-19
Add :reason step to ManageConsent journey

tvararu:
Penultimate step before the Health Questions page which is more complex and should be PRed on its own.

Screenshot 2023-12-19 at 14 24 04

### DIFF
--- a/app/models/consent.rb
+++ b/app/models/consent.rb
@@ -121,10 +121,21 @@ class Consent < ApplicationRecord
                 },
                 presence: true
     end
+
+    with_options if: -> { required_for_step?(:reason) } do
+      validates :reason_for_refusal,
+                inclusion: {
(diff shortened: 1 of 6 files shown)


app/models/consent.rb
```diff
@@ -121,10 +121,21 @@ class Consent < ApplicationRecord
...
+                presence: true
+      validates :reason_for_refusal_other,
+                presence: true,
+                if: -> { reason_for_refusal == "other" }
+    end
   end
 
   def form_steps
-    %i[who agree confirm].compact
+    [:who, :agree, (:reason if response_refused?), :confirm].compact
```
tvararu, line 138: This code brought to you by Prettier(tm)

misaka (approved):

misaka:
:shipit:

PR #741, opened 2023-12-20
Validate health questions in ManageConsents journey

tvararu:
Add a validation to the model and match the params correctly in order to trigger it.

Set @triage on questions step in ManageConsents journey.

### Screenshots

image


misaka (approved):
🚀

PR #742, opened 2023-12-21
Successfully submit consent in ManageConsents journey

tvararu:
This completes the end to end ManageConsent journey.

Gillick competence is next to do in a follow-up, and then we can flip the feature flag and run the tests against the new journey to see what else needs aligning.

### DIFF
--- a/app/controllers/manage_consents_controller.rb
+++ b/app/controllers/manage_consents_controller.rb
@@ -10,7 +10,9 @@ class ManageConsentsController < ApplicationController
   before_action :set_consent, except: %i[create]
   before_action :set_steps, except: %i[create]
   before_action :setup_wizard_translated, except: %i[create]
-  before_action :set_triage, except: %i[create], if: -> { step == "questions" }
+  before_action :set_triage,
+                except: %i[create],
+                if: -> { step.in?(%w[questions confirm]) }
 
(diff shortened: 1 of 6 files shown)


app/controllers/manage_consents_controller.rb
```diff
@@ -68,6 +68,28 @@ def current_step
...
+    flash[:success] = {
+      heading: "Record saved for #{@patient.full_name}",
+      body:
+        ActionController::Base.helpers.link_to(
+          "View child record",
+          session_patient_triage_path(@session, @patient)
+        )
+    }
+    redirect_path
+  end
```
misaka, line 87: If you're revisiting this code at this point, consider (if you haven't) moving the `case ... when` to the bottom of the method and just returning it's result. `redirect_path` doesn't get used in the flash.

app/views/manage_consents/who.html.erb
```diff
@@ -3,7 +3,11 @@
     href: case @route
       when 'consents'
```
misaka, line 4: Tech debt: one day prettier is going to barf on this. Maybe. 😛

misaka (approved):

misaka:
:shipit: I consent!

PR #743, opened 2023-12-22
Move vaccinations show page to patient_sessions

misaka:
At it's core, this just adds `PatientSessionsController` and moves the route `/sessions/:session_id/patients/:patient_id/vaccinations` from `VaccinationsController#show` to `PatientSessionsController#show`. This makes sense because the `vaccinations`, `consents` and `triage` are already all the same page (with a different back button) put with different routes, controllers and view files. This is the first refactoring step to properly unifying them.

What's complicated here, though, is that `VaccinationsController` was performing dual-duty by handling vaccination recording and displaying the show page. This intertwined them making splitting the `show` page out more complicated, hence the size of this PR.

No front-end changes made so no tests changed. Some testing added for missing coverage.


benilovj (approved):
seems sensible to me

PR #755, opened 2023-12-27
Replace foreman with concurrently

tvararu:
`foreman` doesn't properly forward the echo of input to commands. This means that when using the `debugger` or `binding.pry` statements in Rails, we can't see anything that's being typed into the REPL. This makes debugging difficult.

[`concurrently`](https://github.com/open-cli-tools/concurrently) is a JS CLI tool that can be used as an alternative to foreman. It has a `--handle-input` flag that does what we need.

More context: https://github.com/ddollar/foreman/issues/703

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,4 +1,0 @@
-web: unset PORT && bin/rails server
-css: yarn build:css --watch
-js: yarn build --watch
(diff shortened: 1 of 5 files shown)


benilovj (approved):

misaka:
❤️

PR #757, opened 2023-12-27
Enable :new_consents

tvararu:
Depends on https://github.com/nhsuk/manage-childrens-vaccinations/pull/763.

Switches to the new `ManageConsents` journey and `health_answers` for handling all nurse-facing consent flows. Updates the tests where necessary.

See commits.

### DIFF
--- a/tests/nurse_consent_no_response.spec.ts
+++ b/tests/nurse_consent_no_response.spec.ts
@@ -44,7 +44,7 @@ async function when_i_select_a_child_with_no_consent_response() {
 }
 
 async function when_i_click_get_consent() {
-  await p.getByRole("link", { name: "Get consent" }).click();
+  await p.getByText("Get consent").click();
 }
 const and_i_click_get_consent = when_i_click_get_consent;
 
--- a/app/components/app_health_questions_component.rb
+++ b/app/components/app_health_questions_component.rb
@@ -14,11 +14,10 @@ class AppHealthQuestionsComponent < ViewComponent::Base
     </dl>
   ERB
 
-  def initialize(consents:, use_health_answers: false)
+  def initialize(consents:)
     super
 
     @consents = consents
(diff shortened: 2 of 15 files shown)


app/controllers/dev_controller.rb
```diff
@@ -32,6 +32,8 @@ def reset
       )
     end
 
+    Flipper.enable :new_consents
```
misaka, line 35: Will this be removed before merging?

tests/nurse_consent_no_response.spec.ts
```diff
@@ -44,7 +44,7 @@ async function when_i_select_a_child_with_no_consent_response() {
 }
 
 async function when_i_click_get_consent() {
-  await p.getByRole("link", { name: "Get consent" }).click();
```
misaka, line 47: Why is this change not like the others?

tvararu: This one is actually initially a button and then it's a link. It's a button when it's creating a consent, and a link when it's editing one. One alternative is to create two different `when_i_click_*` methods, which would also work

app/controllers/dev_controller.rb
tvararu: We need to keep it in because tests will fail otherwise, since they've been migrated to the new controller and certain edge cases have been updated. It will be removed in the follow-up PR that removes the feature flag completely.

tests/nurse_consent_no_response.spec.ts
misaka: If there's a reason for it then it's probably ok to leave. There's always the chance the test breaks if "Get consent" shows up elsewhere, but we can wait for that calamity to hit before we act on it.

misaka (approved):
🚀 🚀 🚀

misaka (approved):
🐬

PR #765, opened 2023-12-29
Remove new_consents feature flag

tvararu:
Make it the default and switch the app to use consent.health_answers consistently. Remove the field from the database and remove all the legacy NurseConsentsController code.

🎆


misaka (approved):
✂️  🔥

PR #772, opened 2024-01-02
Encrypt fields on models 

tvararu:
Use AR Encryption to ensure that all models have encrypted fields.

I skipped the `nhs_number` field because it's of type `bigint` and it needs to be a `string` at the database level to be encryptable. That can be done in a follow-up.


misaka (approved):
🚀

PR #780, opened 2024-01-03
Remove encryption check in production

tvararu:
Encryption is now on by default in all environments.

Remove `SKIP_PRODUCTION_CHECKS` env var in the pentest env as it's no longer necessary.


misaka (approved):
🔥

PR #781, opened 2024-01-03
Add add_new_user rake task

tvararu:
This is useful for testing if password reset functionality works.

### DIFF
--- a/lib/tasks/add_new_user.rake
+++ b/lib/tasks/add_new_user.rake
@@ -1,0 +1,26 @@
+desc <<-DESC
+  Add a new user
+  Usage:
+    rake add_new_user[email,password,full_name,team_id]
+  Arguments:
+    email: String, the email of the user.
+    password: String, the password for the user account.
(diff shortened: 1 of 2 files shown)


lib/tasks/add_new_user.rake
```diff
@@ -0,0 +1,26 @@
...
+    email: String, the email of the user.
+    password: String, the password for the user account.
+    full_name: String, the full name of the user.
+    team_id: Integer, the ID of the team to add the user to.
+  Example:
+    rake add_new_user['user@example.com','password123','John Doe',1]
+DESC
+task :add_new_user,
+     %i[email password full_name team_id] => :environment do |_task, args|
+  raise "All arguments are required" if args.values.any?(&:nil?)
```
misaka, line 15: I'd imagine password could be optional, but ¯\_(ツ)_/¯ ... we can achieve the same by mashing the keyboard. The idea would be the user would be have to change their password.

misaka (approved):

PR #789, opened 2024-01-04
Use Settings.host instead of HEROKU_APP_DEFAULT_DOMAIN_NAME 

tvararu:
Makes the default host setting in prod-like envs more agnostic of the hosting environment.

Also use `RCVAPP__IS_REVIEW` instead of `ENV["IS_REVIEW"]`, and populate `Settings.host` in the AWS envs.

### DIFF
--- a/app.json
+++ b/app.json
@@ -5,9 +5,9 @@
   "environments": {
     "review": {
       "env": {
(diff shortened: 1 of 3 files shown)


misaka (approved):
:shipit:

PR #790, opened 2024-01-04
Ignore settings.local.yml files in Docker build

tvararu:
These might contain local secrets that shouldn't make it into the container.


misaka (approved):
Good catch

PR #803, opened 2024-01-08
Rename app prefix from RCVAPP to MAVIS 

tvararu:
We have a better name for the app now.


misaka:
PR title?

tvararu:
@misaka fixed

PR #804, opened 2024-01-08
Remove unused fields from Consent model

misaka:
These were originally added before we understood the data model the way we currently do.


tvararu (approved):
✂️ ✂️ ✂️

PR #805, opened 2024-01-08
Remove terraform setup

tvararu:
Leave `docs/terraform.md` in so that it can be `git blame`'d if someone wishes to bring it back.


misaka (approved):
🔥

PR #807, opened 2024-01-08
Remove unused fields from Consent model

misaka:
Change `Patient` column names and types to align with `ConsentForm` which was done when we better understood our data model. Tests seem to all pass so ... :ship:?


tvararu (approved):

app/components/app_patient_details_component.rb
```diff
@@ -15,24 +15,8 @@ def initialize(session:, patient: nil, consent_form: nil, school: nil)
...
-    if @object.respond_to? :dob
-      @object.dob
-    else
-      @object.date_of_birth
-    end
-  end
-
   def aged
-    "aged #{date_of_birth ? @object.age : ""}"
+    "aged #{@object.date_of_birth ? @object.age : ""}"
```
benilovj, line 19: This method will produce `(aged )` if we don't have a DOB for some reason (is that even possible?). Is that definitely what's desired?

misaka: Might not be possible but that's worth fixing anyway. There's no good situation where we'd want `(aged )`

tvararu: Good spot, maybe it should just be `@object.date_of_birth` as it should always be present

misaka: We've decided to hide the `Date of birth` field here altogether if the `date_of_birth` is missing from the patient.

benilovj: 👍

tvararu (approved):

PR #818, opened 2024-01-09
Update copilot docs and settings

tvararu:
The `flow_logs` and `deny_default` settings need to sit under different sections.

Remove `deny_default: true` as it breaks the app.

### DIFF
--- a/copilot/webapp/manifest.yml
+++ b/copilot/webapp/manifest.yml
@@ -29,10 +29,9 @@ exec: true # Enable running commands in your container.
 network:
   connect: true # Enable Service Connect for intra-environment traffic between services.
   vpc:
-    flow_logs: on
-    deny_default: true
     security_groups:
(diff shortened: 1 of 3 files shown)


copilot/webapp/manifest.yml
```diff
@@ -29,10 +29,9 @@ exec: true # Enable running commands in your container.
 network:
   connect: true # Enable Service Connect for intra-environment traffic between services.
   vpc:
-    flow_logs: on
-    deny_default: true
     security_groups:
-      - from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-dbSecurityGroup
+      groups:
+        - from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-dbSecurityGroup
```
misaka, line 34: Do we know that this works? I know that's how we've understood the docs, but just wondering if this is tested vs what we had before without `groups:`

tvararu: The app does boot successfully, so I assume so...

misaka (approved):

PR #830, opened 2024-01-15
Add validation for missing CSV and invalid CSV

tvararu:
Screenshot 2024-01-15 at 13 19 37


paulrobertlloyd:
Is ‘valid’ the right word here? Is it more the case that the file is not in the correct format? (Valid is something is more a way about describing the data, not the file). @emjhf

misaka (approved):

misaka (approved):

PR #831, opened 2024-01-15
Add child fields to registration form

misaka:
image

## Validations

image
----

> [!Note]
> In the screenshot below, errors have been moved to the top of the page as in the screen shot above, but I just can't be bothered to regenerate the screenshot right now.

image


app/views/registrations/new.html.erb
```diff
@@ -25,31 +25,98 @@
...
+                             label: { text: 'Postcode' },
+                             class: 'nhsuk-input--width-10' %>
+    <% end %>
+
+    <%= f.govuk_text_field :nhs_number,
+                           label: { text: "NHS number (optional)", size: "s" },
+                           class: "nhsuk-input--width-10" %>
   <% end %>
+
+  <%= govuk_section_break visible: true, size: "l" %>
```
misaka, line 119: @paulrobertlloyd As discussed, I'm moving the section break out of the `fieldset`. This just has a minor effect of adding a little bit more space between the input box directly above and the `hr` generated.

The prototype puts the `hr` and following `<h1>Terms and conditions</h1>` inside the `fieldset` above.

tvararu (approved):

PR #832, opened 2024-01-15
Validate cohort list CSV rows

tvararu:
This partially implements the `#generate_cohort!` method. A new `CohortListRow` model is used to loop over all the rows and validate them against standard Rails-y validations. Each error is added to the `CohortList` model and displayed on a separate page.

See commits.

Known issues:

- Postcode validator can't be overridden, I tried to but it looked like a rabbit hole, so to be revisited

### Malformed CSV

Screenshot 2024-01-17 at 07 34 48

### Invalid headers

Screenshot 2024-01-16 at 17 39 59

### Invalid fields

Screenshot 2024-01-16 at 17 40 12
Screenshot 2024-01-16 at 17 40 22

### Valid upload (still faked)

Screenshot 2024-01-16 at 17 40 34


app/models/cohort_list_row.rb
```diff
@@ -0,0 +1,44 @@
...
+                :child_last_name,
+                :child_common_name,
+                :child_date_of_birth,
+                :child_address_line_1,
+                :child_address_line_2,
+                :child_address_town,
+                :child_address_postcode,
+                :child_nhs_number
+
+  validates :parent_name, presence: true
```
misaka, line 18: I think we should replicate the length validations here. Even if it's (slightly) less susceptible to attack, it'd be a good idea to just keep them lined up.

misaka: Just had a thought ... the length validations will kick in later, when trying to save the record wherever these are destined to go ... so maybe it's not that crucial to have length validations here actually.

misaka (approved):
This is tricky stuff, nice work!

PR #842, opened 2024-01-16
Add some documentation about S3 server access logging

loladzeus:


docs/aws-copilot.md
```diff
@@ -390,3 +390,30 @@ The following sites contain some valuable examples that can be used for inspirat
...
 - https://asecure.cloud/l/cloudwatch/
 - https://www.intelligentdiscovery.io/controls/cloudwatch/
+
+## Server Access Logging on S3 Buckets
+
+This is a feature provided by AWS. Server access logging for S3 buckets in AWS provides detailed records for the requests made to a specific bucket. These logs are invaluable for security and audit purposes.
+
+### Enabling Server Access Logging
+
+This can be done through the Management Console as well as the AWS CLI, but for the sake of convenience, we will only conver doing it though former.
```
misaka, line 400:
```suggestion
This can be done through the Management Console as well as the AWS CLI, but for the sake of convenience, we will only cover doing it though former.
```

docs/aws-copilot.md
```diff
@@ -390,3 +390,30 @@ The following sites contain some valuable examples that can be used for inspirat
...
+1. Navigate to the AWS Management Console
+2. Navigate to S3
+3. Select the bucket you want to enable logging for.
+4. Click the Properties tab.
+5. Find Server access logging and click Edit.
+6. Configure the destination where the logs will be written.
+
+- here you can use the same bucket, but configure a separate path in the root directory of the bucket for visibility.
+- here you can also create the prefix directory at the root level of the S3 bucket.
+
```
misaka, line 413: I believe if you indent the list like so, it'll nest better under the number list item above.

```suggestion
    - here you can use the same bucket, but configure a separate path in the root directory of the bucket for visibility.
    - here you can also create the prefix directory at the root level of the S3 bucket.

```

misaka: What settings should we use on this edit page in general? Those would be handy to have here.

misaka (approved):

PR #846, opened 2024-01-17
Add #generate_patients! method to CohortList 

tvararu:
This loops through each CohortListRow, converts them to patient records, and then saves the records.

~This doesn't use the method in the controller, that's coming in a follow-up.~ It does now, see last commit!

Screenshot 2024-01-17 at 16 41 44


app/models/cohort_list.rb
```diff
@@ -48,6 +51,16 @@ def parse_rows!
         .map { CohortListRow.new(_1) }
   end
 
+  def generate_patients!
+    return if invalid?
```
misaka, line 55: With the exclamation in the method name, I kind of expect these to raise an error. I know that's not the only meaning for `!` in methods, but it would be pretty easy to just use `validate!` here, or would break things?

config/locales/en.yml
```diff
@@ -127,6 +130,12 @@ en:
               invalid: Choose a CSV file in the correct format
         cohort_list_row:
           attributes:
+            submitted_at:
+              blank: "is required but missing"
+              invalid: "‘%{value}’ should be formatted as YYYY-MM-DDTHH:MM:SSZ"
```
misaka, line 135: Does Excel have a name for this format? Or does it call it `YYYY-MM-DDTHH:MM:SSZ`? Just thinking that this may not mean anything to nurses in this format.

app/models/cohort_list.rb
tvararu: Then we'd have to catch them... Isn't that unusual? I used `!` here to signify that it's a method that mutates the instance.

config/locales/en.yml
tvararu: Hopefully they don't edit this field, it's one of the ones that should be pre-filled by the download... The validation is more of a precaution.

Also, this field uses `Date.parse` which is quite lax on what it accepts, other formats should work fine.

That being said, we do ask them to input `YYYY-MM-DD` (ISO8601) for the date of birth, and this is the same format... but expanded to include exact time, as this will be used to disambiguate which are the first 100 parent records.

app/models/cohort_list.rb
misaka: In Ruby `!` is used for variants of methods that may do unexpected things. I don't think there are many, if any, examples of methods that only have a `!` version.

Here the method isn't doing anything unexpected, so not sure it needs the `!`, but it's a minor issue.

misaka (approved):

PR #847, opened 2024-01-17
Add registration closed page

misaka:
This gets shown when registration is not open, as controlled by a feature flag.

Still need to hear back if this is the right email address to use, though. Will wait for that before merging.

image

### DIFF
--- a/app/views/registrations/closed.html.erb
+++ b/app/views/registrations/closed.html.erb
@@ -10,5 +10,5 @@
 
 <p>
   If you’re interested in taking part in future research with the NHS, please
-  email <a href="mailto:england.mavis@nhs.net">anthony.green11@nhs.net</a>.
+  email <%= govuk_mail_to t("service.nhs_ur_email") %>.
 </p>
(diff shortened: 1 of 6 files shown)


app/views/registrations/closed.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+<p>
+  If you have not already responded to messages from your child’s school about
+  the HPV vaccination, please do so. The routine vaccination campaign is running
+  in parallel with the pilot.
+</p>
+
+<p>
+  If you’re interested in taking part in future research with the NHS, please
+  email <a href="mailto:england.mavis@nhs.net">anthony.green11@nhs.net</a>.
```
tvararu, line 13: We should update and use `service.email` from `en.yml`:

```suggestion
  email   <a class="nhsuk-link" href="mailto:<%= t('service.email') %>"><%= t('service.email') %></a>.
```

tvararu (approved):
Pre-approving

app/views/registrations/closed.html.erb
tvararu: Can you also update `app/views/registrations/new.html.erb:128`?

PR #850, opened 2024-01-18
Split out CohortListsController

tvararu:
Extract this functionality from the PilotController to slim it down.


misaka (approved):
:shipit:

PR #852, opened 2024-01-18
Add filtering to patient tables

paulrobertlloyd:
This PR uses the `<action-table-filters>` web component to filter rows in patient tables.

Screenshot of patient table with filters.

I’ve created a ViewComponent for the filters, which takes a `tab_id` (used in `id` and `for` attributes on the filter fields) and `filter_actions`, a boolean to determine if the filters for actions is shown).

A test for this new component is failing currently, not sure how/what I need to change to clear this.

### DIFF
--- a/app/assets/stylesheets/_action-table.scss
+++ b/app/assets/stylesheets/_action-table.scss
@@ -1,3 +1,14 @@
+action-table-filters,
+action-table-no-results {
+  display: none;
+
+  @include mq($from: tablet) {
+    &:defined {
+      display: block;
+    }
+  }
+}
+
 action-table {
   display: block;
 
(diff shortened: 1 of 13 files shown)


tvararu:
Without JS, the tabs collapse into a single page, and each tab section has its own filters that don't do anything:

Screenshot 2024-01-19 at 09 44 39

*Proposed solution:* Perhaps we should hide the filtering if there is no JS?

---

On mobile, with JS, the collapsing behaviour is the default, and if there's not a lot of results (like after you filter), it's easy to see two different filtering widgets in the same screen, which can be a bit confusing ("which one is for which?")

Screenshot 2024-01-19 at 09 45 58

*Proposed solution:* Perhaps we should hide the filtering on mobile?

paulrobertlloyd:
@tvararu Am now hiding `<action-table-filters>` unless it’s `:defined` (initiated by JS), and, at least for now, until the viewport is larger than the `tablet` width.

I’m going to do a follow on PR to add captions to each table, as they should be included, and the page makes more sense in the movie layout when these are shown.

I’ve also looked at the tab font size, which I think we can make smaller between `tablet` and `desktop` and make the tabs look less broken on narrower screens. Again, separate PR for that, I think.

tvararu (approved):

PR #856, opened 2024-01-19
Generate registrations as part of example data 

tvararu:
These will help us more easily test out cohort related functionality.

Also add a `registration` factory.

### Soon in https://github.com/nhsuk/manage-childrens-vaccinations/pull/851

Screenshot 2024-01-19 at 17 05 50


db/sample_data/example-flu-campaign.json
```diff
@@ -1201,5 +1201,92 @@
         }
       ]
     }
+  ],
+  "registrations": [
+    {
+      "firstName": "Chasidy",
```
misaka, line 1207: This reminds me that we really outta tidy up the rest of the snake-case in these JSON files.

misaka (approved):
🚀

db/sample_data/example-flu-campaign.json
tvararu: Tidy up in what way? Isn't it all camel cased? That's idiomatic in JS/JSON, I think we should leave it as is.

PR #857, opened 2024-01-19
Allow users to download pilot cohort

tvararu:
Adds the button to the pilot registrations page.

### Button

Screenshot 2024-01-19 at 17 37 57

### DIFF
--- a/app/controllers/pilot_controller.rb
+++ b/app/controllers/pilot_controller.rb
@@ -13,4 +13,10 @@ class PilotController < ApplicationController
         .where(location_id: current_user.team.locations.map(&:id))
         .group_by { |r| r.location.name }
   end
+
+  def download
+    registrations = Registration.where(location: current_user.team.locations)
+    csv = CohortList.from_registrations(registrations).to_csv
+    send_data(csv, filename: "registered_parents.csv")
+  end
 end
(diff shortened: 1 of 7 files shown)


app/models/cohort_list.rb
```diff
@@ -61,6 +80,23 @@ def generate_patients!
     end
   end
 
+  def self.from_registrations(registrations)
+    new.tap do |cohort_list|
+      cohort_list.data =
+        registrations.map do |registration|
+          REGISTRATION_ATTRS.map { registration.send(_1) }
```
misaka, line 87: I don't now why but `send` always feels like a smell to me. Here's a slightly more idiomatic way that avoids it:

```suggestion
          registration.values_at(*REGISTRATION_ATTRS)
```

app/controllers/pilot_controller.rb
```diff
@@ -13,4 +13,10 @@ def registrations
         .where(location_id: current_user.team.locations.map(&:id))
         .group_by { |r| r.location.name }
   end
+
+  def download
+    registrations = Registration.where(location: current_user.team.locations)
```
misaka, line 18: Feels like this should be a scope in the locations Pundit policy. Not urgent though, puntable.

misaka (approved):
I'd fix the `send`, but not blocking. :shipit:

PR #858, opened 2024-01-22
Add draft sessions

misaka:
Does the basic foundation that will be required for the full "Add a session" journey. No Wicked forms yet, this only tweaks the DB table to allow drafts, and allows the user to create a session at a random school. It also adds a confirmation page, and until the session details are confirmed the draft session object won't show up in the sessions list.

> [!NOTE]
> The design of the confirmation page does not use a visual card component for the session details, unlike other confirmations pages we have. I've but it into a card for now until Paul gets back to us about it.

image


app/controllers/sessions_controller.rb
```diff
@@ -1,9 +1,23 @@
 class SessionsController < ApplicationController
-  before_action :set_session, only: %i[show make_in_progress]
+  before_action :set_session, only: %i[show make_in_progress confirm update]
   before_action :set_school, only: %i[show]
 
+  layout "two_thirds", except: :index
+
+  def create
+    team = current_user.teams.first
+    campaign = team.campaigns.first
```
tvararu, line 9:
```suggestion
    campaign = current_user.team.campaigns.first
```

app/controllers/sessions_controller.rb
```diff
@@ -1,9 +1,23 @@
...
+
+    @session =
+      Session.create! campaign:,
+                      date: Time.zone.today,
+                      location: current_user.teams.first.locations.sample
```
tvararu, line 14:
```suggestion
                      location: current_user.team.locations.first
```

tvararu (approved):
You can use `user.team` instead of `user.teams.first` 🚢

PR #860, opened 2024-01-22
Render pages from content in Markdown files, add privacy policies

paulrobertlloyd:
- Adds `render_content_page` helper to render Markdown using `govuk_markdown` (with `govuk` namespace in classes renamed `nhsuk` to ensure the correct typographic styles are used).
- Adds content controller to serve rendered markdown files
- Add privacy policy and summary privacy policy
- Adds link to privacy policy to application footer

There’s a linting issue with using a `@global` variable in a helper, not sure what’s the best way around this. Also, how much testing do we want/need around these new features?


app/controllers/content_controller.rb
```diff
@@ -0,0 +1,12 @@
+class ContentController < ApplicationController
```
tvararu, line 1: Alternate name:

```suggestion
class StaticController < ApplicationController
```

tvararu:
@paulrobertlloyd fixed your linting issue by disabling that rule in this case

tvararu (approved):

app/controllers/content_controller.rb
paulrobertlloyd: I like it!

PR #861, opened 2024-01-22
Fix service name aria-label

tvararu:
This was a placeholder value.


paulrobertlloyd:
Does the `svg` get ignored? Or does it get read out as an unlabelled image?

paulrobertlloyd (approved):

PR #878, opened 2024-01-23
Add session consent timestamps

tvararu:
These will be populated when adding a new session based on the date of the session.


app/avo/resources/session.rb
```diff
@@ -15,5 +15,8 @@ def fields
     field :location, as: :belongs_to
     field :patients, as: :has_and_belongs_to_many
     # add fields here
+    field :send_consent, as: :date_time
```
misaka, line 18:
```suggestion
    field :send_consent_at, as: :date_time
```

misaka (changes requested):
Missing `_at`

misaka (approved):

misaka (approved):

PR #879, opened 2024-01-23
Allow NHS numbers with spaces in the cohort spreadsheet

benilovj:
This mirrors the validations on the Patient model


misaka (approved):
image

PR #882, opened 2024-01-24
Add WizardFormConcern 

tvararu:
This will be reused across models to DRY up the common logic around running certain validations on particular form steps.

First commits are not related.

Review without whitespace.


app/models/concerns/wizard_form_concern.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  included { attr_accessor :form_step }
+
+  class_methods do
+    def on_wizard_step(step, exact: false, &block)
+      with_options on: :update do
+        with_options if: -> { required_for_step?(step, exact:) }, &block
+      end
+    end
+  end
```
tvararu, line 12: One thing missing here is some mechanism to force the user to define a `form_steps` method on the class. I didn't look into doing this, not sure what would be idiomatic, would it go into the `included` bit, perhaps in an initialiser?

app/models/consent.rb
```diff
@@ -92,51 +93,49 @@ class Consent < ApplicationRecord
 
   validates :reason_for_refusal_other, length: { maximum: 1000 }
 
-  with_options on: :update do
-    with_options if: -> { required_for_step?(:gillick, exact: true) } do
-      validate :patient_session_valid?
-    end
+  on_wizard_step :gillick, exact: true do
```
misaka, line 96: I like how this reads

misaka (approved):

PR #887, opened 2024-01-25
Add timeline page to new session journey

tvararu:
This adds a `:timeline` step to the wizard and the corresponding HTML and model validations. The validations are quite complex as the fields don't directly manipulate dates. Instead, temporary `attr_accessors` are being set using model callbacks.

The code is fairly light on tests... I would have liked more time to fix that, but decided to test the happy path and leave the rest for later.

### Timeline

Screenshot 2024-01-25 at 20 12 54

### Validations
Screenshot 2024-01-25 at 20 13 49

### Confirmation

Screenshot 2024-01-25 at 20 14 01


app/models/session.rb
```diff
@@ -28,17 +38,122 @@ class Session < ApplicationRecord
...
+    (date - send_consent_at).to_i / 1.day
+  end
+
+  def days_between_consent_and_reminder
+    (send_reminders_at - send_consent_at).to_i / 1.day
+  end
+
+  private
+
+  def set_timeline_attributes
```
misaka, line 109: It feels like this method and the one below are view / controller / UI concerns and don't really belong here. Maybe one day we should consider using a decorator or something similar.

misaka (approved):
> The code is fairly light on tests... I would have liked more time to fix that, but decided to test the happy path and leave the rest for later.

Agreed that the code in the model could use some more tests. But, ya, it's not as urgent considering this is session setup. 🚢

PR #889, opened 2024-01-26
Do not show registration page when school registration is closed

misaka:
Finishes the work of the previous PR to add a "close pilot for school" journey by ensuring that the registration page does not show when the school's registration is closed.

Also adds a test for the close pilot registration journey.


benilovj:
What happens if you try to submit a registration form (that you started while registration was still open) after registration is closed?

misaka:
> What happens if you try to submit a registration form (that you started while registration was still open) after registration is closed?

There are no guards against that, so it would go through.

benilovj (approved):

PR #893, opened 2024-01-29
Add location page to new session form

misaka:
image

## Confirmation page

image

## Validations

This validates that `location_id` has been submitted, and that it is within the list of locations that are associated with the user's team.

image

### DIFF
--- a/app/controllers/edit_sessions_controller.rb
+++ b/app/controllers/edit_sessions_controller.rb
@@ -9,15 +9,20 @@ class EditSessionsController < ApplicationController
   before_action :setup_wizard_translated
 
   def show
+    case current_step
+    when :location
+      @locations = current_user.team.locations
+    end
+
     render_wizard
   end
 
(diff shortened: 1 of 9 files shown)


app/controllers/edit_sessions_controller.rb
```diff
@@ -58,4 +62,8 @@ def update_params
   def set_steps
     self.steps = @session.form_steps
   end
+
+  def set_locations
+    @locations = policy_scope(Location)
```
benilovj, line 67:
```suggestion
    @locations = policy_scope(Location).order(:name)
```

benilovj (approved):
Pre-emptive approval, small suggestion around sorting of schools.

app/controllers/edit_sessions_controller.rb
misaka: Good suggestion. 👍

PR #912, opened 2024-02-01
Update fieldset and label spacing

paulrobertlloyd:
- Update CSS fix for `nhsuk-label` and `nhsuk-fieldset` that normalises the spacing
- Fix the alignment of first label on patient table filter
- Consistent conditional label/fieldset size on session timeline page


misaka:
Once merged should we have a look to see if there are other pages that need updating?

PR #915, opened 2024-02-05
Session cohort

misaka:
There are a few (known) issues that should be addressed in follow-up PRs:
- The spacing doesn't quite match the prototype, in particular the space between the checkbox and the patient name is considerably bigger.
- There's no checkbox at the top to select/deselect all patients.
- The validations may need a little bit of tweaking, not sure if there should be a red line next to all the checkboxes for example, or whether the linking should work.

Also, the session create tests are getting pretty big. I'd propose we move the validations to a separate test file.

image

## Validations

image

### DIFF
--- a/app/controllers/edit_sessions_controller.rb
+++ b/app/controllers/edit_sessions_controller.rb
@@ -88,6 +88,7 @@ class EditSessionsController < ApplicationController
 
   def set_steps
     self.steps = @session.form_steps
+    @previous_step = previous_step
   end
 
   def set_locations
--- a/app/controllers/edit_sessions_controller.rb
+++ b/app/controllers/edit_sessions_controller.rb
@@ -25,9 +25,12 @@ class EditSessionsController < ApplicationController
   def update
     case current_step
     when :confirm
-      copy_entire_cohort_to_session
       @session.draft = false
     when :location, :vaccine
+      # At this step in the form a location or vaccinee (aka campaign) will not
+      # be set. Validations rely on the team being set here to be able to
+      # validate the user has access to the location or campaign being set
+      # during this step.
       @session.team = current_user.team
       @session.assign_attributes update_params
     else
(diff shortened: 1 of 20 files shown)


config/locales/en.yml
```diff
@@ -505,3 +526,5 @@ en:
             time_of_day:
               blank: Select a time of day
               inclusion: Select a time of day
+            patients:
+              blank: Cannot continue without selecting patients
```
tvararu, line 530:
```suggestion
              blank: Select one or more patients to add to the session
```

app/models/patient.rb
```diff
@@ -70,6 +70,13 @@ def vaccination_records_for_session(session)
     patient_sessions.find_by_session_id(session.id).vaccination_records
   end
 
+  def year_group
+    first_school_year = date_of_birth.year + 5
+    first_school_year += 1 if date_of_birth.month >= 9
+    Time.zone.today.year - first_school_year +
+      (Time.zone.today.month >= 9 ? 1 : 0)
+  end
```
tvararu, line 78: I get the feeling this can be golfed but 🚢

tvararu (approved):

app/models/patient.rb
tvararu: immediately and then applies the two conditional increments:

```suggestion
  def year_group
    result = Time.zone.today.year - date_of_birth.year - 5
    result -= 1 if date_of_birth.month >= 9
    result += 1 if Time.zone.today.month >= 9
    result
  end
```

Accept it if you think it's clearer, reject if not 👍

misaka: Ya, 🚢 for now I say. Once we have the baseline set we can start playing golf with it. 😄

app/controllers/edit_sessions_controller.rb
```diff
@@ -88,6 +88,7 @@ def update_params
 
   def set_steps
     self.steps = @session.form_steps
+    @previous_step = previous_step
```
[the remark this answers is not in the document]
misaka: I don't believe this is available in the views. Unless I was doing something wrong ... but I did try it and got an error, and it appears to be documented as a controller helper (vs views helpers) here: https://github.com/zombocom/wicked#quick-reference

config/locales/en.yml
misaka: Fix from a mini snag session with Emma this morning. 🙏

PR #918, opened 2024-02-05
Session stats: bugfixes

benilovj:
* Avoid counting draft consent forms when counting unmatched forms

Without this fix, you see an inconsistency:

image

image

* Don't count patients in triage or without consent as 'ready to vaccinate'
* Extract a `SessionStats` model to make it easier to test the logic

### DIFF
--- a/app/models/session_stats.rb
+++ b/app/models/session_stats.rb
@@ -2,9 +2,21 @@ class SessionStats
   def initialize(patient_sessions:, location:)
     @patient_sessions = patient_sessions
     @location = location
+
+    @stats = calculate_stats
   end
 
-  def call
+  def [](key)
+    @stats[key]
+  end
+
+  def to_h
+    @stats
+  end
+
+  private
+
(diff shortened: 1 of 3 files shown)


misaka (approved):

app/models/session_stats.rb
```diff
@@ -0,0 +1,32 @@
+class SessionStats
+  def initialize(patient_sessions:, location:)
+    @patient_sessions = patient_sessions
+    @location = location
+  end
+
+  def call
```
misaka, line 7: Mixing in this service-style `call` into a model feels a little odd to me. I think it would fit the "model" pattern better if the stats were attributes.

benilovj: Agreed. Hopefully addressed in [17bcbc4](https://github.com/nhsuk/manage-childrens-vaccinations/pull/918/commits/17bcbc45ba395bbe8fd3b3fe18a5dbdbabc1a021)

PR #934, opened 2024-02-05
Quote phone fields in cohort list

misaka:
If we don't quote these then Excel prompts the user to process the file to remove the leading zeros.

image

However, quoting fields can have other less severe side-effects because essentially we're turning these fields into formulas that simply return a string. Here's an example where Excel has flagged cells containing "Grandmother" and "Mother" because it thinks the formulas may be inconsistently entered formulas.

image

So for this PR I've opted to only quote the phone numbers, since that appears to give the cleanest experience. We can change this as required.

### DIFF
--- a/app/models/cohort_list.rb
+++ b/app/models/cohort_list.rb
@@ -22,24 +22,25 @@ class CohortList
     CHILD_NHS_NUMBER
   ].freeze
 
-  REGISTRATION_ATTRS = %i[
-    created_at
-    location_id
-    location_name
-    parent_name
-    formatted_parent_relationship
-    parent_email
-    parent_phone
-    first_name
-    last_name
-    common_name
-    date_of_birth
-    address_line_1
-    address_line_2
-    address_town
-    address_postcode
-    nhs_number
-  ].freeze
+  REGISTRATION_ATTRS =
+    [
+      :created_at,
+      :location_id,
+      :location_name,
+      :parent_name,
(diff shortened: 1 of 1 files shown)


app/models/cohort_list.rb
```diff
@@ -22,24 +22,25 @@ class CohortList
...
+      :first_name,
+      :last_name,
+      :common_name,
+      :date_of_birth,
+      :address_line_1,
+      :address_line_2,
+      :address_town,
+      :address_postcode,
+      :nhs_number
+    ].map { _1.is_a?(Array) ? _1 : [_1, nil] }.freeze
```
misaka, line 43: This bit may seem over-engineered, but I was mostly focused on keeping the list of attrs as clean as readable as possible so it's very clear which attrs are quoted. There are a couple of other ways to do this, but this seems cleanest.

Open to comments and suggestions to improve this.

app/models/cohort_list.rb
```diff
@@ -67,6 +68,7 @@ def parse_rows!
     self.rows =
       data
         .map { |raw_row| raw_row.to_h.transform_keys { _1.downcase.to_sym } }
+        .map { |row| row.transform_values { _1&.sub(/^="(.*)"$/, '\1') } }
```
misaka, line 71: I added automatic unquoting of fields so that users can directly copy-and-paste into the upload sheet. It should be safe to automatically do this on all fields coming in.

PR #935, opened 2024-02-06
Force quotes when generating cohort CSV

tvararu:
Graphical spreadsheet editors like Excel, Google Sheets, and Numbers tend to mangle data such as phone numbers (which have a leading 0) when importing CSV data. Using quotes around all fields should help to prevent this.


benilovj (approved):
Tested editing in Excel on macOS, it appears to work correctly.

PR #936, opened 2024-02-06
Add all select/deselect checkbox for session cohort table

tvararu:
Allow users to select/deselect all patients in the cohort in one go.

The checkbox is hidden by default using CSS and shown if JS is enabled, as it's a progressive enhancement.

### Screenshots

Screenshot 2024-02-06 at 09 06 26


paulrobertlloyd:
Looks good. Alternative might be to add a [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute to the select all checkbox, and remove that with JavaScript.

tvararu:
@paulrobertlloyd I like that, updated!

misaka (approved):
Looks good! Check in review app

PR #938, opened 2024-02-06
Find the correct patient_session

misaka:


benilovj (approved):
Fixes the issue locally.

PR #944, opened 2024-02-07
Add prompting to all pilot ops scripts

tvararu:
Makes them easier to use.

### DIFF
--- a/lib/tasks/add_new_hpv_team.rake
+++ b/lib/tasks/add_new_hpv_team.rake
@@ -1,27 +1,30 @@
+require_relative "../task_helpers"
+
 desc <<-DESC
-  Add a new HPV team
+  Add a new HPV team consisting of a team, campaign, vaccine, and
+  health questions.
 
   Usage:
+    rake add_new_hpv_team # Complete the prompts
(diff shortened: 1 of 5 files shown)


lib/tasks/add_new_hpv_team.rake
```diff
@@ -1,27 +1,32 @@
...
+  if args.to_a.empty? && $stdin.isatty && $stdout.isatty
+    email = prompt_user_for "Enter team email:", required: true
+    name = prompt_user_for "Enter name:", required: true
+    ods_code = prompt_user_for "Enter ODS code:"
+    privacy_policy_url = prompt_user_for "Enter privacy policy URL:"
+  elsif args.to_a.size == 4
+    email = args[:email]
+    name = args[:name]
+    ods_code = args[:ods_code]
+    privacy_policy_url = args[:privacy_policy_url]
```
misaka, line 24: Puntable ... but I was thinking it would be nice if `prompt_user_for` took `args[:email]`, for example, and just returned it if already populated. It would clean up this bit here a little bit.

misaka (approved):
Just added a suggested improvement for `prompt_user_for` but ready to 🚢  as-is.

PR #946, opened 2024-02-07
Tidy up registration path

tvararu:
This makes the registration URL more intuitive, allowing an arbitrary slug to go after the /:school_id.

Before: 

```
/schools/1/registration/new
```

After:

```
/1 # or
/1/pool-hayes-primary
```


misaka:
Seems to break on review apps. Can we make them work? Happy to punt if that's too hard, but I think it would be important to have them sorted.

image

misaka (approved):

tvararu:
The `/1/` URL isn't actually working as it's not specific enough and conflicts with `/422`. There isn't enough time to keep iterating this right this moment.

PR #951, opened 2024-02-08
Don't allow cohorts larger than 100 records 

tvararu:
This adds a validation to the cohort upload to check that the total amount of records across all the locations does not exceed a max cohort size.

It also displays the current cohort size, to allow users to know how many more records they can upload before their upload will fail.

### Screenshot

Screenshot 2024-02-08 at 18 02 23


config/locales/en.yml
```diff
@@ -169,6 +169,7 @@ en:
             csv:
               blank: Choose a file
               invalid: Choose a CSV file in the correct format
+              too_large: "Choose a a CSV with %{remaining} or fewer rows"
```
benilovj, line 172:
```suggestion
              too_large: "Choose a CSV with %{remaining} or fewer rows"
```

benilovj (approved):
Looks good (1 minor typo)

PR #953, opened 2024-02-09
Validate school ids in cohort list belong to team

misaka:
Otherwise someone could add a child to a school that belongs to another school, which would make that child visible to the team that owns that school.

### DIFF
--- a/app/models/cohort_list.rb
+++ b/app/models/cohort_list.rb
@@ -70,7 +70,7 @@ class CohortList
     self.rows =
       data
         .map { |raw_row| raw_row.to_h.transform_keys { _1.downcase.to_sym } }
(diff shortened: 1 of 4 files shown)


app/controllers/cohort_lists_controller.rb
```diff
@@ -6,7 +6,9 @@ def new
   end
 
   def create
-    @cohort_list = CohortList.new(cohort_list_params)
+    @cohort_list = CohortList.new(
+      cohort_list_params.merge(team: current_user.team)
+    )
```
tvararu, line 11: If you rebase this then it shouldn't be necessary anymore, we're `.merge`ing team in the `cohort_list_params` method below on `main`

tvararu (approved):

PR #954, opened 2024-02-09
Move the cluster deletion protection to the addons/db

loladzeus:


misaka:
So this requires that we click "Modify" on a DB Cluster, untick "Delete protection", and then delete is, is that right? Do you know how it works for instances offhand? Do we need to remove delete protection on those to delete them too, or do we remove it from the cluster? Probably the later I guess, just curious.

Also, is this an alternative to the MFA deletion protection recommended in the pen test?

loladzeus:
@misaka this isn't an alternative to the MFA deletion protection, that was for the S3 buckets. this is basically moving one of the functionalities in the aws-account-setup script (the database cluster deletion protection) so that it's part of the infrastructure code, rather then part of the post-setup script.

regarding the deletion question - that's right, first we will need to modify the cluster to remove the deletion protection property, and then we can do the actual deletion. this is at a cluster level, so it applies to the individual databases that are a part of the cluster as well. in practice, if a database cluster has the deletion protection property enabled, then we cannot delete the cluster of the individual databases that are a part of it. disabling it momentarily will be the workaround.
I think this is a useful property to have in place, but it does interfere with automation - e.g. if we want the cluster to be deleted as part of the copilot app/service/environment tear down if that makes sense.

misaka (approved):

PR #958, opened 2024-02-12
Add an end-to-end test for the pilot

misaka:
To facilitate this test a new, empty campaign has been created which only has a location and some registrations linked to it. This also means a new nurse login, Nurse Flo <nurse.flo@example.com>.

Steps taken:

- [As parent]: Register interest
- [As nurse]: Close the registration
- Download registrants, edit out a couple
- Upload cohort list
- Create a new session selecting uploaded children
- Check that the details entered above show up on consent page

### DIFF
--- a/spec/lib/example_campaign_generator_spec.rb
+++ b/spec/lib/example_campaign_generator_spec.rb
@@ -32,11 +32,7 @@ RSpec.describe ExampleCampaignGenerator do
     # generated.
     #
     # To regenerate:
-    #     rails generate_example_campaign \
-    #       seed=42 \
-    #       presets=default \
-    #       type=hpv \
-    #       username="Nurse Joy" > db/sample_data/example-hpv-campaign.json
+    #     bin/generate_example_campaigns
     Timecop.freeze(2024, 2, 3, 12, 0, 0) do
       generator =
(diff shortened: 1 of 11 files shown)


spec/lib/example_campaign_generator_spec.rb
```diff
@@ -75,6 +81,34 @@
...
+    # When the fixture json is regenerated, e.g. when the generater is updated,
+    # the time here will need to be changed to match the day when it was
+    # generated.
+    #
+    # To regenerate:
+    #     rails generate_example_campaign \
+    #       seed=44 \
+    #       presets=empty_pilot \
+    #       username="Nurse Flo" > db/sample_data/example-pilot-campaign.json
+    Timecop.freeze(2024, 2, 8, 12, 0, 0) do
```
tvararu, line 99: We should probably combine this + above ones in a script so it's 1 manual step not 3...

tvararu (approved):

spec/lib/example_campaign_generator_spec.rb
misaka: Added commit

tvararu (approved):

PR #970, opened 2024-02-13
Deploy to test instead of staging

tvararu:
Staging is now more prod-like so we don't want continuous deployment.

Also rename ci.yml to deploy.yml.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Deploy
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -1,12 +1,12 @@
...
+name: Deploy
 
 on:
   push:
     branches: [main]
 
 jobs:
   # Permissions to request and use OIDC tokens and access repository contents
-  deploy:
+  copilot:
```
loladzeus, line 9: Can this bit only be tested upon merging?

loladzeus (approved):

.github/workflows/deploy.yml
misaka: This is setup to go to "test" so I would argue that this _is_ the test. 😀

misaka (approved):
LGTM

PR #982, opened 2024-02-15
Add user_research_observation_agreed boolean to Registration

tvararu:
This will be used to track whether users expressing interest in the pilot consent to their child being observed as part of the session.

### Screenshot

Screenshot 2024-02-15 at 12 41 37


db/schema.rb
```diff
@@ -301,6 +302,7 @@
     t.boolean "terms_and_conditions_agreed"
     t.boolean "data_processing_agreed"
     t.boolean "consent_response_confirmed"
+    t.boolean "observed_session_agreed"
```
benilovj, line 305: I wonder if this might read better if it were `agreed_to_user_research_observation`

db/schema.rb
```diff
@@ -235,6 +235,7 @@
     t.integer "team_id", null: false
     t.boolean "registration_open", default: false
     t.string "urn"
+    t.boolean "permission_to_observe_required"
```
benilovj, line 238: Could read slightly clearer if `involved_in_user_research_observation`
Very wordy though.

benilovj (commented):
Very very optional naming suggestions, as the existing ones made me pause a minute to parse.

db/schema.rb
misaka: Just my €0.02 ... I think `observed_session_agreed` -> `agreed_to_user_research_observation` gets us good value, more than this one.

misaka (approved):
:rocket: ... I think the first renaming would be good, the second would also help but a bit wordy.

db/schema.rb
tvararu: `user_research_observation_agreed` matches the other fields which all end in `_agreed`, shall I do that instead?

PR #990, opened 2024-02-17
Add more testing around the agree to observe checkbox

misaka:
It's important that the behaviour of the checkboxes the parents use to agree to having their child observed work reliably. This adds extra testing to ensure:

- if the parent does not check the observation checkbox (or any of the other agreement/confirmation checkboxes) they cannot join the pilot
- if the school does not require agreement to observation the checkbox does not appear

This also adds a new approach to generating fixtures by using a `TestingController` to try to simplify the process.

Questions about test coverage:
- We don't have a test to ensure that if a parent fills in the registration form but does _not_ check any of the required agreements / confirmations, their child is not added to the system. Worth adding?

### DIFF
--- a/tests/pilot_registration_checkboxes.spec.ts
+++ b/tests/pilot_registration_checkboxes.spec.ts
@@ -1,0 +1,157 @@
+import { test, expect, Page, APIRequestContext } from "@playwright/test";
+import { signInTestUser, fixtures } from "./shared";
+
+let p: Page;
+let r: APIRequestContext;
+let registrationPage: string;
+let locationName: string;
+
+test("Pilot registration - check-boxes", async ({ page, request }) => {
+  p = page;
+  r = request;
+
+  await given_a_location_requiring_permission_to_observe();
+  await and_registration_is_open();
+
+  await when_i_go_to_the_registration_page_for_a_school();
+  await and_i_enter_all_the_details();
+  await and_i_check_all_the_boxes_except_the_one_to_agree_to_take_part();
+  await and_i_click_submit();
+  await then_i_see_a_message_saying_i_must_agree_to_all_conditions();
+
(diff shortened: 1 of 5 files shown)


tests/pilot_registration_checkboxes.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+        permission_to_observe_required: true,
+        registration_open: true,
+      },
+    },
+    timeout: 10000,
+    headers: {
+      Accept: "application/json",
+    },
+  });
+  let json = await response.json();
```
tvararu, line 52: This way of doing test setup is clunky. I don't want to expand this pattern to other data types. Instead, I think we should have a single `/reset` endpoint that sets up a few hundred campaigns/teams so that every test can use isolated data and run in parallel. I think we should remove these tests and `TestingController` once we finish with the pilot EOI stuff.

tvararu (approved):

tvararu:
> We don't have a test to ensure that if a parent fills in the registration form but does not check any of the required agreements / confirmations, their child is not added to the system. Worth adding?

No, that shouldn't be necessary

tests/pilot_registration_checkboxes.spec.ts
misaka: I had a think about using a single `/reset`, but my concern is that setting up a few hundred campaigns would be slow, especially when running less than the full suite of tests. If you find the parameterisation clunky then the other approach I considered was to have `/testing/generate-pilot-campaign` with no parameters, but I opted for parameterisation because it does a better job of keeping some high-level setup with the test, similar to how our rspec tests work.

re: parallelisation, the approach above achieves this, and in fact protects better against clashing with other tests because tests won't have access to or know about the testing campaigns created for other tests.

PR #1005, opened 2024-02-20
Update deploy workflow with test env and adding name of env to job name

misaka:
image

Also found a way to update the run name displayed for the job:

image


tvararu (approved):
:ship:

PR #1012, opened 2024-02-21
Lift common mailer attrs out of ConsentFormMailer

tvararu:
Follow-up to https://github.com/nhsuk/manage-childrens-vaccinations/pull/1011, same idea, but taken a step further.

These can sit in the ApplicationMailer to help with code reuse across different mailers.

### DIFF
--- a/app/mailers/consent_form_mailer.rb
+++ b/app/mailers/consent_form_mailer.rb
@@ -19,86 +19,17 @@ class ConsentFormMailer < ApplicationMailer
 
   def opts(consent_form)
     @consent_form = consent_form
+    @patient = consent_form
+    @session = consent_form.session
 
-    { to:, reply_to_id:, personalisation: }
+    { to:, reply_to_id:, personalisation: consent_form_personalisation }
   end
 
-  def to
-    @consent_form.parent_email
-  end
-
-  def reply_to_id
-    @consent_form.session.campaign.team.reply_to_id
-  end
-
-  def personalisation
-    {
-      full_and_preferred_patient_name:,
-      location_name:,
-      long_date:,
-      observed_session:,
-      parent_name:,
-      reason_for_refusal:,
-      short_date:,
-      short_patient_name:,
-      short_patient_name_apos:,
-      team_email:,
-      team_phone:,
-      vaccination:
-    }
-  end
-
-  def full_and_preferred_patient_name
-    if @consent_form.common_name.present?
-      @consent_form.full_name + " (known as #{@consent_form.common_name})"
-    else
-      @consent_form.full_name
-    end
-  end
-
-  def short_patient_name
-    @consent_form.common_name.presence || @consent_form.first_name
-  end
-
-  def short_patient_name_apos
-    apos = "'"
-    apos += "s" unless short_patient_name.ends_with?("s")
-    short_patient_name + apos
-  end
-
-  def observed_session
-    @consent_form.session.location.permission_to_observe_required?
-  end
-
-  def location_name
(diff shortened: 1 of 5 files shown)


app/mailers/consent_form_mailer.rb
```diff
@@ -19,86 +19,17 @@ def confirmation_refused(consent_form)
 
   def opts(consent_form)
     @consent_form = consent_form
+    @patient = consent_form
```
misaka, line 22: This seems like a point of confusion to me. We have a `Patient` model so calling a `ConsentForm` "patient" doesn't feel right. I'm also not quite convinced about putting everything in `ApplicationMailer` here, tbh. I feel like it'd be fine, and easier to understand, if `personalisation` was defined in the mailer file here.

spec/mailers/consent_form_mailer_spec.rb
```diff
@@ -1,112 +1,15 @@
...
-    end
-  end
-
   describe "#confirmation_injection" do
-    let(:notify_template_id) { "4d09483a-8181-4acb-8ba3-7abd6c8644cd" }
-
     it "calls template_mail with correct personalisation" do
+      allow_any_instance_of(Mail::Notify::Mailer).to receive(
+        :template_mail
+      ).with(kind_of(String), ->(options) { @template_options = options })
```
misaka, line 8: btw, there's another way to pull out the personalisation that doesn't rely on mocking. From a test I'm working on (`ConsentMailer` is the working name for the consent requests/reminders mailer):

```
      mail = ConsentMailer.consent_request(patient, session)
      expect(mail.message.header["personalisation"].unparsed_value).to(
        include(
          short_date: session.date.strftime("%-d %B")
        )
      )
```

Might be cleaner than the stubbing approach.

misaka (approved):
Ok to ship, but I'm worried naming and abstracting away how personalisation is built may be confusing. Let's roll forward and we'll see how bad it is as we implement new mailers.

app/mailers/consent_form_mailer.rb
tvararu: It's because some of the fields in ApplicationMailer rely on `@patient`, but `@patient` and `@consent_form` share many fields. It's not perfect but the alternative is to potentially duplicate a lot of code

misaka: I wonder if there's a better name than `@patient`, though. `@patient_or_consent_form` is longer but would be clearer. `@object` or `@record` is a bit abstract but would at least avoid the trap of assuming it's a `Patient` record ... ugh. Naming is hard.

But what I was saying is that I think DRYing makes it less clear. I think it would be better to build `personalisation` in this file, and DRY things up by using help methods (where needed). For example (note `short_date`, `short_patient_name` and `short_patient_name_apos`):

```
  def personalisation
    {
      full_and_preferred_patient_name: ...,
      location_name: @session.location.name,
      long_date: @session.date.to_fs(:app_short_date),
      observed_session: ...,
      parent_name: @consent_form.parent_name,
      short_date: @session.date.to_fs(:app_short_date),
      short_patient_name: short_patient_name(@consent_form),
      short_patient_name_apos: short_patient_name_apos(@consent_form),
      team_email: @session.team.email,
      team_phone: @session.team.phone,
      vaccination: ...
    }
  end
```

tvararu: I see your point, I think your suggestion would work as well. I considered something like that, but thought I'd try the Superdry (TM) version first. Let's try this version for now and if it's clunky we can roll forward to a wetter medium.

PR #1013, opened 2024-02-21
Add phone field to team

tvararu:
This will be used to populate emails sent to parents so they know how to get in contact.


benilovj (approved):
Do we want to validate that it's an actual phone number, to reduce the chances of fat-fingering?

benilovj (approved):

PR #1019, opened 2024-02-21
Improve small-screen appearance of session tab when a table has no results

paulrobertlloyd:
- Use a heading for each tab (visually hidden from desktop breakpoint), instead of a table caption, and show this above both the filters and table on smaller viewports
- Remove card component for no results and align appearance with that for no _filtered_ results
- Show filter sidebar on mobile, but hide all filters except name search until tablet breakpoint

This should work better than the current implementation, but probably need to move away from tabs to instead having separate pages.

Also, need to properly update the tests `AppEmptyListComponent`


benilovj:
Is SonarCloud wanting captions on the tables?

app/views/vaccinations/_patients_with_actions.html.erb
```diff
@@ -1,26 +1,28 @@
...
+      </tbody>
+    </table>
+    <action-table-no-results>
+      <%= render AppEmptyListComponent.new(
+        text: t("table.no_filtered_results")
+      ) %>
+    </action-table-no-results>
   </div>
-</action-table>
+</action-table>
```
tvararu, line 28: @paulrobertlloyd can you configure your editor to include a trailing newline at the end of each file? That's why GitHub is showing a red circle with a dash in the middle

tvararu:
@benilovj it is, but we should replace those tables with a component. It's not urgent to do so I'd just ignore or WONTFIX the Sonarcloud issues.

tvararu (approved):

PR #1021, opened 2024-02-21
Move link to unmatched responses page

benilovj:
It used to be on the session page, now it's on the 'check consent responses' page

## Sessions page

There is no more link here to the unmatched responses page:

image

## Check consent responses

The link is now here:

image


misaka (approved):
🚀

PR #1022, opened 2024-02-21
Add job to send consent requests for sessions

misaka:
This is email sent out by the scheduled job to request that parents submit consent for their children. This PR adds a job `ConsentRequestsJob` that can be scheduled to run daily at an appropriate time. It will trigger another job, `ConsentRequestsSessionBatchJob` which will send an email to the parent of each patient in a session that has not yet been sent a consent request.

The cron scheduler hasn't been setup yet, this will be done after we do some more testing of how Good Job's scheduled jobs work.


tvararu (approved):
🙌

PR #1023, opened 2024-02-22
Add consent reminder email

misaka:
This will be used to send consent request reminders to patient's parents.

Also includes an update to the initial consent request email to differentiate between the session and close consent dates.

Both templates have been tested.

### DIFF
--- a/config/initializers/date_formats.rb
+++ b/config/initializers/date_formats.rb
@@ -2,3 +2,5 @@
 Date::DATE_FORMATS[:nhsuk_date] = "%-d %B %Y"
 Date::DATE_FORMATS[:nhsuk_date_day_of_week] = "%A %-d %B %Y"
 Date::DATE_FORMATS[:nhsuk_date_short_month] = "%-d %b %Y"
+Date::DATE_FORMATS[:app_date] = "%A %-d %B"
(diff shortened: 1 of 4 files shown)


config/initializers/date_formats.rb
tvararu, line 1: Wait, was this file re-introduced? I thought I deleted it in favour of `config/initializers/date_time_formats.rb` which just has them all in one place to reduce confusion... Maybe I didn't delete it, but could we do that please?

spec/mailers/consent_request_mailer_spec.rb
```diff
@@ -11,8 +11,50 @@
...
+                 close_consent_date:
+                   session.close_consent_at.strftime("%A %-d %B")
+               )
+      end
+      it do
+        should include(
+                 close_consent_short_date:
+                   session.close_consent_at.strftime("%-d %B")
+               )
+      end
```
tvararu, line 57: This indentation brought to you by Prettier(TM) 😅

tvararu (approved):

config/initializers/date_formats.rb
misaka: Sure. And I think we probably could use a cleanup there, I'm not sure whether to use app_ or nhsuk_ prefixes, etc

PR #1025, opened 2024-02-22
Bring unmatched consent responses page into line with latest designs

benilovj:
Changes:

* moving the page's path from `/schools/XX` to `/sessions/XX/consents/unmatched-responses` to reflect that it's part of the `Check consent responses` area of the product. "Unmatched consent responses" under schools doesn't feel quite correct, as a school will have multiple sessions (i.e. unmatched responses are session-specific, not school-specific)
* updating breadcrumbs
* updated designs

## Before

image

## After

image

### DIFF
--- a/tests/schools_match_response.spec.ts
+++ b/tests/schools_match_response.spec.ts
@@ -51,9 +51,16 @@ async function then_i_am_on_the_unmatched_responses_page() {
 }
 
 async function when_i_click_on_an_unmatched_response() {
-  await p
-    .getByRole("link", { name: fixtures.unmatchedConsentFormParentName })
-    .click();
+  const row = await p
+    .locator(
+      `//td[contains(., '${fixtures.unmatchedConsentFormParentName}')]/parent::tr`,
+    )
+    .first();
+  if (row) {
+    await row.getByRole("link", { name: "Find match" }).click();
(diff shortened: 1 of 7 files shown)


benilovj:
I just noticed that:

* the table should probably be sorted initially in the backend
* the table needs the filters

I'll add that now.

tests/schools_match_response.spec.ts
```diff
@@ -46,14 +46,21 @@ async function and_i_click_on_the_unmatched_responses_link() {
...
 
 async function when_i_click_on_an_unmatched_response() {
-  await p
-    .getByRole("link", { name: fixtures.unmatchedConsentFormParentName })
-    .click();
+  const row = await p
+    .locator(
+      `//td[contains(., '${fixtures.unmatchedConsentFormParentName}')]/parent::tr`,
+    )
+    .first();
```
tvararu, line 58: That's a funky selector

tvararu (approved):

tests/schools_match_response.spec.ts
benilovj: yeah i'm sure there's a better way of doing it, i got an assist from my good friend who works at OpenAI.

PR #1028, opened 2024-02-22
Send consent requests on session create

misaka:
Quick and dirty, testing will have to be sorted out later

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -525,7 +525,7 @@ en:
           attributes:
             send_consent_at:
               blank: Choose a date to send consent requests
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -525,7 +525,7 @@ en:
           attributes:
             send_consent_at:
               blank: Choose a date to send consent requests
-              greater_than_or_equal_to: Enter a date in the future
+              greater_than_or_equal_to: Enter a date thats not in the past
```
tvararu, line 528:
```suggestion
              greater_than_or_equal_to: Enter a date that is not in the past
```

tvararu (approved):

PR #1035, opened 2024-02-22
Match consent and consent_form reason_for_refusals

tvararu:
Make them consistent so creating consents from consent_forms works OOTB.


misaka (approved):
¯\_(ツ)_/¯ 🚢

PR #1040, opened 2024-02-22
Rudimentary manual matching of consent forms to patient records in the cohort

benilovj:
## Search for a child

search-for-a-child

## Link consent response with child record?

image

## Not in this PR – will be added in subsequent PRs

* success banner at the end of the transaction
* actual useful information on the "Link consent response with child record?"
* archiving of consent forms


app/controllers/consent_forms_controller.rb
```diff
@@ -1,30 +1,44 @@
...
-    @session =
-      policy_scope(Session).find(
-        params.fetch(:session_id) { params.fetch(:id) }
-      )
+  def match
+    @consent_form = policy_scope(ConsentForm).find(params[:id])
+    @patient_session =
+      policy_scope(PatientSession).find(params[:patient_session_id])
+
+    Consent.from_consent_form(@consent_form, @patient_session)
```
tvararu, line 32: `from_consent_form` was built in a pretty rushed way. I think I'd like to amend it to either be a `!` method as it introduces a couple of side effects, or probably more flexibly, it should return a consent that you can then `.save` if you wish. Also the state machine stuff in it should probably be taken out and left to the caller.

app/views/consent_forms/show.html.erb
```diff
@@ -5,16 +5,28 @@
...
-    <%= render AppCardComponent.new(heading: "Consent") do
-      render AppConsentDetailsComponent.new(consents: [@consent_form])
-    end %>
-  </div>
-</div>
+<%= render AppCardComponent.new(heading: "Children in this cohort") do
+  if @patient_sessions.count > 0
+    render AppPatientTableComponent.new(
+      patient_sessions: @patient_sessions,
+      tab_id: '123',
```
tvararu, line 24: Temporary?

config/routes.rb
```diff
@@ -127,7 +127,16 @@
       get "confirmation", on: :collection
     end
   end
-  resources :consent_forms, only: [:show]
+  resources :consent_forms, only: [:show] do
```
tvararu, line 130: We should avoid underlines in URLs:

```suggestion
  resources :consent_forms, path: "consent-forms", only: [:show] do
```

tvararu (approved):

config/routes.rb
benilovj: good point - fixed

app/controllers/consent_forms_controller.rb
benilovj: I've added a bang in [d32d318](https://github.com/nhsuk/manage-childrens-vaccinations/pull/1040/commits/d32d318166b42b695866a11ca175539a5c73d949). As discussed, we can roll forward and improve it when we do more work on it.

PR #1042, opened 2024-02-23
Add feature flag to hide gillick competence journey

tvararu:
We need to hide this in certain environments.


benilovj:
we could hide it until the day of the vaccination – that would save having another feature flag.

tvararu:
> we could hide it until the day of the vaccination – that would save having another feature flag.

Thanks, I wasn't sure what the intention was of the original snag, but figured someone would mention it in the PR phase 😄 

Will update!

benilovj (approved):

PR #1045, opened 2024-02-23
Protect good-job dashboard with basic auth

misaka:
This job will send out consent request emails to parents on the configured day for each session.

### DIFF
--- a/config/initializers/good_job.rb
+++ b/config/initializers/good_job.rb
@@ -1,0 +1,23 @@
+# Prompt for a username if either the support user is defined, or we are in
+# production. In production, if no support user is defined then no access is
+# available.
+
+support_user_defined =
+  Settings.support_username.present? && Settings.support_password.present?
+
+if Rails.env.production? || support_user_defined
+  GoodJob::Engine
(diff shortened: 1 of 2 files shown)


config/initializers/good_job.rb
```diff
@@ -0,0 +1,23 @@
+# Prompt for a username if either the support user is defined, or we are in
+# production. In production, if no support user is defined then no access is
+# available.
+
+support_user_defined =
+  Settings.support_username.present? && Settings.support_password.present?
```
tvararu, line 6: Support user creds are currently set via `manifest.yml` and are hard-coded in git for prod. Can we move them to `SECRETS` instead? That should be done as a separate PR/deploy to first roll it out in every env, followed by this PR to add the `/good-job` dashboard.

tvararu (changes requested):

tvararu (approved):
👍 creds now moved to secrets so this can go in after we check that all is good with the other release

misaka:
This was manually deployed to test and it is working as expected.

PR #1046, opened 2024-02-23
Useful content to help manual matching

benilovj:
image


benilovj:
@paulrobertlloyd the designs also had the GP name as an attribute for comparison, but I discovered we don't have that on the patient record 🙀 so I've omitted it for now.

tvararu (approved):
Main thing missing from this is a test for the new component, but 🚢

benilovj:
> Main thing missing from this is a test for the new component, but 🚢

Added in https://github.com/nhsuk/manage-childrens-vaccinations/pull/1054

PR #1047, opened 2024-02-23
Use pundit policies in all necessary controllers

misaka:
This is a security measure to ensure we always scope the DB requests appropriately to the current user.

We also fix the data as many patients were being created without a link to location.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,6 +9,9 @@ class ApplicationController < ActionController::Base
   before_action :set_phase_banner_text
   before_action :authenticate_basic
 
+  after_action :verify_policy_scoped,
+               if: -> { Rails.env.development? || Rails.env.test? }
+
   class UnprocessableEntity < StandardError
   end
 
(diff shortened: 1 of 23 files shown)


benilovj (approved):
Code looks fine but @tvararu should review too

app/controllers/application_controller.rb
```diff
@@ -9,6 +9,9 @@ class ApplicationController < ActionController::Base
   before_action :set_phase_banner_text
   before_action :authenticate_basic
 
+  after_action :verify_policy_scoped,
+               if: -> { false && Rails.env.development? || Rails.env.test? }
```
tvararu, line 13: Is the `false &&` just for testing? 🤔

tvararu (approved):
1 comment, rest LGTM

app/controllers/application_controller.rb
misaka: Aaahh, ya. doh! I was just forcing it to test that path, good spot thanks!

misaka:
I'm just going to ignore the SonarCloud code duplication issue ... I couldn't even make heads or tails of what it was complaining about specifically, never mind how to dismiss it.

PR #1049, opened 2024-02-24
Add automatic matching

tvararu:
This attempts to match a consent form to a patient using the matching criteria we've defined for the pilot. If any 3 of first_name, last_name, date_of_birth, address_postcode are exact matches, then we can safely assume that the ConsentForm belongs to that particular Patient record.

A few things are left to be done in follow-up work:

- Case insensitive matching
- Additional criteria (such as using common_name)
- ~Creating Consent records (combined with Consent#from_consent_form)~ ✅ 

See first 2 commits as well for some related work.


benilovj:
New E2E test is now reliably failing on an empty database, while passing after a reset dataset. Investigating whether it's a factory issue...

benilovj:
OK, next problem – it doesn't appear that the jobs are running during the test, so there's an ugly hack in the middle of the feature spec.

misaka:
> ~OK, next problem – it doesn't appear that the jobs are running during the test, so there's an ugly hack in the middle of the feature spec.~ Solved by replacing the `.perform_later` with `.perform_now`.

I think we want this job to be performed out-of-band. It might be ok size-wise for the pilot, but in general this would be better OOB ... beyond blocking issues, if any errors happen I don't think it makes sense to expose the user to them.

About the testing, did you try `perform_enqueued_jobs`?

benilovj:
> > ~OK, next problem – it doesn't appear that the jobs are running during the test, so there's an ugly hack in the middle of the feature spec.~ Solved by replacing the `.perform_later` with `.perform_now`.
> 
> I think we want this job to be performed out-of-band. It might be ok size-wise for the pilot, but in general this would be better OOB ... beyond blocking issues, if any errors happen I don't think it makes sense to expose the user to them.
> 
> About the testing, did you try `perform_enqueued_jobs`?

Yes thanks @misaka that's exactly what I was looking for.

misaka:
> > > ~OK, next problem – it doesn't appear that the jobs are running during the test, so there's an ugly hack in the middle of the feature spec.~ Solved by replacing the `.perform_later` with `.perform_now`.
> 
> > 
> 
> > I think we want this job to be performed out-of-band. It might be ok size-wise for the pilot, but in general this would be better OOB ... beyond blocking issues, if any errors happen I don't think it makes sense to expose the user to them.
> 
> > 
> 
> > About the testing, did you try `perform_enqueued_jobs`?
> 
> 
> 
> Yes thanks @misaka that's exactly what I was looking for.

Now that we're in Ruby land a lot of these things get a lot easier. 😅

spec/features/parent_gives_consent_spec.rb
```diff
@@ -0,0 +1,350 @@
+require "rails_helper"
+
+module SessionCreationSteps
```
benilovj, line 3: This module is an experiment. It's currently only used in this spec but I suspect it's going to be reusable beyond this spec, and so could be extracted to a helper file. It could also help maintenance – if microcopy or journey tweaks are made, we don't want lots of feature specs breaking.

misaka: Just a couple thoughts ...

- I think I much prefer it when the top of the file is the journey being tested, so I do think if we keep this Steps module we should put it in a separate file
- I generally don't mind having a file with common steps, but I think it should be used sparingly: while it was a slight pain with the Playwright tests to update multiple tests for a small front-end change, I think it was better than having to remember all the available step methods when building a test. I'd probably err on the side of less shared steps

tests/parental_consent.spec.ts
```diff
@@ -1,201 +0,0 @@
...
-    "Has your child been diagnosed with asthma?",
-  );
-}
-
-async function when_i_submit_all_the_health_questions() {
-  for (let i = 1; i <= 8; i++) {
-    await p.getByRole("radio", { name: "No" }).click();
-    await and_i_click_continue();
-  }
-}
```
misaka, line 201: 🎉

spec/features/parent_gives_consent_spec.rb
```diff
@@ -0,0 +1,350 @@
...
+
+  def and_i_submit_the_consent_form
+    click_on "Confirm"
+    expect(page).to have_content(
+      "#{@child.full_name} will get their HPV vaccination at school"
+    )
+  end
+
+  def and_i_get_a_confirmation_email
+    # not sure how to do this yet
```
misaka, line 323: A couple ways I can think of testing this.
- Use the rspec helper `have_enqueued_mail` to see if the correct `Mail` object is in the queue
- Use the rspec helper `have_been_delivered` to check if the correct `Mail` objects was delivered (e.g. was enqueued and then the queue was processed, I think?)
- Examine `ActionMailer::Base.deliveries` yourself, e.g. https://github.com/nhsuk/manage-childrens-vaccinations/blob/c2ea34b218196053277ee735fa8b0c9b6505c211/spec/jobs/consent_requests_session_batch_job_spec.rb. You can also examine the headers of each `Mail` object that was "delivered" if that's what you need.

spec/features/parent_gives_consent_spec.rb
benilovj: > I think I much prefer it when the top of the file is the journey being tested, so I do think if we keep this Steps module we should put it in a separate file

yes, agreed. I've added that in [2a37357](https://github.com/nhsuk/manage-childrens-vaccinations/pull/1049/commits/2a373573ebc789dca47945b719c55dc3ceb22955)

> I generally don't mind having a file with common steps, but I think it should be used sparingly: while it was a slight pain with the Playwright tests to update multiple tests for a small front-end change, I think it was better than having to remember all the available step methods when building a test. I'd probably err on the side of less shared steps

Yep, have definitely been in that "pile of steps" scenario on previous projects and it's another type of maintenance nightmare.

I think there's a balance that could be struck if we don't go overboard with it; I think there's a fair amount of value in being able to tersely define a journey that maybe isn't core to what's being specced in that particular scenario – kinda like how you can skip a lot of the details when using FactoryBot factories if you don't care about them. Here's an example of what would be nice to do:

```ruby
  def and_the_sais_team_has_organised_a_session_at_my_childs_school
    sign_in_as_nurse_testy
    start_new_session(
      school_name: @school.name,
      session_date: 1.week.from_now,
      time_of_day: "Morning"
    )
    select_cohort(names: [@child.full_name])
    schedule_key_dates(send_consent_on_date: Time.zone.today)
    confirm_session_details(
      number_of_children: 1,
      send_consent_on_date: Time.zone.today,
      session_date: 1.week.from_now
    )
  end
```

I guess we could rely on factories to achieve the same thing, but having fixed a few factory bugs recently, it's nice to be able to do stuff just through the frontend.

spec/features/parent_gives_consent_spec.rb
benilovj: Thanks. I've given it a go here: 5cc0e7f3

spec/features/parent_gives_consent_spec.rb
benilovj: Updated here [2a37357](https://github.com/nhsuk/manage-childrens-vaccinations/pull/1049/commits/2a373573ebc789dca47945b719c55dc3ceb22955)

tvararu: @benilovj I think you've got the right idea 👍 

**Do:** abstract reusable setup and repetitive code that gets used **inside** steps

```ruby
# spec/system/foo_spec.rb
def when_i_am_signed_in
  sign_in_as_nurse_testy
end

# spec/system/bar_spec.rb
def and_i_am_signed_in
  sign_in_as_nurse_testy
end

# spec/system/baz_spec.rb
def then_i_sign_in
  sign_in_as_nurse_testy
end
```

**Don't**: abstract the **actual** steps, those are more like documentation and should be more flexible

```ruby
# DONT:
include SharedSteps

# ...
when_i_am_signed_in
```

tvararu:
@benilovj the scope of this PR has expanded and now it's pretty big; what do you think about splitting it up so that a first PR introduces the `find_matching_patient` method + test and then a subsequent one uses it + tests it e2e?

benilovj (approved):

PR #1052, opened 2024-02-25
Only set run name on workflow_dispatch events

misaka:
This is an attempt to fix the run name being set to "Deploy to " for the deploy GitHub actions. After some testing `null` appears to restore the default behaviour, so we use that if we've been triggered by anything other than a `workflow_dispatch` event.

The use of a `null` value has been tested, but the real test will be when this gets deployed on PR merge.


tvararu (approved):
🧪🧑‍🔬

PR #1061, opened 2024-02-26
Track sent_reminder_at on patient model 

tvararu:
To make the ConsentReminderSessionBatchJob idempotent, we need to track when this has successfully run on a particular patient record.

Also ensures that we don't send reminders where a parent has already given consent. It only works if automatic watching works or a nurse does a manual match before the reminders are scheduled to go out.


app/jobs/consent_reminders_session_batch_job.rb
```diff
@@ -12,9 +12,9 @@ class ConsentRemindersSessionBatchJob < ApplicationJob
   queue_as :default
 
   def perform(session)
-    session.patients.consent_not_sent.each do |patient|
+    session.patients.without_consent.reminder_not_sent.each do |patient|
```
benilovj, line 15: It might be slightly cleaner / more encapsulated to push this logic by defining a 3rd scope on `Patient`:

```ruby
scope :needing_consent_reminder, -> { without_consent.reminder_not_sent }
```

benilovj (approved):
LGTM, non-blocking suggestion.

app/jobs/consent_reminders_session_batch_job.rb
tvararu: Great suggestion!

PR #1064, opened 2024-02-26
Add db:prepare:ignore_concurrent_migration_exceptions

tvararu:
Because we deploy to multiple machines at the same time, it's possible to run into concurrent migration issues. It's safe to swallow these errors and let the deploy carry on.

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -2,7 +2,7 @@
 
 # If running the rails server then create or migrate existing database
 if [ "${*}" == "./bin/rails server" ]; then
-  ./bin/rails db:prepare
+  ./bin/rails db:prepare:ignore_concurrent_migration_exceptions
 fi
 
 exec "${@}"
--- a/bin/release
+++ b/bin/release
@@ -6,5 +6,5 @@ if [[ "$MAVIS__IS_REVIEW" == "true" && `bin/rails db:version | grep version` == 
   bin/rails db:schema:load
   bin/rails db:seed
 else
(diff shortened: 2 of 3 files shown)


benilovj (commented):
According to the Rails Guides, `db:prepare` will load seeds in certain cases. Has this happened in prod already?

tvararu:
@benilovj ah TIL, I thought it just ran `db:setup && db:migrate` based on the description:

```
$ bin/rails db:prepare --help
bin/rails db:prepare
    Run setup if database does not exist, or run migrations if it does
```

Our seeds only contain some feature flagging stuff. I'm wondering now if we should switch to a normal `db:migrate` setup instead of `db:prepare`, or stick to `db:prepare` and potentially even make use of the seeds in the future should we need to add stuff that should always be in production

misaka (approved):
+1 for using `db:migrate` (until we have a use for seeds in prod), but I think this looks safe to ship for now.

benilovj (approved):
So after looking in the Rails source code [1] [2]:

[1]: <https://github.com/rails/rails/blob/main/activerecord/lib/active_record/railties/databases.rake>
[2]: <https://github.com/rails/rails/blob/main/activerecord/lib/active_record/tasks/database_tasks.rb#L176>

I *believe* seeds will only be run if the schema migrations table doesn't exist yet. So to me it looks safe to call `db:prepare` here... 🤷

tvararu:
@benilovj thanks for digging that up! Let's use this then :shipit:

PR #1073, opened 2024-02-27
Remove contact method page from parental consent

misaka:
At the moment the service won't be able to honour this if set.

This also removes the form page and validation, since they're easy to re-add, but leaves the db field in case we want to populate it another way at some point.

### DIFF
--- a/app/models/consent_form.rb
+++ b/app/models/consent_form.rb
@@ -143,7 +143,7 @@ class ConsentForm < ApplicationRecord
     validates :parent_phone, phone: true, if: :parent_phone?
   end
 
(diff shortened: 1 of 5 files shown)


benilovj:
I'm in two minds about this; if we were to rip everything out, we should also remove stuff like `Consent#parent_contact_method_other`. An alternative could be to feature flag this and roll forward to complete the feature in future – the feature flag will remind us that it's incomplete.

tvararu:
Yeah, maybe let’s feature flag it. Make it off by default to save us the admin of going in and updating the envs.

tvararu (approved):

PR #1074, opened 2024-02-27
Use parent name from consent in triage email

misaka:
The consent object gets the parent's name from the consent form or from the nurse when they call the parent. This is more likely to be accurate and up-to-date than what's on the patient record.

There's still the question of what to do when more than one parent has submitted consent, how do we know we have the right parent?

### DIFF
--- a/app/mailers/triage_mailer.rb
+++ b/app/mailers/triage_mailer.rb
@@ -1,5 +1,7 @@
 class TriageMailer < ApplicationMailer
   def vaccination_will_happen(patient_session)
+    @patient_session = patient_session
+
     template_mail(
       "fa3c8dd5-4688-4b93-960a-1d422c4e5597",
       **opts(patient_session)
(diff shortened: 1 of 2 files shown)


app/mailers/triage_mailer.rb
```diff
@@ -1,15 +1,33 @@
...
   end
+
+  private
+
+  def consent
+    @patient_session.consents.order(:created_at).last
+  end
+
+  def to
+    consent.parent_email
```
benilovj, line 27: This will blow up if `consent` is nil...

tvararu: This shouldn't trigger... happy to merge this

tvararu (approved):

app/mailers/triage_mailer.rb
misaka: Also, if we protect this it will fail silently when trying to send. I thought blowing up in these extremely unlikely situations would be better.

PR #1083, opened 2024-02-27
Run good_job as an external service

tvararu:
This changes from :async to :external and spins up a new service in our Copilot infrastructure.


misaka (approved):
Tested and working in `test`

PR #1086, opened 2024-02-28
Add Timecop freeze to pilot journey test

tvararu:
Validations started failing in this test because we went past the 27th of Feb and certain dates can't be in the past.


benilovj (changes requested):
I think it needs an unfreeze or it will affect other tests

tvararu:
@benilovj updated, I decided to use before/after blocks in the end. I think they're fine just for this specific instance

PR #1089, opened 2024-02-28
Remove FHIR-related code

tvararu:
This is old and unmaintained at this point.


benilovj (approved):
😢 it's not old, it's "heritage"!

PR #1096, opened 2024-02-29
Tidy up Heroku setup 

tvararu:
Remove some of the redundant review app env vars, and rename `bin/release` to `bin/heroku-release` to make it more obvious that it's not in use for anything else.

To check this PR, check the review app works as expected.

The basic auth for review apps is now `manage/vaccinations`, not `test/test`.

Also changes the `phonelib` media-only phone RegExp to use a Setting instead of ENV detection, to allow us to more easily toggle it in certain envs.


tvararu:
This is finally working correctly 🚀 

Screenshot 2024-02-29 at 10 21 22

PR #1097, opened 2024-02-29
Set good_job to async in production 

tvararu:
This simplifies our infrastructure and deploy pipeline, and should be resilient now that we found the Puma restart bug that previously led to jobs not being enqueued successfully in production.

First commit sets up Puma to cleanly restart good_job when it restarts following the advice in https://github.com/bensheldon/good_job?tab=readme-ov-file#execute-jobs-async--in-process

Second makes `test` more similar to `staging`, so that we can test this more thoroughly and also shut down staging once we switch to using `test` more (after this change ships successfully most likely).

Third removes the `good_job` execution mode override in `production.rb`, finishing the transition.

Follow-up PR to this will remove the `goodjob/manifest.yml` service and infrastructure.

This also fixes job execution on Heroku which is currently broken.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -119,7 +119,6 @@ Rails.application.configure do
   }
 
   config.good_job.enable_cron = true
-  config.good_job.execution_mode = :external
   config.good_job.cron = {
     consent_request: {
       cron: "every day at 9am",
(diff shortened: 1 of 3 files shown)


misaka (approved):
YOLO

PR #1104, opened 2024-02-29
Allow SAIS team members to add vaccine batches, see existing batches

benilovj:
## New link in the header

image

## Dashboard

image

## Manage vaccines

image

## Add a batch

image

## Add a batch with errors

image

## Batch added successfully

image

## Notes for reviewers

This PR includes another experiment in the feature spec; instead of using FactoryBot to set up data, I'm calling operational Rake tasks (as this is how teams / vaccines / users are set up in production-like environments). This should hopefully mean that this feature spec is more likely to spot issues with these Rake tasks, and less prone to suffer from factory bugs. The  Rake task calls are wrapped in a method that temporarily suppresses STDOUT output, to prevent stray noise in the test runner.

### DIFF
--- a/spec/features/manage_batches_spec.rb
+++ b/spec/features/manage_batches_spec.rb
@@ -1,0 +1,108 @@
+require "rails_helper"
+
+require "rake"
+
+RSpec.feature "Manage batches", type: :feature do
+  before do
+    Rails.application.load_tasks
+    Timecop.freeze(Time.zone.local(2024, 2, 29)) # so we don't worry about expiry dates
+  end
+
+  after { Timecop.return }
+
+  scenario "Add a new batch" do
+    given_my_team_is_running_an_hpv_vaccination_campaign
+
+    when_i_manage_vaccines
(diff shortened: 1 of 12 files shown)


paulrobertlloyd:
@benilovj Added a utility class for the header navigation that can be put on the item that pushes the other items to the right. And then added this to the new ‘Manage vaccines’ navigation item, pushing this over to the left grouping.

benilovj:
Thanks for sorting @paulrobertlloyd !

spec/features/manage_batches_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    suppress_output do
+      Rake::Task["add_new_user"].invoke(
+        "nurse.testy@example.com",
+        "nurse.testy@example.com",
+        "Nurse Testy",
+        created_team.id,
+        ""
+      )
+    end
```
tvararu, line 47: This is not the ideal way to set up stuff. I have some testing improvements in a branch that I'd like to adopt across. I'll update this after merging. 👍

tvararu (approved):

PR #1108, opened 2024-03-01
Transition to do_not_vaccinate after verbal consent refused

tvararu:
If a nurse records a verbal consent refused from a parent, we should transition the patient session to the do not vaccinate state, instead of leaving them in triage.

### After recording verbal consent as refused

Screenshot 2024-03-01 at 11 06 11


app/controllers/manage_consents_controller.rb
```diff
@@ -77,6 +77,10 @@ def handle_confirm
       @consent.recorded_at = Time.zone.now
       @consent.save!
 
+      if @consent.response_refused?
+        @triage.update! status: "do_not_vaccinate", user: current_user
+      end
```
tvararu, line 82: This can be done more cleanly, I will come back to it in this PR in a bit...

spec/features/manage_consent_refused_spec.rb
```diff
@@ -11,7 +11,7 @@
 
     # TODO: This is a bug in the app, their status should be do not vaccinate
```
misaka, line 12: Sounds like this can be removed now

app/controllers/manage_consents_controller.rb
misaka: What state is the `patient_session` in , and why doesn't it need to change?

benilovj (approved):

app/controllers/manage_consents_controller.rb
tvararu: It goes into `consent_refused` but in that state we can still edit consent. It should go into `triaged_do_not_vaccinate` which requires a triage, so that's what this PR is adding.

tvararu: In the interest of speed I'm leaving this in, but note that I think this `handle_confirm` method could be refactored a bit to make it clearer.

misaka: Ah, I see, it updates `triage` and then runs the state transitions below. 👍

misaka (approved):

PR #1109, opened 2024-03-01
Show nurse who recorded verbal consent 

tvararu:
Depends on https://github.com/nhsuk/manage-childrens-vaccinations/pull/1110.

This uses the recorded_by attribute which captures the user who handled the verbal consent.

Must be deployed only after all environments have `recorded_by` populated on `Consent`.

### Before (Parent)

Screenshot 2024-03-01 at 12 49 55

### After (Nurse Joy)

Screenshot 2024-03-01 at 12 49 44

### DIFF
--- a/spec/components/app_consent_response_component_spec.rb
+++ b/spec/components/app_consent_response_component_spec.rb
@@ -30,9 +30,10 @@ RSpec.describe AppConsentResponseComponent, type: :component do
   end
 
   context "with consent taken over the phone" do
-    let(:consent) { create(:consent, :given, route: "phone") }
+    let(:consent) { create(:consent, :given_verbally) }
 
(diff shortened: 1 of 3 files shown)


spec/components/app_consent_response_component_spec.rb
```diff
@@ -30,9 +30,10 @@
   end
 
   context "with consent taken over the phone" do
-    let(:consent) { create(:consent, :given, route: "phone") }
+    let(:consent) { create(:consent, :given_verbal) }
```
misaka, line 33: Just for the readingz

```suggestion
    let(:consent) { create(:consent, :given_verbally) }
```

misaka (approved):

PR #1113, opened 2024-03-01
Prevent users from double-tapping on stand-alone buttons

misaka:
We've confirmed that double-tapping these stand-alone buttons cause errors to show, even if the actual action is performed. Usually the first POST/PUT works, but the subsequent one is the one that error and that the user sees.

This change relies on https://github.com/x-govuk/govuk-components/pull/513 being accepted and merged (which seems imminent), and released. Until then we are using the fork I created get the functionality now.


tvararu (approved):
🎉

PR #1116, opened 2024-03-02
Before vaccination, open the health questions if there are notes

benilovj:
As they're ultimately the responsible clinician, vaccinators will often double check what notes the parents left as part of a light-touch triage just before vaccination, even if triage has already been done.

To support this, we should "unfold" the health questions whenever there are notes, before vaccination. If this isn't the case, vaccinators would have to manually open every patient's health questions to see if there are any notes there or not.

### DIFF
--- a/app/components/app_consent_component.rb
+++ b/app/components/app_consent_component.rb
@@ -16,7 +16,8 @@ class AppConsentComponent < ViewComponent::Base
   end
 
   def open_health_questions?
-    @patient_session.next_step == :triage
+    @patient_session.consents.any?(&:health_answers_require_follow_up?) &&
+      @patient_session.vaccination_records.administered.none?
   end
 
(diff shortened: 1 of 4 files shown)


paulrobertlloyd:
Makes sense to me; perhaps can close after vaccination has occurred, but not a pressing change.

benilovj:
> Makes sense to me; perhaps can close after vaccination has occurred, but not a pressing change.

Thanks @paulrobertlloyd, I've made this change in [b40813a](https://github.com/nhsuk/manage-childrens-vaccinations/pull/1116/commits/b40813ae7f9484dc0e236315565b678abbadc730).

Just a though – could it be worth changing the details summary to indicate whether the (consolidated) health responses have any notes or not?

tvararu (approved):

PR #1118, opened 2024-03-03
Enable component previews

benilovj:
This will allow to develop or debug some of the more complex view components in the app, without needing to massage the data manually. This becomes increasingly more important as the possible data constellations in the app become more complex.

image

image


tvararu (approved):
Nice!

PR #1120, opened 2024-03-04
Protect flipper UI with basic auth

tvararu:
Also show it in every environment.


misaka (approved):

misaka:
Presumably failing because Playwright tests rely on being able to turn on/off features with the Flipper UI which is now basic authed in test.

PR #1121, opened 2024-03-04
Remove production checks

tvararu:
We're running in production now and it feels unlikely that we'd ship an environment without Sentry setup given that we use the same IAAS templates for everything.

Second commit allows RAILS_ENV=production without a Notify key being set, making it easier to test it locally. This is slightly more controversial as you can argue that it's better if the app refuses to boot up if Notify isn't set, but same logic as the SENTRY_DSN; I think we're protecting from a fairly low risk...


misaka (approved):
LGTM

PR #1125, opened 2024-03-04
Hook up session reminder emails and add job to send them

misaka:
Session reminders are sent out the day before a session to remind parents that their children will be vaccinated the next day.

- [x] Mailer and jobs added.
- [x] cron job added for production env
- [x] `SessionRemindersBatchJob` is idempotent, will only send a patient a reminder email once

This implementation adds One More Column to track sent reminders to patients ... we should consider changing this soon.

Sample of email:

image


app/jobs/session_reminders_job.rb
```diff
@@ -0,0 +1,16 @@
...
+# that needs them sent today.
+
+class SessionRemindersJob < ApplicationJob
+  queue_as :default
+
+  def perform(*_args)
+    return unless Flipper.enabled?(:scheduled_emails)
+
+    Session.active.each do |session|
+      if session.date == Time.zone.today + 1.day
```
benilovj, line 11:
```suggestion
      if session.date == Time.zone.tomorrow
```

PR #1126, opened 2024-03-04
Show the preferred name in consent, triage, record and matching patient lists

benilovj:
## Matching

image

## Consent

image

## Triage

image

## Record

image

## Improvements to ViewComponent previews

To test this change, it was handy to add a preview for `AppPatientTableComponent`. The preview relied on factories and I ran into problems when refreshing the preview because of sequence clashes in the factories.

To deal with this, I introduced a custom preview controller (see 7ad28aeb84bd16969892391efdfe64994f85c9d7) that wraps a transaction around the preview action and silently rolls everything back after it's run, so the development DB isn't polluted and the sequence clashes can be avoided. This seems to be working well so far 🤞 .

### DIFF
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -32,7 +32,7 @@ FactoryBot.define do
     sequence(:full_name) { |n| "Test User #{n}" }
     sequence(:email) { |n| "test-#{n}@localhost" }
     sequence(:teams) { [Team.first || create(:team)] }
-    password { "power overwhelming" }
+    password { "power overwhelming!" } # avoid a password that was found in a data breach
     registration { "SW608658 (HCPC)" }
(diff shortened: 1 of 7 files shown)


spec/factories/users.rb
```diff
@@ -32,7 +32,7 @@
     sequence(:full_name) { |n| "Test User #{n}" }
     sequence(:email) { |n| "test-#{n}@localhost" }
     sequence(:teams) { [Team.first || create(:team)] }
-    password { "power overwhelming" }
+    password { "power overwhelming!" } # avoid a password that was found in a data breach
```
tvararu, line 35: Lol

tvararu (approved):

PR #1136, opened 2024-03-05
Miscellaneous fixes

tvararu:
See commits. Controversial ones can be split off and discussed individually if necessary.


benilovj (approved):
I broadly agree with the changes on this PR.

I still feel quite strongly about wanting some scenarios that:

* test complete end-to-end scenarios
* avoid factories as much as possible (ideally completely) and set things up "through the interface"

There don't have to be **many** such tests (only a handful would give me confidence) but I'm also conscious of a bunch of stuff that's slipped through recently where our tests are all green, but upon closer inspection the factories aren't setting the app up in the same way that the actual user journey does it.

benilovj:
It looks like the test failure is a result of the factory change...

tvararu:
Agree @benilovj, I want to add more tests to fill in the gaps

PR #1137, opened 2024-03-05
Implement consent deadline

tvararu:
Close parental consent by forcing users to go to a separate page once the deadline has passed.

TODO:

- [x] End to end session creation spec that tests this

Screenshot 2024-03-05 at 12 16 22

### DIFF
--- a/config/initializers/email_templates.rb
+++ b/config/initializers/email_templates.rb
@@ -1,4 +1,4 @@
-EMAIL_TEMPLATES = {
+EMAILS = {
   confirmation_the_hpv_vaccination_has_taken_place:
     "8a65d7b5-045c-4f26-8f76-6e593c14cb6d",
   confirming_the_hpv_vaccination_didnt_happen:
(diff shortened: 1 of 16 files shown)


config/initializers/email_templates.rb
```diff
@@ -1,4 +1,6 @@
-EMAIL_TEMPLATES = {
+EMAILS = {
+  consent_request: "6aa04f0d-94c2-4a6b-af97-a7369a12f681",
+  consent_reminder: "ceefd526-d44c-4561-b0d2-c9ef4ccaba4f",
```
misaka, line 3: The reason the other templates here have annoyingly long names is because they copy the names given in Notify, which can be very different from what we think of them. For example the `consent_request` one above is  called "Initial email to parents - HPV" in Notify, which was annoying to find because of the mismatch.

Using the Notify name makes it easier to find when we've forgotten the context or when someone new is looking at this, and I think we should use those. We should also / could instead rename the templates in Notify to be more sensible ... happy to work towards that too.

misaka:
image

🤦

benilovj (approved):

PR #1140, opened 2024-03-05
Display refusal notes from consent forms in Mavis

benilovj:
## Consents on the patient page

image

## When manually matching consent forms to cohort records

image

### DIFF
--- a/spec/factories/patients.rb
+++ b/spec/factories/patients.rb
@@ -108,6 +108,20 @@ FactoryBot.define do
       end
     end
 
+    trait :consent_refused_with_notes do
+      after(:create) do |patient, evaluator|
+        create(
+          :consent,
+          :refused,
+          :from_mum,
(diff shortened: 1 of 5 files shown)


spec/factories/patients.rb
```diff
@@ -108,6 +108,20 @@
...
+          :consent,
+          :refused,
+          :from_mum,
+          campaign: evaluator.campaign,
+          reason_for_refusal: "already_vaccinated",
+          reason_for_refusal_notes: "Already had the vaccine at the GP",
+          patient:
+        )
+      end
+    end
```
misaka, line 123: I think these should be sub-factories as they seem pretty complete versions. I'm also wondering if we need the `after(:create)` part ... I've been playing around and I think these get created correctly with inline association definition.

🥾🥫

misaka (approved):

PR #1141, opened 2024-03-05
Add emailing when vaccination outcome is recorded

misaka:
This uses the recently-added VaccinationMailer to send vaccination emails when a patient has been vaccinated, or when the nurse tried and was unable to vaccinate.

One issue uncovered is that when patients were assessed with Gillick assessment, they're consent information doesn't include a name or email for parents, which is what's used for sending these vaccination emails. For now sending emails for patients with self-consent has been disabled until we know how to deal with the issue.


benilovj (approved):

benilovj:
Merging to get it onto Test.

Gemfile
```diff
@@ -68,6 +68,7 @@ end
 
 group :test do
   gem "capybara"
+  gem "capybara-screenshot"
```
tvararu, line 71: Nice

PR #1145, opened 2024-03-06
Cache assets forever

tvararu:
We already use cache-busting strings, so this should be safe to do.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -44,6 +44,10 @@ module ManageVaccinations
 
     config.middleware.use Rack::Deflater
 
+    config.public_file_server.headers = {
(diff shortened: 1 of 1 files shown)


misaka (approved):

config/application.rb
```diff
@@ -44,6 +44,10 @@ class Application < Rails::Application
 
     config.middleware.use Rack::Deflater
 
+    config.public_file_server.headers = {
+      "Cache-Control" => "public, max-age=2147483648"
```
paulrobertlloyd, line 48: Should we add immutable, too?

```suggestion
      "Cache-Control" => "public, max-age=2147483648, immutable"
```

PR #1155, opened 2024-03-06
Show right reason when unable to vaccinate

benilovj:
If a child couldn't be vaccinated at the session for any reason other than refusal, the patient page was always showing "refused vaccine" rather than the actual recorded reason.

Below, the vaccinator recorded that the child couldn't be vaccinated due to contraindications:

## Before

image

## After

image

## What's been changed

* add a component preview for this case
* fix the bug in the component spec (one was already written but it wasn't actually testing what was on the page)
* fix the bug in the component itself
* remove full stops from the banner messages (to be consistent with the other banners)

### DIFF
--- a/app/components/app_status_banner_component.rb
+++ b/app/components/app_status_banner_component.rb
@@ -176,9 +176,16 @@ class AppStatusBannerComponent < ViewComponent::Base
   end
 
   def reason_do_not_vaccinate_summary
-    I18n.t(
-      "patient_session_statuses.unable_to_vaccinate.banner_explanation.#{state}"
-    )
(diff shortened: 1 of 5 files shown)


misaka:
> fix the bug in the component spec (one was already written but it wasn't actually testing what was on the page)

found this comment interesting. I often feel like we're under-testing in our tests, and my tendency would be to do more testing. It's a fine line to the land of overly-brittle tests, but ... ya, maybe we need to ratchet up the amount we test in component and feature tests just a little.

misaka (approved):

PR #1156, opened 2024-03-06
More banner fixes and improvements

benilovj:
**Depends on #1155, rebase after that's merged.**

* Add previews for most patient banner types
* Remove full-stops from banner messages
* 8205df30c2a4986cc72998d3ee46c3aaead7600a: fix typo in a banner message (it might not actually be used though!)

### DIFF
--- a/config/locales/patient_session_statuses.en.yml
+++ b/config/locales/patient_session_statuses.en.yml
@@ -34,7 +34,6 @@ en:
       colour: red
       text: Could not vaccinate
       banner_title: Could not vaccinate
-      banner_explanation: "%{triage_nurse} that %{full_name} should not be vaccinated"
     triaged_kept_in_triage:
       colour: blue
       text: Triage started
--- a/config/locales/patient_session_statuses.en.yml
+++ b/config/locales/patient_session_statuses.en.yml
@@ -24,17 +24,17 @@ en:
       colour: orange
       text: Check conflicting consent
       banner_title: Conflicting consent
-      banner_explanation: "You can only vaccinate if all respondents give consent."
+      banner_explanation: "You can only vaccinate if all respondents give consent"
     delay_vaccination:
       colour: red
       text: Delay vaccination
       banner_title: Delay vaccination to a later date
-      banner_explanation: "%{triage_nurse} decided that %{full_name}’s vaccination should be delayed."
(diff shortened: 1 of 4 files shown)


config/locales/patient_session_statuses.en.yml
```diff
@@ -34,7 +34,7 @@ en:
       colour: red
       text: Could not vaccinate
       banner_title: Could not vaccinate
-      banner_explanation: "%{triage_nurse} that %{full_name} should not be vaccinated"
+      banner_explanation: "%{triage_nurse} decided that %{full_name} should not be vaccinated"
```
misaka, line 37: > However, I'm not convinced this message is actually shown anywhere. This state is an outcome state and the banner explanation may not be shown at all.

Maybe delete this and set [raise_on_missing_translations](https://api.rubyonrails.org/v7.0.0/classes/ActionView/Helpers/TranslationHelper.html)? Apparently that was [recently changed to cover more cases of the missing translation being used](https://github.com/rails/rails/pull/47105).

misaka (approved):
I feel like I'm going to need to figure out how to use these previews, but ...LGTM!

benilovj:
> I feel like I'm going to need to figure out how to use these previews, but ...LGTM!

There's not a great deal to them, really. They're very reminiscent of mailer previews. I added a few notes about them in the README a few PRs ago:

https://github.com/nhsuk/manage-childrens-vaccinations?tab=readme-ov-file#previewing-view-components

The one caveat is that I've set them up so they run in a transaction, so anything you do in the preview "controller" gets rolled back immediately after rendering and doesn't pollute your development DB.

config/locales/patient_session_statuses.en.yml
benilovj: Oh yeah, that's a really good idea... Is there any reason not to enable it generally in dev and test? https://github.com/nhsuk/manage-childrens-vaccinations/pull/1158

benilovj: So:
* I enabled the `raise_on_missing_translations` locally in dev and test
* removed the string
* re-ran the tests and nothing broke

So I've now yanked that line, instead of fixing the typo. I will do another clean-up of the strings in that translation file in a follow-up PR.

PR #1160, opened 2024-03-07
Rollback healthcheck changes

misaka:
With further testing it was found that these don't have an effect on deploys. Should've manually tested deploys from local first.


tvararu (approved):
😿

PR #1165, opened 2024-03-08
Split up AppStatusBannerComponent

benilovj:
The outcome banner logic goes into `AppOutcomeBannerComponent` (these are the banners with lots of fields).
The simple banner logic goes into `AppSimpleOutcomeBannerComponent` (these are the 1-liner banners).


tvararu (approved):
Nice!

PR #1166, opened 2024-03-09
Replace parental_consent_authentication playwright spec

tvararu:
This tests that parents are unable to view each other's consent forms.


benilovj (approved):
Looks good as a rewrite of the playwright tests.

I did find the authentication spec non-obvious though... are we testing that starting a new form for another school blows away the original one?

tvararu:
> Looks good as a rewrite of the playwright tests.
> 
> I did find the authentication spec non-obvious though... are we testing that starting a new form for another school blows away the original one?

Sort of, we’re testing that attempting to access a different consent form to the one you’re currently working on isn’t possible…

PR #1170, opened 2024-03-11
NIVS export

benilovj:
## Dashboard

screencapture-localhost-4000-dashboard-2024-03-11-20_05_15

## Reports page

screencapture-localhost-4000-reports-2024-03-11-20_05_31

## The CSV report

From this vaccination:

image

The following row is produced:

image

image

(the missing cells in the report are to do with our fake dataset, rather than the functionality of the report)

### DIFF
--- a/app/models/nivs_report.rb
+++ b/app/models/nivs_report.rb
@@ -1,0 +1,36 @@
+require "csv"
+
+class NivsReport
+  HEADERS = %w[
+    ORGANISATION_CODE
+    SCHOOL_URN
+    SCHOOL_NAME
+    NHS_NUMBER
+    PERSON_FORENAME
+    PERSON_SURNAME
+    PERSON_DOB
+    PERSON_GENDER_CODE
+    PERSON_POSTCODE
+    DATE_OF_VACCINATION
+    VACCINE_GIVEN
+    BATCH_NUMBER
+    BATCH_EXPIRY_DATE
+    ANATOMICAL_SITE
+    DOSE_SEQUENCE
+    LOCAL_PATIENT_ID
+    LOCAL_PATIENT_ID_URI
+    CARE_SETTING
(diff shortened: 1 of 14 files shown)


app/models/nivs_report.rb
```diff
@@ -0,0 +1,36 @@
+require "csv"
+
+class NivsReport
+  EXPECTED_HEADERS = %w[
```
misaka, line 4: Why not simply `HEADERS`? `EXPECTED_HEADERS` makes it feel like we are in a test.

spec/models/nivs_report_row_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+           "20210101",
+           "Left Upper Arm",
+           "1",
+           "MAVIS-#{vaccination.patient.id}",
+           "",
+           "1 - School"
+         ]
+    end
+
+    describe "for nasal flu" do
```
misaka, line 62: nit-picky, sorry, but ... I think this is test setup, not what we're testing?

```suggestion
    context "for nasal flu" do
```

spec/models/nivs_report_row_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      it "includes the correct delivery site" do
+        vaccination.batch.update!(
+          vaccine: create(:vaccine, :flu, :fluenz_tetra)
+        )
+        expect(subject[13]).to eq "Nasal"
+      end
+    end
+
+    describe "for lower arm" do
```
misaka, line 71: same nit ...

```suggestion
    context "for lower arm" do
```

misaka (approved):
A couple minor nit-picks, but ... ❤️  ... this is really good to see!

benilovj:
All good points - updated.

PR #1175, opened 2024-03-12
The NIVS report shouldn't include draft vaccs records

benilovj:


app/controllers/reports_controller.rb
```diff
@@ -8,6 +8,7 @@ def show
     vaccinations =
       policy_scope(VaccinationRecord)
         .administered
+        .recorded
```
misaka, line 11: Makes me wonder if these shouldn't be default scopes ... as imperfect as those can be.

misaka (approved):

PR #1176, opened 2024-03-12
Return users to vaccination section when doing triage while vaccinating

misaka:
The implementation here is pretty simple, although I'm still a little on the fence about it. Basically when they visit either the "Triage health questions" or the "Record vaccinations" pages, this is recorder in their session info as their `current_section` and can be `triage` or `vaccinations`. Then, the `create` or `update` actions during triage will examine `current_section` and return them to the appropriate page.

The alternative to this, admittedly quite simple, approach is to use the paths, e.g.:

> [!Note]
> This is an illustration of an alternative implementation, not what was implemented in this PR

```
# List of patients for vaccinations (existing route)
/sessions/1/vaccinations

# Viewing a patient for vaccination
/sessions/1/vaccinations/patients/2

# Then the URL used for submitting triage for this patient
/sessions/1/vaccinations/patients/2/triage

# For illustration, viewing a patient while performing triage:
/sessions/1/triage/patients/2

# And the path to POST to update triage
/sessions/1/triage/patients/2/triage
```

Reasons to do this:
- transparent and clearer what's going on (imo) vs using cookies 
- stateless approach, can be linkable

Reasons not to do this:
- URLs are a bit long
- technically the patient pages don't change (currently) between these views

Work that needs to be done in follow-up PRs:
- sort out how this same feature is implemented for other actions and back buttons
- either use or remove `:route` from the URL paths (part of above)
- don't send out a "give feedback" email after doing triage during vaccinations (bug, see tests in this PR)
- stop using `:patient_session_id` for the patient page, use `:patient_id` to make it follow convention

### DIFF
--- a/app/controllers/triage_controller.rb
+++ b/app/controllers/triage_controller.rb
@@ -104,7 +104,7 @@ class TriageController < ApplicationController
     if session[:current_section] == "vaccinations"
       vaccinations_session_path(@session)
     elsif session[:current_section] == "consents"
-      consents_session_path(@session)
+      consents_session_path(@session, anchor: "given")
     else # if current_section is triage or anything else
       triage_session_path(@session)
     end
--- a/app/controllers/triage_controller.rb
+++ b/app/controllers/triage_controller.rb
@@ -38,6 +38,8 @@ class TriageController < ApplicationController
 
     # ensure all tabs are present
     tabs_to_states.each_key { |tab| @tabs[tab] ||= [] }
+
+    session[:current_section] = "triage"
   end
 
   def create
@@ -50,7 +52,7 @@ class TriageController < ApplicationController
         patient: @patient,
         view_record_link: session_patient_triage_path(@session, @patient)
       )
-      redirect_to triage_session_path(@session)
+      redirect_to redirect_path
     else
       render "patient_sessions/show", status: :unprocessable_entity
     end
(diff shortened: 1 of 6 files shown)


app/controllers/triage_controller.rb
```diff
@@ -97,4 +99,14 @@ def set_patient_session
   def triage_params
     params.require(:triage).permit(:status, :notes)
   end
+
+  def redirect_path
+    if session[:current_section] == "vaccinations"
+      vaccinations_session_path(@session)
+    elsif session[:current_section] == "consents"
```
benilovj, line 106: I can't see where `session[:current_section] = "consents"` is set?

spec/features/hpv_vaccination_triage_during_vaccination_spec.rb
```diff
@@ -0,0 +1,61 @@
+require "rails_helper"
+
+RSpec.describe "HPV Vaccination" do
```
benilovj, line 3: Not sure whether this matches the new scope of the test. 

Also the spec name has vaccination in it twice; maybe `spec/features/hpv_nurse_triage_during_vaccination_spec.rb`

app/controllers/triage_controller.rb
misaka: Ya, I meant to take this out. Will remove.

spec/features/hpv_vaccination_triage_during_vaccination_spec.rb
misaka: Ok, taken another stab at naming ...

app/controllers/triage_controller.rb
misaka: I added setting the current_section in consents now.

benilovj (approved):

PR #1177, opened 2024-03-12
Extend the 'parental consent refused' feature spec

benilovj:
The spec wasn't actually testing whether:
* the refusal could even be submitted
* the right emails were sent
* the patient is left in the right state in Mavis

### DIFF
--- a/spec/features/parental_consent_refused_spec.rb
+++ b/spec/features/parental_consent_refused_spec.rb
@@ -15,6 +15,11 @@ RSpec.describe "Parental consent" do
 
     when_i_refuse_consent
     then_i_can_check_my_answers
+
+    when_i_confirm_my_answers
+    then_i_see_the_confirmation_page
+    and_i_receive_an_email_confirming_that_my_child_wont_be_vaccinated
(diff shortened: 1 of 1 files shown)


spec/features/parental_consent_refused_spec.rb
```diff
@@ -90,4 +99,57 @@ def then_i_see_the_consent_page
...
+    )
+  end
+
+  def and_i_receive_an_email_confirming_that_my_child_wont_be_vaccinated
+    expect(enqueued_jobs.first["scheduled_at"]).to be_nil
+    expect(
+      Time.zone.parse(enqueued_jobs.second["scheduled_at"]).to_i
+    ).to be_within(1.second).of(1.hour.from_now.to_i)
+
+    expect_email_to "jane@example.com",
```
misaka, line 119: After finding that too many emails were being sent out as part of the triage needed vaccination journey, I'd be tempted to add the expectation on the number of emails being sent. We don't test that part of the journeys anywhere else.

misaka (approved):
Maybe just add the email count expectation, but either way a big improvement.

PR #1184, opened 2024-03-13
Patient records should appear in 'triage' and 'record vaccination' if their vaccination has been delayed

benilovj:
## Triage page after vaccination has been delayed

image

## Record vaccinations

image

## Patient page

image

## Notes for reviewers

* this PR extends the triage_delay_vaccination spec and ports it to RSpec
* there's some duplication between `app/helpers/patient_session_helper.rb` and `config/locales/patient_session_statuses.en.yml` that I would like to reduce in a subsequent PR


misaka (approved):
🚀

PR #1187, opened 2024-03-13
Re-implement the skipped parental_consent_change_answers test in RSpec

benilovj:
The spec was temporarily disabled in order to ship #1137.

Not sure whether we'd want to keep using `/random_consent_form?` but I've left it in for now. I've also split the original spec into 3 scenarios as there are clearly 3 separate parts to it.


misaka (approved):
I think it's worth keeping these scenarios together like this ... even if it's just to try it out for size. 👍

PR #1188, opened 2024-03-13
Update AppPaitentDetailsComponent to use #call instead of ERB

benilovj:
The template wasn't doing any actual rendering so it was just not-very-useful indirection.


misaka (approved):
:shipit:

PR #1190, opened 2024-03-15
Verbal consent attempt, no response: update designs

benilovj:
Copy changes after discussion with designers. The patient page will be aligned in a follow-up PR.

Before             |  After
:-------------------------:|:-------------------------:
image |  image

PR #1191, opened 2024-03-15
Verbal consent attempt, no response: update designs (part 2)

benilovj:
Follow-up to #1190. Copy changes, after discussion with designers, to the consent list on the patient page.

Before             |  After
:-------------:|:----------------:
image|image

### DIFF
--- a/app/components/app_consent_component.rb
+++ b/app/components/app_consent_component.rb
@@ -55,12 +55,20 @@ class AppConsentComponent < ViewComponent::Base
 
   def consents_grouped_by_parent
     @consents_grouped_by_parent ||=
-      @patient_session.consents.group_by do |consent|
-        response = consent.human_enum_name(:response).capitalize
-        summary = "#{response} by #{consent.name}"
-        summary +=
-          " (#{consent.who_responded})" unless consent.via_self_consent?
-        summary
+      @patient_session.consents.group_by { |consent| summary_for(consent) }
+  end
+
+  private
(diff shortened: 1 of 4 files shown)


app/components/app_consent_component.rb
```diff
@@ -55,12 +55,20 @@ def contact_parent_or_guardian_link(consents)
...
-          " (#{consent.who_responded})" unless consent.via_self_consent?
-        summary
+      @patient_session.consents.group_by { |consent| summary_for(consent) }
+  end
+
+  private
+
+  def summary_for(consent)
+    summary =
+      if consent.respond_to?(:response_not_provided?) &&
```
misaka, line 65: Why do we need to check it responds to `response_not_provided?`?

app/components/app_consent_response_component.rb
```diff
@@ -18,7 +18,12 @@ def initialize(consent:)
     end
 
     def response
-      consent.human_enum_name(:response).capitalize
+      if consent.respond_to?(:response_not_provided?) &&
```
misaka, line 21: Ditto ...

misaka (approved):

app/components/app_consent_component.rb
benilovj: because `consent` here can be either a `Consent` or a `ConsentForm`. Perhaps we shouldn't be using the same component for displaying both.

PR #1194, opened 2024-03-17
On the outcome banner, display a location only when the vaccination record has been recorded

benilovj:
This prevents draft vaccs records from causing the school to be displayed.

### DIFF
--- a/app/components/app_outcome_banner_component.rb
+++ b/app/components/app_outcome_banner_component.rb
@@ -48,8 +48,8 @@ class AppOutcomeBannerComponent < ViewComponent::Base
   end
 
   def show_location?
-    # location only makes sense if an attempt to vaccinate on site was made
-    @patient_session.vaccination_records.any?
+    # showing the location only makes sense if an attempt to vaccinate on site was made
+    @patient_session.vaccination_records.recorded.any?
   end
 
   def vaccine_summary


app/components/app_outcome_banner_component.rb
```diff
@@ -48,8 +48,8 @@ def rows
   end
 
   def show_location?
-    # location only makes sense if an attempt to vaccinate on site was made
-    @patient_session.vaccination_records.any?
+    # showing the location only makes sense if an attempt to vaccinate on site was made
+    @patient_session.vaccination_records.recorded.any?
```
misaka, line 52: A part of me wonders if we shouldn't continue with adding default scopes ... maybe something to think about.

misaka (approved):

app/components/app_outcome_banner_component.rb
benilovj: I think you might be right... I'll try a default scope before merging this PR.

benilovj: https://github.com/nhsuk/manage-childrens-vaccinations/pull/1194

benilovj: Oops I meant https://github.com/nhsuk/manage-childrens-vaccinations/pull/1197

benilovj:
We've merged #1197 so closing this one.

PR #1206, opened 2024-03-20
Simplify consent summary code

benilovj:
This PR replaces `AppConsentDetailsComponent` and `AppConsentResponseComponent` with a much dumber (but simpler) `AppConsentSummaryComponent`.

This does come at a cost of making `AppConsentComponent` more complex. I will look at simplifying this one in a follow-up PR. There's also likely value in replacing `AppConsentCardComponent` with `AppConsentSummaryComponent` to further pare things down.

This PR is best reviewed commit-by-commit.


app/components/app_consent_component.rb
```diff
@@ -55,20 +55,36 @@ def contact_parent_or_guardian_link(consents)
...
-           consent.response_not_provided?
-        "Contacted #{consent.name}"
-      else
-        "#{consent.human_enum_name(:response).capitalize} by #{consent.name}"
-      end
-    summary += " (#{consent.who_responded})" unless consent.via_self_consent?
-    summary
+  def render_grouped_consents(consents)
+    first_consent = consents.first
+    first_refused_consent = consents.find_all(&:response_refused?).first
```
misaka, line 65: I think a plain old `find` would work here too.

```suggestion
    first_refused_consent = consents.find(&:response_refused?)
```

app/components/app_consent_component.rb
```diff
@@ -55,20 +55,36 @@ def contact_parent_or_guardian_link(consents)
...
+          { text: consent.summary_with_route, timestamp: consent.recorded_at }
+        end
+    }
+    unless first_consent.via_self_consent?
+      props[:relationship] = first_consent.who_responded
+      props[:contact] = {
+        phone: first_consent.parent_phone,
+        email: first_consent.parent_email
+      }
+    end
```
misaka, line 80: I found this hard to read on the first pass. I think this scans slightly better, what do you think?

```suggestion
    props.merge(
      relationship: first_consent.who_responded
      contact: {
        phone: first_consent.parent_phone,
        email: first_consent.parent_email
      }) unless first_consent.via_self_consent?
```

app/components/app_consent_component.html.erb
```diff
@@ -9,7 +9,7 @@
 <% if patient_session.consents.present? %>
   <% consents_grouped_by_parent.each do |summary, consents| %>
     <%= render AppDetailsComponent.new(summary:, open: open_consents?) do %>
-      <%= render AppConsentDetailsComponent.new(consents:) %>
+      <%= render_grouped_consents(consents) %>
```
misaka, line 12: Just had another thought about making this a bit more readable:

```suggestion
      <%= render AppConsentSummaryComponent.new(**grouped_consents) %>
```

app/components/app_consent_component.rb
```diff
@@ -55,20 +55,36 @@ def contact_parent_or_guardian_link(consents)
...
-      end
-    summary += " (#{consent.who_responded})" unless consent.via_self_consent?
-    summary
+  def render_grouped_consents(consents)
+    first_consent = consents.first
+    first_refused_consent = consents.find_all(&:response_refused?).first
+
+    props = {
+      name: first_consent.name,
+      response:
```
misaka, line 69: This is a list of responses, so plural?

```suggestion
      responses:
```

misaka: nvm, I see below that it can be a single `response` in some circumstances too. 🤔

misaka:
Quite a lot here. How about adding a component preview?

benilovj:
> Quite a lot here. How about adding a component preview?

Agreed – added in [bafba7d](https://github.com/nhsuk/manage-childrens-vaccinations/pull/1206/commits/bafba7d051c5f349ae7ad673f4394d47b8e446e2)

app/components/app_consent_component.rb
benilovj: I've left it as `response` for now, as that way it can be mandatory. I suppose there's an option to allow both `response` and `responses` but that would require checking that at least one is populated and picking the right one within the component 🤷 . I can do that if needed...

app/components/app_consent_component.html.erb
benilovj: Yep, good idea, added in [b7a8b34](https://github.com/nhsuk/manage-childrens-vaccinations/pull/1206/commits/b7a8b348e2c1ecc34071fea5aa42f8b38b19eb67)

app/components/app_consent_component.rb
benilovj: I tried a variation of this in [b7a8b34](https://github.com/nhsuk/manage-childrens-vaccinations/pull/1206/commits/b7a8b348e2c1ecc34071fea5aa42f8b38b19eb67) – the linter didn't like sticking the `unless` at the end and immediately changed it back it but I've adopted the `merge`.

app/components/app_consent_component.rb
benilovj: Yep, updated in [b7a8b34](https://github.com/nhsuk/manage-childrens-vaccinations/pull/1206/commits/b7a8b348e2c1ecc34071fea5aa42f8b38b19eb67)

app/components/app_consent_component.rb
misaka: Ya thought that might happen. I'll see if it still reads better ... hopefully I've forgotten how it works sufficiently. 😆

app/views/consent_forms/review_match.html.erb
```diff
@@ -21,7 +21,22 @@
...
+          email: @consent_form.parent_email
+        },
+        response: {
+          text: @consent_form.summary_with_route,
+          timestamp: @consent_form.recorded_at
+        },
+        refusal_reason: {
+          reason: @consent_form.reason,
+          notes: @consent_form.reason_notes
+        }
```
misaka, line 38: If consent is given then this `refusal_reason` section gets included, and I think the rendering isn't quite right since it'll show an empty "Refusal reason" entry on the "Link consent response ..." page. Tested locally:

image

benilovj: Ah yes, I came across this in the subsequent PR and fixed it in 928d015c52755018b5c3ed83cf3ad34f4542d2a8. I've cherry-picked that commit onto this branch: [5d6004d](https://github.com/nhsuk/manage-childrens-vaccinations/pull/1206/commits/5d6004dc45d2e287c31f57d02d15677dfe5caea7)

misaka (approved):
🚢

PR #1207, opened 2024-03-21
Simplify consent summary code: part 2

benilovj:
> ~Rebase after #1206 is merged~ ✅ 

This PR continues to pare down the view components used in rendering consents and consent forms, by using `AppConsentSummaryComponent` in more views.


app/views/consent_forms/review_match.html.erb
```diff
@@ -33,7 +33,7 @@
           timestamp: @consent_form.recorded_at
         },
         refusal_reason: {
-          reason: @consent_form.reason,
+          reason: @consent_form.human_enum_name(:reason).presence,
```
misaka, line 36: This surprised me a little. I wonder if we should change `human_enum_name` to return `nil` instead of an empty string when the enum is `nil`. 👟🥫

misaka (approved):
Nice one. I see some tie-ins with the work I've been knocking my head against, hopefully that should be done soon.

:shipit:

PR #1208, opened 2024-03-22
Allow consent to be changed to given via verbal consent

misaka:
This fixes an error triggereg when a nurse tries to record consent given verbally when it was previously refused by the parent.

We fix this by cloning the existing consent instead of editing the existing one. This means when determining the state of consent we need to only look at the most recent response from a parent. It's also exposed other missing state transitions, such as if the parent verbally refuses consent again in the above scenario, or if another consent form is filled in. Fixes to these other transitions will be done in a follow-up PR to keep this one small(er).

Re: the cloning implementation, there are a couple things to point out. Because the link to changing the consent response is a link and not a button, we can't clone consent response when they click that link. We instead clone the consent object when the nurse to submits the response on the first page. I did try adding cloning functionality to `update` or `create` so that the form didn't have to change, but that started to get weird from a REST POV so I decided to just add the `clone` action to keep it clean.

### DIFF
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -64,4 +64,11 @@ class PatientSession < ApplicationRecord
     !unable_to_vaccinate? && !unable_to_vaccinate_not_assessed? &&
       !unable_to_vaccinate_not_gillick_competent?
   end
+
+  def latest_consents
+    consents
+      .recorded
+      .group_by(&:name)
+      .map { |_, consents| consents.max_by(&:recorded_at) }
+  end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -68,6 +68,9 @@ end
 
 group :test do
   gem "capybara"
+  gem "capybara_accessible_selectors",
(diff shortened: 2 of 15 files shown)


misaka:
With the benefit of a weekend's worth of hindsight, I think the feature specs in here should probably be shortened to not exercising the consent form bit, and just start from the nurse's journey. I'll update the PR.

app/models/patient_session.rb
```diff
@@ -64,4 +64,11 @@ def able_to_vaccinate?
     !unable_to_vaccinate? && !unable_to_vaccinate_not_assessed? &&
       !unable_to_vaccinate_not_gillick_competent?
   end
+
+  def latest_consents_for_each_parent
```
tvararu, line 68: I'm not sure the `_for_each_parent` is adding much:

```suggestion
  def latest_consents
```

tvararu (approved):
🚀

app/models/patient_session.rb
misaka: Ya, I went back and forth on this. Will shorten

PR #1209, opened 2024-03-25
Add state concern to replace state machine

misaka:
Now that we understand how states work better, and put all the states we need into the various objects connected to PatientSessions, it seems clear that the state machine isn't actually adding anything beyond additional hurdles and a false sense of comfort.

The new approach in this concern is something we discussed in the beginning; we "calculate" the PatientSession state by looking at the state of the various connected objects: consent, triage and vaccination_record. This was all being done in the state machine as it stands, anyway, the only additional complexity is the order in which we look at these objects. What's removed, really, are the transitions, and all the code to define, manage and test them.

Tasks remaining:
- [x] Some basic benchmarks, maybe even just benchmarking existing tests. The new approach requires a bigger database hit for each request cycle that looks at `PatientSession`, but this is unlikely to be an issue.
- [ ] Commit to this approach; remove state machine code for good.
- [ ] Remove `do_*` transition stub methods and calls to them
- [ ] Aspirational: See what tidying up we can do in `PatientSessionStateConcern` now that it doesn't conform to AASM.
- [ ] Aspirational: See if there's a way to tidy up `state` method. If we grouped together the vaccination_record, triage and consent checks, would that be better?

### DIFF
--- a/app/controllers/consents_controller.rb
+++ b/app/controllers/consents_controller.rb
@@ -36,7 +36,7 @@ class ConsentsController < ApplicationController
       @session
         .patient_sessions
         .strict_loading
-        .includes(:campaign, :consents, :patient)
+        .includes(:campaign, :consents, :patient, :triage)
         .sort_by { |ps| ps.patient.full_name }
   end
 end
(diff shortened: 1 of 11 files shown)


tvararu (approved):
🚀

PR #1235, opened 2024-04-10
Make triage tabs do page reloads

misaka:
Make the triage tabs do full page reloads instead of using JS to hide/show tabs.

Currently the consent, triage and vaccination sections load up all the tabs and use JS to switch between them. The current implementation has a couple downsides:

- back-buttons are harder to implement
- the page is a mess on mobile and non-JS
- and to a lesser extent, there will be performance risks as we have sessions with more users

This PR is the first step to changing them all to page reloads. Current issues with this implementation:

- nhsuk-frontend currently breaks because it depends on discovering the tab label by examining the link for the tab and using the fragment from it (see https://github.com/nhsuk/nhsuk-frontend/blob/main/packages/components/tabs/tabs.js#L306). This means that the `role` and other aria-related attributes are broken, although the behaviour seems to work
- scroll position is lost since, predictably, the page reload loses the position. Maybe this could be fixed if we implemented content-replacement?

Work remaining after this PR:
- [ ] convert consents and vaccinations sections
- [ ] move the patient pages into the new URL paths, e.g. `/sessions/1/triage/needed/patients/2` and change the back buttons to rely on the path instead of cookies
- [ ] re-implement the tabs themselves, these are being redesigned and we won't be able to use the govuk tabs components (see above issues)

### DIFF
--- a/app/components/app_tab_component.rb
+++ b/app/components/app_tab_component.rb
@@ -21,10 +21,20 @@ class AppTabComponent < GovukComponent::Base
   class Tab < GovukComponent::Base
     attr_reader :label, :text
 
-    def initialize(label:, text: nil, id: nil, classes: [], html_attributes: {})
+    def initialize(
+      label:,
+      selected: nil,
+      link: nil,
+      text: nil,
+      id: nil,
+      classes: [],
+      html_attributes: {}
+    )
       @label = label
       @text = text
       @id = id || label.parameterize
+      @selected = selected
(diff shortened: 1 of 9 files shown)


tvararu (approved):
Non-blocking suggestion for an improvement to the HTML; I think we should do it but can be done in a follow-up too!

misaka:
Agreed with the two comments. I was going to just get the three sections here converted, and then finesse the component some more after.

PR #1239, opened 2024-04-15
Reset more fields on ConsentForm when switching paths

tvararu:
This consolidates some of the existing before_save hooks into a single method that handles all the different cases.

### DIFF
--- a/app/models/consent_form.rb
+++ b/app/models/consent_form.rb
@@ -48,8 +48,6 @@ class ConsentForm < ApplicationRecord
   include AgeConcern
 
   before_save :reset_unused_fields
-  before_save :seed_health_questions_if_consent_given
-  before_save :remove_health_questions_if_consent_refused
 
   scope :unmatched, -> { where(consent_id: nil) }
   scope :recorded, -> { where.not(recorded_at: nil) }
@@ -196,7 +194,7 @@ class ConsentForm < ApplicationRecord
   end
 
   def address_postcode=(str)
-    super UKPostcode.parse(str.to_s).to_s
+    super str.nil? ? nil : UKPostcode.parse(str.to_s).to_s
   end
 
   def full_name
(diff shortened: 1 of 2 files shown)


app/models/consent_form.rb
```diff
@@ -196,7 +194,7 @@ class ConsentForm < ApplicationRecord
   end
 
   def address_postcode=(str)
-    super UKPostcode.parse(str.to_s).to_s
+    super str.nil? ? nil : UKPostcode.parse(str.to_s).to_s
```
misaka, line 197: Would this have the same effect? Would `str` ever be false?

```suggestion
    super str || UKPostcode.parse(str.to_s).to_s
```

app/models/consent_form.rb
```diff
@@ -342,26 +340,41 @@ def ask_for_contact_method?
...
   def reset_unused_fields
     self.common_name = nil unless use_common_name?
 
     self.contact_method = nil unless ask_for_contact_method?
     self.contact_method_other = nil unless contact_method_other?
 
-    self.reason = nil unless consent_refused?
-    self.reason_notes = nil unless consent_refused?
+    if consent_refused?
+      self.gp_response = nil
```
misaka, line 353: I thought Ruby must have a `.clear` for this ... and looks like it does. Completely optional, though, it's just a teensy bit of sugar.

```suggestion
      self.gp_response.clear
```

app/models/consent_form.rb
```diff
@@ -342,26 +340,41 @@ def ask_for_contact_method?
...
+
+      self.address_line_1 = nil
```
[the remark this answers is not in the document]
misaka: Unfortunately `Integer`s don't have this, so not sure the sugar would really be worth it, actually.

misaka (approved):
Looks good. My comments are just around sweetening it up a tiny bit, no blockers. Not sure if the `.clear` idea is worth it, but left the comment in case you hadn't considered it already.

app/models/consent_form.rb
tvararu: Hmm doesn't this do the opposite of what we need? E.g. it wouldn't parse and reformat `str` if it's truthy.

PR #1240, opened 2024-04-15
Strip parent_phone and parent_email

tvararu:
Allows more lenient user input for these fields, ensuring whitespace doesn't trigger validation errors.

Also makes sure phone numbers are always saved with consistent whitespace (none).


benilovj (approved):
LGTM

is it worth adding a spec for emails with a capital letter, as documentation?

PR #1241, opened 2024-04-15
Fix back link on manage consents journey

tvararu:
When nurses amend an existing consent, the link sends them to the middle of a consent wizard. `previous_wizard_path` links to the first step in the journey, which causes the back button to go into a loop, because the first step is the :agree one which the user is currently on.

This adds a special `back_link_path` helper that checks for this case and works around it by instead sending a user to the session path for their route. The code to do this is extracted from the :who step. This way, every step is now a possible "first step" depending on how the logic to handle the list of `form_steps` changes.

TODO:

- [x] Tests

### DIFF
--- a/app/controllers/manage_consents_controller.rb
+++ b/app/controllers/manage_consents_controller.rb
@@ -17,6 +17,9 @@ class ManageConsentsController < ApplicationController
   before_action :set_triage,
                 except: %i[create],
                 if: -> { step.in?(%w[questions confirm]) }
+  before_action :set_back_link,
+                only: %i[show],
+                if: -> { wizard_value(step).present? }
 
   def create
     @consent = Consent.create! create_params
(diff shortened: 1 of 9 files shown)


app/controllers/manage_consents_controller.rb
```diff
@@ -28,6 +28,7 @@ def create
   end
 
   def show
+    current_step # Set the current_step for the back link
```
tvararu, line 31: This is a bit awkward, must be a nicer way to generate `@current_step` in the case of the `show` action...

app/helpers/manage_consents_helper.rb
```diff
@@ -10,4 +10,21 @@ def form_method_for(consent)
   def include_clone_fields_for(consent)
     consent.recorded?
   end
+
+  # rubocop:disable Rails/HelperInstanceVariable
```
tvararu, line 14: I didn't feel like passing in 5 different arguments to `back_link_path`.

Rubocop's docs say that [Rails/HelperInstanceVariable](https://www.rubydoc.info/gems/rubocop-rails/RuboCop/Cop/Rails/HelperInstanceVariable) is useful to make helpers more reusable, but this helper is veeery specific to this controller. So I decided to overrule it.

misaka (approved):

PR #1257, opened 2024-04-22
Handle case where patient is absent from school

tvararu:
This fell through and was not being handled. Patients absent were being treated as "Unable to vaccinate" whereas it should be "Delay vaccination."

Related to https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1157.

### DIFF
--- a/app/components/app_simple_status_banner_component.rb
+++ b/app/components/app_simple_status_banner_component.rb
@@ -15,6 +15,7 @@ class AppSimpleStatusBannerComponent < ViewComponent::Base
           default: "",
           full_name:,
           triage_nurse:,
+          vaccination_nurse:,
           who_responded:,
           who_refused:
         )
@@ -31,6 +32,11 @@ class AppSimpleStatusBannerComponent < ViewComponent::Base
 
   def most_recent_triage
     @most_recent_triage ||= @patient_session.triage.order(:created_at).last
+  end
+
+  def most_recent_vaccination
+    @most_recent_vaccination ||=
+      @patient_session.vaccination_records.order(:created_at).last
   end
 
   def who_responded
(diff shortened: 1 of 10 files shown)


config/locales/patient_session_statuses.en.yml
```diff
@@ -29,7 +29,7 @@ en:
       colour: red
       text: Delay vaccination to a later date
       banner_title: Delay vaccination to a later date
-      banner_explanation: "%{triage_nurse} decided that %{full_name}’s vaccination should be delayed"
+      banner_explanation: "%{vaccination_nurse} decided that %{full_name}’s vaccination should be delayed"
```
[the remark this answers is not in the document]
tvararu: Hmm 👍 will do

app/components/app_simple_status_banner_component.rb
```diff
@@ -50,8 +55,16 @@ def full_name
...
-  def triage_nurse
-    most_recent_triage&.user&.full_name
+  def nurse
+    most_recent_event =
+      if most_recent_triage&.created_at.to_i >
+           most_recent_vaccination&.created_at.to_i
+        most_recent_triage
+      else
+        most_recent_vaccination
+      end
```
misaka, line 65: Golf is popular up in Scotland, isn't it?

```suggestion
      [most_recent_triage, most_recent_vaccination].compact.max_by(
        &:created_at
      )
```

Assuming you can compare the `created_at` directly without the `to_i`, which I would've thought you could?

misaka (approved):

app/components/app_simple_status_banner_component.rb
tvararu: Thanks, I wanted to golf something like this but my brain wasn't working at the time!

PR #1265, opened 2024-04-23
Align application to look more like new prototype

tvararu:
These are some quick fixes that aren't worth carding up.

See commits.


misaka (approved):
Had a look on the review app deploy, looks good. 👍

PR #1267, opened 2024-04-24
Align sign in page to prototype

tvararu:
Also tweaks the forgot password and account unlock pages to make them consistent.

Disables pundit policy checking on these controllers as it isn't relevant.

Width of input fields isn't set to 20 as that's more involved for the password field and can be done separately.

### After

image

image

image


misaka (approved):
🔑

PR #1268, opened 2024-04-24
Use AppPatientTableComponent to render the vaccinations tables

misaka:
No functional changes, really, aside from sorting of patients which the component does by default.

## Before

image

## After

image


misaka:
Also, there is a slight spacing difference there. I kind of like the new spacing better, but I figure this will change anyway with the new designs ... 🥾 🥫

tvararu (approved):

PR #1269, opened 2024-04-25
Use one_half layout on sign in pages

tvararu:
This aligns it closer to the prototype and is better than writing custom CSS code.

### DIFF
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -1,7 +1,7 @@
 module Users
   class PasswordsController < Devise::PasswordsController
-    layout "two_thirds"
-
     skip_after_action :verify_policy_scoped
+
+    layout "one_half"
   end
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/_globals.scss
```diff
@@ -4,3 +4,8 @@ li code {
   padding: 2px 8px;
   word-break: break-word;
 }
+
+// Use NHSUK colour for <mark>
+mark {
+  background-color: $color_nhsuk-pale-yellow;
+}
```
paulrobertlloyd, line 11: Has this slipped in from somewhere? I think it was moved to `_overrides.scss`

tvararu: I moved it to globals because it's closer to a global, but removed the change now.

paulrobertlloyd (approved):

PR #1270, opened 2024-04-25
Set up training with custom domains

tvararu:
Add the new certificates and specify them in the manifest.

Also delete the old `staging` config.


docs/aws-copilot.md
```diff
@@ -90,7 +90,7 @@ Certificates.
 Once the cert is approved, feed the ARN using the CLI:
 
 ```bash
-copilot env init --import cert arn:aws.....
+copilot env init --import-cert-arns arn:aws.....
```
misaka, line 93: Is it worth mentioning that this can be done after the env is setup too?

misaka (approved):

PR #1280, opened 2024-05-03
Refactor consents routes, nest patients page

misaka:
This rejigs the routes so that we can nest the patients page under the
correct consents tab:

/sessions/:session_id/consents/:tab/patients/:patient_id

And also as a side-effect changes the generated helpers for the routes,
to something more sensible imo:

consents_session_path -> session_consents_path


app/components/app_consent_component.rb
```diff
@@ -43,11 +44,10 @@ def contact_parent_or_guardian_link(consents)
     link_to(
       "Contact #{consent.parent_name} (the #{role} who refused)",
       session_patient_manage_consent_path(
-        session,
-        patient,
-        @route,
-        consent.id,
-        :agree
+        session_id: session,
```
tvararu, line 47: Hmm, does this work as-is? Maybe for consistency:

```suggestion
        session_id: session.id,
```

Or if the var was an id all along:

```suggestion
        session_id:,
```

app/controllers/manage_consents_controller.rb
```diff
@@ -83,16 +83,16 @@ def current_step
...
-    case @route
+    case @section
     when "consents"
-      consents_session_path(@session)
+      session_consents_path(@session)
     when "triage"
       triage_session_path(@session)
     else
-      vaccinations_session_path(@session)
+      session_vaccinations_path(@session)
```
tvararu, line 95: - session_consents_path
- triage_session_path
- session_vaccinations_path

Triage is the odd one out, how come? I thought it should be `session_triage_path` according to the new routes

app/controllers/patient_sessions_controller.rb
```diff
@@ -47,11 +47,11 @@ def set_back_link
     @back_link =
       case @route
       when "vaccinations"
-        vaccinations_session_path(@session)
+        session_vaccinations_path(@session)
       when "triage"
         triage_session_path(@session)
       when "consents"
-        consents_session_path(@session)
+        session_consents_path(@session)
```
tvararu, line 54: See above

app/controllers/vaccinations_controller.rb
```diff
@@ -90,34 +91,37 @@ def create
         if @draft_vaccination_record.delivery_site_other
           redirect_to edit_session_patient_vaccinations_delivery_site_path(
                         @session,
-                        @patient
+                        patient_id: @patient
```
tvararu, line 94:
```suggestion
                        patient_id: @patient.id
```

app/controllers/vaccinations_controller.rb
```diff
@@ -90,34 +91,37 @@ def create
...
                       )
         elsif @draft_vaccination_record.batch_id.present?
           redirect_to confirm_session_patient_vaccinations_path(
                         @session,
-                        @patient
+                        patient_id: @patient
```
tvararu, line 99:
```suggestion
                        patient_id: @patient.id
```

app/controllers/vaccinations_controller.rb
```diff
@@ -90,34 +91,37 @@ def create
...
                         @session,
-                        @patient
+                        patient_id: @patient.id
                       )
         end
       else
         redirect_to edit_reason_session_patient_vaccinations_path(
                       @session,
-                      @patient
+                      patient_id: @patient
```
tvararu, line 110:
```suggestion
                      patient_id: @patient.id
```

app/helpers/application_helper.rb
```diff
@@ -30,15 +30,20 @@ def page_title(service_name)
...
+  # end
+
+  # def session_patient_vaccinations_path(*args, **kwargs)
+  #   session_patient_path(
+  #     *args,
+  #     **kwargs,
+  #     section: "vaccinations",
+  #     tab: params[:tab]
+  #   )
+  # end
```
tvararu, line 48:
```suggestion
```

app/views/sessions/show.html.erb
```diff
@@ -32,16 +34,22 @@
         </p>
 
         <h2 class="nhsuk-heading-s nhsuk-u-margin-bottom-1">
-          <%= link_to "Triage health questions",
-            triage_session_path(@session) %>
+          <%= link_to(
+            "Triage health questions",
+            # TODO: Can we get rid of the section param here?
```
tvararu, line 39: Worth checking this in like a timeboxed ~5 minutes but I'd remove these TODOs before merging even if we don't. It's really not that bad having to specify it just in this instance.

config/initializers/tab_paths.rb
```diff
@@ -19,3 +19,5 @@
     "could-not" => :could_not_vaccinate
   }
 }.with_indifferent_access
+
+# TAB_SECTIONS = TAB_PATHS.keys
```
tvararu, line 23:
```suggestion
```

tvararu (approved):
New routes and URLs are super clean 👍 

New path helpers also clearer

Back link works predictably based on a bit of testing

Left some comments but pre-approving pending commit tidy up, well done! 🙌

app/components/app_consent_component.rb
misaka: In my experimentation, `session_id` can be set to either `session` or `session.id`, so the original code above should work, ya. However, like you suggest, I've been trying to standardise on `session.id`, I probably just missed that one. It's a shame that `session: session` doesn't work as that could've been nicely abbreviated as `session:`

misaka:
The comments are spot-on, and things that just haven't been changed because this is still a WIP. I'll update it, probably addressing most if not all the comments here, and then maybe you can have another look @tvararu . Cheers!

PR #1287, opened 2024-05-10
Update session show page

tvararu:
Match it up to the new design in the prototype.

### Screenshots

image


app/models/session_stats.rb
```diff
@@ -39,6 +42,9 @@ def calculate_stats
         s.consent_given_triage_not_needed?
     end
 
+    counts[:not_ready_to_vaccinate] = @patient_sessions.size -
+      counts[:ready_to_vaccinate]
```
misaka, line 46: 🤢

misaka (approved):

PR #1291, opened 2024-05-13
Remove expressions of interest from the app

benilovj:
Expressions of interest were only used for the pilot and are unnecessary going forward.

This PR untangles registrations from everything else and deletes the associated controllers, routes, models, etc. The commits are in a logical progression (and have their tests passing), and can be reviewed sequentially.

This PR leaves the pilot area intact but with just one item – uploading a cohort spreadsheet. This can be expanded later into more fully-featured cohorting.


misaka (approved):
🚢

PR #1298, opened 2024-05-14
Add AppSecondaryNavigationComponent

tvararu:
This is a port of `x-govuk-secondary-navigation` and is required by the new patients table layout.

https://x-govuk.github.io/govuk-prototype-components/secondary-navigation/

Also add `AppCountComponent`.

### Screenshot

image


misaka (approved):
:shipit:

PR #1299, opened 2024-05-15
Use new patients table for consents index 

tvararu:
Follows the new design.

Tweak the consents one and implement captions, plus tidy up the localisations.

### Screenshots

image


misaka (approved):
🚀

PR #1300, opened 2024-05-15
Use new patients table for vaccinations index

tvararu:
Follow the new design from the prototype.

Removing the "Action needed" section and the ability to see patients that haven't been vaccinated yet in the vaccination section prompted a pretty extensive cleanup of a lot of legacy playwright tests that either are already being covered by rspec feature tests, or that won't be relevant in the redesign.

See commits for individually removed tests and their explanations.

### Screenshots

image


misaka (approved):
A bit of 🥾 🥫 in here, but that's fine, the prototype is in flux. :shipit:

PR #1301, opened 2024-05-16
Remove playwright tests 🪓 

tvararu:
See commits.


spec/features/user_authorisation_spec.rb
```diff
@@ -63,4 +63,8 @@ def then_i_should_see_page_not_found
   def when_i_go_to_the_consent_page_of_another_team
     visit "/sessions/#{@other_session.id}/consent"
   end
+
+  def when_i_go_to_the_session_page_of_another_team
+    visit "/sessions/#{@other_session.id}"
+  end
```
misaka, line 69: I think we need to add a test that a nurse can't see a different patient. I know we went back and forth on this a little, but we literally had a bug that was showing the wrong patient when you clicked on them (because the `patient_id` was being mixed up with a `patient_session_id`).

PR #1302, opened 2024-05-16
Extract and refactor logic from AppHealthQuestionsComponent

benilovj:
This introduces a ConsolidatedHealthAnswers class with a slightly different implementation and more thorough unit testing.

### DIFF
--- a/app/models/consolidated_health_answers.rb
+++ b/app/models/consolidated_health_answers.rb
@@ -1,0 +1,30 @@
+class ConsolidatedHealthAnswers
+  def initialize
+    @answers = {}
+  end
+
+  def add_answer(responder:, question:, answer:, notes: nil)
+    @answers[question] ||= []
+    @answers[question] << { responder:, answer:, notes: }
+  end
+
+  def to_h
+    consolidated_answers
+  end
+
+  private
+
+  def consolidated_answers
+    @answers.transform_values { |answers| consolidate_answers(answers) }
+  end
+
+  def consolidate_answers(answers)
+    if answers.length == 1
+      answers
+    elsif answers.uniq { |a| [a[:answer], a[:notes]] }.length == 1
+      [answers.first.merge(responder: "All")]
+    else
+      answers
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


app/models/consolidated_health_answers.rb
```diff
@@ -0,0 +1,30 @@
...
+    @answers.transform_values { |answers| consolidate_answers(answers) }
+  end
+
+  def consolidate_answers(answers)
+    if answers.length == 1
+      answers
+    elsif answers.uniq { |a| [a[:answer], a[:notes]] }.length == 1
+      [answers.first.merge(responder: "All")]
+    else
+      answers
```
tvararu, line 27: "I can golf this"
"Wait"
"I can't golf this"

tvararu (approved):

app/models/consolidated_health_answers.rb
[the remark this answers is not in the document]
misaka: I think I agree with the golfing attempt here, but it might also make it more readable to use a helper method for `answers.uniq { |a| [a[:answer], a[:notes]] }` somehow.

app/models/consolidated_health_answers.rb
```diff
@@ -0,0 +1,30 @@
...
+  def initialize
+    @answers = {}
+  end
+
+  def add_answer(responder:, question:, answer:, notes: nil)
+    @answers[question] ||= []
+    @answers[question] << { responder:, answer:, notes: }
+  end
+
+  def to_h
```
misaka, line 11: As @benilovj mentioned in person, some comments similar to those removed from `AppHealthQuestionsComponent` would help.

app/models/consolidated_health_answers.rb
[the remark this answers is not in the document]
benilovj: Agree with suggestion, incorporated in eb0be58f083225739a519d7bbdd3a4d67d09125d

app/models/consolidated_health_answers.rb
benilovj: Added in eb0be58f083225739a519d7bbdd3a4d67d09125d

misaka (approved):
🚢

PR #1307, opened 2024-05-18
Patient table consent status: new designs

benilovj:
Scenario|Before             |  After
:-------------:|:-------------:|:----------------:
Consent given|image|image
Consent refused|image|image
Conflicting consent|image|image

One point to check: the prototype doesn't have any icon for conflicting consent. Is this deliberate or not? I kept the cross as it seemed more visually obvious but happy to remove it if its omission is deliberate.

### DIFF
--- a/app/assets/stylesheets/_status.scss
+++ b/app/assets/stylesheets/_status.scss
@@ -8,8 +8,8 @@
     margin-left: nhsuk-spacing(1) * -1;
   }
 
-  &--green {
-    color: $color_nhsuk-green;
+  &--blue {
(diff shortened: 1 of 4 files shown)


tvararu (approved):

benilovj:
> One point to check: the prototype doesn't have any icon for conflicting consent. Is this deliberate or not? I kept the cross as it seemed more visually obvious but happy to remove it if its omission is deliberate.

I've clarified that we do indeed want to keep the cross for conflicting consent, so the code does what it's supposed to now – merging.

PR #1308, opened 2024-05-20
Use turbo for table sorting

tvararu:
This implements the `<action-table>` sorting logic server-side in Ruby, and combines with Turbo to offer a similar user experience.

TODO:

- [x] CSS 🎨


misaka (approved):
🚀

PR #1310, opened 2024-05-20
Remove action table

tvararu:
It will be replaced with Turbo in a follow-up PR.

This also adds Turbo and uses it for the secondary navigation.


app/views/layouts/application.html.erb
```diff
@@ -11,6 +11,9 @@
 
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
+
+    <%= turbo_refreshes_with method: :morph, scroll: :preserve  %>
+    <%= yield :head %>
```
misaka, line 16: Do we really want `scroll: preserve` everywhere? Will this preserve the scroll position between, for example, the consents index and the patient page?

misaka: I've answered my own question: nope. I'm guessing it's linked to the `data: { turbo_action: 'replace' }` parameter added to the tab links, but I'm still trying to understand why client-side navigation is global with the above change, but only the tabs preserve scroll position.

tvararu: Oh and yeah, I did test it and scroll isn't preserved anywhere in the app, as expected...

misaka: Ya, had a read of that doc, very useful. So `replace` also / really is designed to reload the page without changing the browsing history, meaning that the "back" button won't work between  tabs. For example, if we're on the "no consent" tab in the consents section, and click on "given", using the browser back button won't take us to "no consent", but to wherever we were before that.

I think that that's probably good, and works with the scroll position not changing. 👍

PR #1324, opened 2024-05-21
Patient page: use a table to present consents, tuck the consent details away onto a dedicated page

benilovj:
This is a follow-on from https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1307.

## Before – patient page

image

## After – patient page

image

## After – consent details page (consent given)

consent-given

## After – consent details page (consent refused)

consent-refused

## Reviewer notes

While the previous implementation used a view component, the new implementation is done with a view, and is predominantly tested by feature specs. This approach trades indirection and complexity for reduced reusability.

Looking ahead, consent details will need to be presented in a very similar way on the confirmation page of the redesigned consent recording flow; that page will include edit links so the presentation may not be directly reusable.

### DIFF
--- a/app/components/app_consent_component.html.erb
+++ b/app/components/app_consent_component.html.erb
@@ -7,7 +7,6 @@
 <% end %>
 
 <% if patient_session.consents.present? %>
-  <% consents = patient_session.consents.order(:recorded_at) %>
   <%= govuk_table do |table| %>
     <%= table.with_head do |head| %>
       <%= head.with_row do |row| %>
@@ -18,7 +17,7 @@
     <% end %>
 
     <%= table.with_body do |body| %>
-      <% patient_session.consents.order(recorded_at: :desc).each do |consent| %>
+      <% consents.each do |consent| %>
         <%= body.with_row do |row| %>
           <%= row.with_cell do %>
             <%= govuk_link_to consent.parent_name, details_session_patient_manage_consents_path(consent, session_id: session.id, patient_id: patient.id, section:, tab:) %>
--- a/app/components/app_consent_component.html.erb
+++ b/app/components/app_consent_component.html.erb
@@ -21,7 +21,7 @@
       <% patient_session.consents.order(recorded_at: :desc).each do |consent| %>
         <%= body.with_row do |row| %>
           <%= row.with_cell do %>
-            <%= consent.parent_name %>
+            <%= govuk_link_to consent.parent_name, details_session_patient_manage_consents_path(consent, session_id: session.id, patient_id: patient.id, section:, tab:) %>
             <br>
             <span class="nhsuk-u-font-size-16"><%= consent.human_enum_name(:parent_relationship).humanize %></span>
           <% end %>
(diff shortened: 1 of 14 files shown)


app/components/app_consent_component.html.erb
```diff
@@ -7,28 +7,47 @@
 <% end %>
 
 <% if patient_session.consents.present? %>
-  <% consents_grouped_by_parent.each do |summary, consents| %>
-    <%= render AppDetailsComponent.new(summary:, open: open_consents?) do %>
-      <%= render AppConsentSummaryComponent.new(**grouped_consents(consents)) %>
+  <% consents = patient_session.consents.order(:recorded_at) %>
```
tvararu, line 10: I think this could be a method inside the component Ruby class:

```ruby
def consents
  patient_session.consents.order(:recorded_at)
end
```

app/components/app_consent_component.html.erb
```diff
@@ -7,28 +7,47 @@
...
-            patient,
-            consents.first.id,
-            :agree,
-            section: @section,
-            tab: @tab
-          ),
-          secondary: true
-        ) %>
+    <%= table.with_body do |body| %>
+      <% patient_session.consents.order(recorded_at: :desc).each do |consent| %>
```
tvararu, line 21: Since there's only ever a few consents, you can also rely on `order(:recorded_at)` being applied on the previous `consents` method and do:

```ruby
consents.reverse.each do
```

Alternatively, just not a fan of coming up with names for this:

```ruby
def consents_reversed
  patient_session.consents.order(recorded_at: :desc)
end
```

app/views/consents/show.html.erb
```diff
@@ -0,0 +1,116 @@
+<% page_title = "Consent response from #{@consent.name}" %>
+
+<% content_for :before_main do %>
+  <%= render AppBacklinkComponent.new(
+    href: session_patient_path(id: @consent.patient.id),
+    name: "patient page",
+  ) %>
+<% end %>
+
+<%= h1 page_title, page_title: %>
```
tvararu, line 10: `page_title` is automatically set if it's the only argument to `h1`:

```suggestion
<% content_for :before_main do %>
  <%= render AppBacklinkComponent.new(
    href: session_patient_path(id: @consent.patient.id),
    name: "patient page",
  ) %>
<% end %>

<%= h1 "Consent response from #{@consent.name}" %>
```

tvararu (approved):

app/components/app_consent_component.html.erb
benilovj: Yup, thanks for catching – I changed this component too quickly and missed another clean-up opportunity. Fixed in [b5ba1bd](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1324/commits/b5ba1bd3a80d7293edf086cc407d10728be8b195)

app/components/app_consent_component.html.erb
benilovj: Yup, fixed in Fixed in [b5ba1bd](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1324/commits/b5ba1bd3a80d7293edf086cc407d10728be8b195)

PR #1330, opened 2024-05-22
Bring back a list of children to vaccinate

benilovj:
Without this, the vaccinator needs to navigate to the specific child's page to record a vaccination.

Before:

image

After:

image

### DIFF
--- a/app/controllers/concerns/patient_tabs_concern.rb
+++ b/app/controllers/concerns/patient_tabs_concern.rb
@@ -18,6 +18,7 @@ module PatientTabsConcern
       ]
     },
     vaccinations: {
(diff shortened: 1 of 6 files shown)


paulrobertlloyd:
Got something to demo tomorrow which I think should be a lot clearer.

benilovj:
> Got something to demo tomorrow which I think should be a lot clearer.

👍 I did spot the new 'on the day' view in the prototype, if that's what you mean.
That does look like a good improvement, I just want to make this change temporarily so I can show the latest changes in a live demo tomorrow...

tvararu (approved):

PR #1331, opened 2024-05-23
Add hotwire-livereload in development

tvararu:
This enables live reloading of CSS, JS, and even HTML in local development, improving the developer experience by shortening the feedback loop between making a change and seeing that change reflected in the frontend.

I've found some caveats when testing this locally:

- CSS is a cascading language, so updates are applied on top of each other, and code that has already been declared can't be "undeclared." As such, if you remove CSS and live-reload a new version of the CSS, those old rules will persist until a hard reload
- The JS for `nhsuk-button` isn't idempotent, and applying it multiple times to the same elements leads to a JS error. I believe it's harmless
- When reloading HTML, the turbo websocket appears to temporarily disconnect. It reconnects on its own in a few seconds but if you attempt to make multiple quick changes back to back, they won't all get livereloaded
- [hotwire-livereload](https://github.com/kirillplatonov/hotwire-livereload) suggests that it depends on Redis to work, but that's not my experience, as it works without in my other projects. But perhaps Redis is necessary to fix the HTML livereloading issue. I don't think it's worth the complexity cost in adding Redis, so let's not, especially since Rails 8 will include improvements that make Redis less relevant for small apps and local development

Even with those caveats, it's worth adding this and using it for local dev.

### JS console output when livereloading

Screenshot 2024-05-23 at 09 52 19

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -6,14 +6,15 @@
 
     <title><%= page_title(@service_name) %></title>
 
-    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
-    <%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>
+    <%= stylesheet_link_tag "application", "data-turbo-track": Rails.env.development? ? "" : "reload" %>
+    <%= javascript_include_tag "application", "data-turbo-track": Rails.env.development? ? "" : "reload", defer: true %>
 
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
     <%= turbo_refreshes_with method: :morph, scroll: :preserve  %>
     <%= yield :head %>
+    <%= hotwire_livereload_tags if Rails.env.development? %>
   </head>
 
   <body class="nhsuk-template__body<%= ' app-signed-in' if current_user.present? %>">
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,7 @@ group :development do
   gem "asciidoctor"
   gem "asciidoctor-diagram"
   gem "dockerfile-rails", ">= 1.0.0"
+  gem "hotwire-livereload"
   gem "prettier_print", require: false
   gem "rails-erd"
   gem "rladr"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -175,6 +175,7 @@ GEM
       concurrent-ruby (~> 1.1)
       webrick (~> 1.7)
       websocket-driver (>= 0.6, < 0.8)
+    ffi (1.16.3)
     flipper (1.3.0)
       concurrent-ruby (< 2)
     flipper-active_record (1.3.0)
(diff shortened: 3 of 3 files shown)


misaka:
> - CSS is a cascading language, so updates are applied on top of each other, and code that has already been declared can't be "undeclared." As such, if you remove CSS and live-reload a new version of the CSS, those old rules will persist until a hard reload

Does this mean that when we deploy changes to the CSS, clients may not see it under certain circumstances? Or does a hard-reload get forced somehow ... or can we force a hard-reload ourselves?

misaka:
> - When reloading HTML, the turbo websocket appears to temporarily disconnect. It reconnects on its own in a few seconds but if you attempt to make multiple quick changes back to back, they won't all get livereloaded

What happens once it reconnects, does it do the live-reload then?

tvararu:
> > * CSS is a cascading language, so updates are applied on top of each other, and code that has already been declared can't be "undeclared." As such, if you remove CSS and live-reload a new version of the CSS, those old rules will persist until a hard reload
> 
> Does this mean that when we deploy changes to the CSS, clients may not see it under certain circumstances? Or does a hard-reload get forced somehow ... or can we force a hard-reload ourselves?

This functionality is only for development, it has no effect on production/deploys.

> > * When reloading HTML, the turbo websocket appears to temporarily disconnect. It reconnects on its own in a few seconds but if you attempt to make multiple quick changes back to back, they won't all get livereloaded
> 
> What happens once it reconnects, does it do the live-reload then?

It drops any live reloads that happened in the interim but becomes available for subsequent ones.

> > * [hotwire-livereload](https://github.com/kirillplatonov/hotwire-livereload) suggests that it depends on Redis to work, but that's not my experience, as it works without in my other projects. But perhaps Redis is necessary to fix the HTML livereloading issue. I don't think it's worth the complexity cost in adding Redis, so let's not, especially since Rails 8 will include improvements that make Redis less relevant for small apps and local development
> 
> Is it possible Redis is required in a multi-server setup? That would've been my expectation ... without it, I would imagine there are certain instances where live reloads wouldn't work, depending on which server the client gets load-balanced to.

It's required because of some limitations with the Async adapter for ActionCable: https://github.com/kirillplatonov/hotwire-livereload/issues/20#issuecomment-1141280214... There shouldn't be multiple clients in local development

app/views/layouts/application.html.erb
```diff
@@ -6,14 +6,15 @@
 
     <title><%= page_title(@service_name) %></title>
 
-    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
-    <%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>
+    <%= stylesheet_link_tag "application", "data-turbo-track": Rails.env.production? ? "reload" : "" %>
+    <%= javascript_include_tag "application", "data-turbo-track": Rails.env.production? ? "reload" : "", defer: true %>
```
misaka, line 10: Should this be:

```suggestion
    <%= javascript_include_tag "application", "data-turbo-track": Rails.env.development? ? "" : "reload", defer: true %>
```

Just to ensure it's either live reload for dev and full page reload for any other env?

misaka (approved):

PR #1334, opened 2024-05-23
Remove JSON example campaign code and data

misaka:
We've created a new `/reset` endpoint that generates an example campaign, and have removed the Playwright tests that relied on these JSON example campaigns. Removing these does mean that if we need to generate static/predictable campaign data again in the future, we'll need to develop a new way of doing it, but we'd prefer to do this with the factories directly instead of using an intermediary JSON file.


benilovj (approved):

tvararu (changes requested):
Think you need to update `bin/heroku-release` as well and remove the reference to `load_campaign_example`. Do a quick `grep` before and see if there's any other references.

misaka:
This also means that our review apps won't have example data loaded. Do we want add (in a later PR) a way to load up example data from a rake task?

tvararu:
> This also means that our review apps won't have example data loaded. Do we want add (in a later PR) a way to load up example data from a rake task?

Once we fix the review apps, we'll worry about that 😂

tvararu (approved):

PR #1335, opened 2024-05-23
Remove team cohort limits as they're unnecessary

benilovj:
These were only in place for the pilot.

## Before

image

## After

image


misaka (approved):
🪓

tvararu (approved):

PR #1336, opened 2024-05-23
Allow users to filter patients by name/dob 

tvararu:
This implements filtering using a server-side request and turbo to progressively enhance it to be in-place.

A form with a method of `GET` is added to the patients table that sends name/dob parameters to the server. These params are handled in the controller to filter the list of patients down. The `sort` and `direction` parameters are maintained through hidden fields, allowing sorting and filtering to occur at the same time.

An `autosubmit` Stimulus controller then progressively enhances the form by hiding the submit button and automatically submitting the fields for the user 250ms after their last input. The content of the input fields are maintained using `data-turbo-permanent`. There is a `Reset filters` button that is set to disabled when the fields are empty and enabled when not.

The `patient_sorting_filtering_spec` is updated with a second scenario that has a type of `:system` and runs with JS. It tests that the submit button disappears, and that filtering and reseting works as expected.

Can be reviewed commit by commit.

TODO:

- [x] CSS 🎨 
- [x] Refactor controller logic to concern
- [x] Fix any broken specs

### Screenshots

image

#### Filtering by name

image

#### Simultaneous sorting/filtering

image

#### Mobile has filters at start

image

### DIFF
--- a/app/assets/stylesheets/_patients.scss
+++ b/app/assets/stylesheets/_patients.scss
@@ -1,28 +1,34 @@
 .app-patients {
   @include top-and-bottom;
 
+  align-items: start;
   background-color: $color_nhsuk-white;
   display: flex;
   flex-wrap: wrap;
 
-  & .app-patients__filters {
+  & &__filters {
     @include nhsuk-responsive-padding(4);
 
+    align-self: stretch;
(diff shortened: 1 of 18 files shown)


misaka (approved):
Nice! Love it. 🚢

PR #1337, opened 2024-05-23
Reverse order of parameters in add_new_location Rake task

benilovj:
This makes the all-in-one invocation consistent with the order in which the questions are asked in interactive mode.


misaka (approved):

tvararu:
1337 PR 🤙

PR #1356, opened 2024-05-29
Reformat patient view HTML

tvararu:
This is split out from https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1355 as it's not exactly related to the work there.

See commits.

### DIFF
--- a/app/components/app_patient_page_component.html.erb
+++ b/app/components/app_patient_page_component.html.erb
@@ -1,5 +1,3 @@
-<%= h1 patient.full_name, id: dom_id(patient) %>
-
 <%= if patient_session.state.in? %w[
                                   vaccinated
                                   triaged_do_not_vaccinate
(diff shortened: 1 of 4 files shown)


app/components/app_patient_page_component.html.erb
```diff
@@ -32,12 +39,16 @@
...
-        session_patient_manage_consents_path(session, patient_id: patient.id, section: @section, tab: @tab),
-        secondary: true,
-        name: "consent", value: "self_consent",
-        prevent_double_click: true
-      ) %>
+            "Assess Gillick competence",
+            session_patient_manage_consents_path(
+              session, patient_id: patient.id,
+                       section: @section,
+                       tab: @tab,
```
misaka, line 44: What's with this formatting? Haven't seen this style before.

tvararu: Rufo 🤷

PR #1358, opened 2024-05-30
Fix flakey patient sorting spec

tvararu:
This test was relying on a race condition and the hidden behaviour of `expect` in rspec system tests.

Because system specs often trigger JavaScript and network requests, they have a built-in "sleep". If the assertion fails immediately, they try again after a short while, and only fail if all the tries fail.

This test was sorting the list as such:

```
- Cassidy
- Casey
- Blair
- Alex
```

And then testing that after filtering for `cas` that we're left with:

```
expect Cassidy
expect Casey
expect no third item
```

However, the implementation had a bug. Namely that `sort` and `direction` parameters were being lost upon using the name filtering. So the actual list from the server would be:

```
- Casey
- Cassidy
```

So why did the test succeed? The first two `expects` should fail. The reason comes back to the aforementioned built-in sleeps: in some scenarios, the first two `expect`s would run and succeed as they run immediately, before the server replies with the new HTML. Then the third would wait for the filtering to finish and then succeed as well, despite the fact that now the sorting has been lost.

This commit makes the tests more reliable by checking for the total number of items first.

It also fixes the underlying issue of losing the sorting order when filtering. `data-turbo-permanent` on the wrapping `<form>` was causing the hidden `sort` and `direction` fields to persist between page morphs, so instead this makes the buttons and `name`/`dob` fields permanent but the rest impermanent.

### DIFF
--- a/app/components/app_patient_table_component.html.erb
+++ b/app/components/app_patient_table_component.html.erb
@@ -5,28 +5,31 @@
                   class: "app-patients__filters",
                   data: { module: "autosubmit",
                           turbo: "true",
-                          turbo_action: "replace",
-                          turbo_permanent: "" },
+                          turbo_action: "replace" },
                   builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
       <%= f.govuk_fieldset legend: { text: "Filter results", size: "s" } do %>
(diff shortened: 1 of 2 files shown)


benilovj (approved):
Top sleuthing 🕵️

PR #1360, opened 2024-05-30
Refactor activity log to separate component

tvararu:
Prerequisite: https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1355

This refactors the log to a component, allowing us to more easily unit test the output against multiple scenarios without relying on heavy feature tests.

It also improves the component by adding the consolidated headers and by creating an `events` array that has a unified structure in the component class. We can add to this with all the different event types in follow-up PRs.

TODO:

- [x] Add consolidated headings as per the prototype
- [x] Component preview screenshots

### Component preview

http://localhost:4000/rails/view_components/app_activity_log_component/default

image

### DIFF
--- a/app/components/app_activity_log_component.rb
+++ b/app/components/app_activity_log_component.rb
@@ -1,0 +1,26 @@
+class AppActivityLogComponent < ViewComponent::Base
+  def initialize(patient_session)
+    super
+
+    @patient_session = patient_session
+  end
+
+  def events_by_day
+    consent_events.group_by { _1[:time].to_fs(:nhsuk_date) }
+  end
+
+  def consent_events
+    @patient_session
+      .patient
+      .consents
+      .recorded
+      .order(recorded_at: :desc)
+      .map do |consent|
+        {
+          title:
+            "Consent #{consent.response} by #{consent.parent_name} (#{consent.who_responded})",
+          time: consent.recorded_at
+        }
+      end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/components/app_activity_log_component.rb
```diff
@@ -0,0 +1,26 @@
...
+  def consent_events
+    @patient_session
+      .patient
+      .consents
+      .recorded
+      .order(recorded_at: :desc)
+      .map do |consent|
+        {
+          title:
+            "Consent #{consent.response} by #{consent.parent_name} (#{consent.who_responded})",
```
misaka, line 21: Just noticing the who_responded can be blank in our test data, is that realistic?

image

misaka (approved):

app/components/app_activity_log_component.rb
benilovj: This shouldn't ever be blank. @misaka can you figure out what the test data is missing, so we might fix it in the generator? e.g. we not populating `consents#parent_relationship_other` if `consents#parent_relationship` is equal to `other`...

Also "Consent #{consent.response} by #{consent.parent_name} (#{consent.who_responded})" will show the wrong thing for self-consent, I think.

tvararu: Yeah, it's a problem with the data. I'll look into this 👍

benilovj: Thanks – there's also the 2nd part which I think is still an issue in the code:

> Also "Consent #{consent.response} by #{consent.parent_name} (#{consent.who_responded})" will show the wrong thing for self-consent, I think.

misaka: Ya, I realised this as well when I tried a different approach to fixing the factories, I tried adding validations that worked outside of the wizard steps. That caught out that self-consent didn't set `parent_relationship`.

I think we should have validations that work outside of the wizard steps to catch these issues.

PR #1361, opened 2024-05-30
Add more events to activity log

tvararu:
Prerequisite: https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1360

Events:

- [x] "Added to session"
- [x] "Triage decision"
- [x] "Vaccinated"

Further refinements:

- [ ] Fix content of "Triage decision" notes (follow-up, after demo)

### Preview 📸 

http://localhost:4000/rails/view_components/app_activity_log_component/default

image

### DIFF
--- a/app/components/app_activity_log_component.html.erb
+++ b/app/components/app_activity_log_component.html.erb
@@ -12,6 +12,7 @@
         </h3>
         <p class="nhsuk-body-s nhsuk-u-margin-0 nhsuk-u-secondary-text-color">
           <%= event[:time].to_fs(:app_date_time) %>
+          <% if event[:by] %> · <%= event[:by] %><% end %>
         </p>
       </div>
     </div>
--- a/app/components/app_activity_log_component.rb
+++ b/app/components/app_activity_log_component.rb
@@ -12,7 +12,17 @@ class AppActivityLogComponent < ViewComponent::Base
   end
 
   def all_events
-    [session_events, consent_events, triage_events].flatten
+    [vaccination_events, triage_events, consent_events, session_events].flatten
+  end
+
+  def vaccination_events
+    @patient_session.vaccination_records.map do
+      {
+        title: "Vaccinated with #{helpers.vaccine_heading(_1.vaccine)}",
+        time: _1.created_at,
+        by: _1.user.full_name
+      }
+    end
   end
 
   def triage_events
--- a/app/components/app_activity_log_component.rb
+++ b/app/components/app_activity_log_component.rb
@@ -12,7 +12,17 @@ class AppActivityLogComponent < ViewComponent::Base
   end
 
   def all_events
-    [session_events, consent_events].flatten
+    [session_events, consent_events, triage_events].flatten
+  end
+
+  def triage_events
+    @patient_session.triage.map do
+      {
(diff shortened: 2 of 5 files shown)


app/components/app_activity_log_component.html.erb
```diff
@@ -12,6 +12,7 @@
         </h3>
         <p class="nhsuk-body-s nhsuk-u-margin-0 nhsuk-u-secondary-text-color">
           <%= event[:time].to_fs(:app_date_time) %>
+          <% if event[:by] %> · <%= event[:by] %><% end %>
```
misaka, line 15: Is this rufo's idea of proper indentation?

app/components/app_activity_log_component.rb
```diff
@@ -6,21 +6,52 @@ def initialize(patient_session)
...
+  end
+
+  def all_events
+    [vaccination_events, triage_events, consent_events, session_events].flatten
+  end
+
+  def vaccination_events
+    @patient_session.vaccination_records.map do
+      {
+        title: "Vaccinated with #{_1.vaccine.brand} (#{_1.vaccine.type})",
```
misaka, line 21: We should be using the I18n version of the vaccine type here.

```suggestion
        title: "Vaccinated with #{_1.vaccine.brand} (#{t("vaccines." + _1.vaccine.type)})",
```

Interpolating it like this gets a bit ugly, I've created a helper for the vaccines index and show page that might be re-usable here.

https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1364/files#diff-7357383d3f6e7f3684590630a1692c540ef8769ebee862a97e63b6adef736e49R2

misaka: Also, this'll break with the current factories, as discussed.

misaka: There's not much left on that PR, btw, I think it just needs a test and a little tidy up

app/components/app_activity_log_component.html.erb
tvararu: I think I collapsed it to one line but it looks fine to me?

app/components/app_activity_log_component.rb
tvararu: Right, will update once your PR is in 👍

app/components/app_activity_log_component.html.erb
misaka: Just looks odd because you wouldn't write Ruby on one line that way, but ... ERB I guess ¯\\\_(ツ)_/¯

tvararu: Alternative would be something like `" · #{event[:by]}" if event[:by]` but that's equally weird as it's a statement that evaluates to a string which normally doesn't do anything... Yeah ¯\\_(ツ)\_/¯

misaka: :shipit:

misaka (approved):
Your choice whether to wait for my PR or just cut a new one for the I18n-isation of `type`.

app/components/app_activity_log_component.rb
tvararu: Updated!

PR #1364, opened 2024-05-31
Add vaccines show page

misaka:
Adds show page for vaccines.

image

The layout is currently two-thirds but looks like it should be three-quarters looking at the prototype. To be addressed later. The prototype also adds the "Dose" field and pre-screening questions, which we don't have yet. Will add these later.

### DIFF
--- a/app/controllers/vaccines_controller.rb
+++ b/app/controllers/vaccines_controller.rb
@@ -1,8 +1,14 @@
 class VaccinesController < ApplicationController
   include TodaysBatchConcern
 
+  layout "two_thirds"
+
   def index
     @vaccines = policy_scope(Vaccine).order(:name)
     @todays_batch_id = todays_batch_id
+  end
+
+  def show
+    @vaccine = policy_scope(Vaccine).find(params[:id])
   end
 end
--- a/app/helpers/vaccines_helper.rb
+++ b/app/helpers/vaccines_helper.rb
@@ -1,0 +1,5 @@
+module VaccinesHelper
+  def vaccine_heading(vaccine)
+    "%s (%s)" % [vaccine.brand, t(vaccine.type.downcase, scope: "vaccines")]
+  end
+end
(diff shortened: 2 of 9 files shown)


app/controllers/vaccines_controller.rb
```diff
@@ -1,8 +1,14 @@
 class VaccinesController < ApplicationController
   include TodaysBatchConcern
 
+  layout "two_thirds"
```
misaka, line 4: This doesn't look quite right , but it's broken in the same way other pages are broken ... I think the prototype uses a three-quarters layout? 🥾 🥫

tvararu: Kick the can yeah, we don't have the 3-quarters layout yet, asked on Slack about it

spec/features/manage_vaccines_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  def then_i_see_an_hpv_vaccine_listed
+    expect(page).to have_content("Gardasil 9 (HPV)")
+  end
+
+  def when_i_click_on_the_vaccine
+    click_on "Gardasil 9 (HPV)"
+  end
+
+  def then_i_see_the_vaccine_details
+    expect(page).to have_selector :heading, "Gardasil 9 (HPV)"
```
misaka, line 38: @tvararu This is slightly new, dunno how you feel about testing this way but let me know.

app/helpers/vaccines_helper.rb
```diff
@@ -0,0 +1,5 @@
+module VaccinesHelper
+  def vaccine_heading(vaccine)
+    sprintf("%s (%s)", vaccine.brand, t(vaccine.type, scope: "vaccines"))
```
tvararu, line 3: Yeah, `sprintf` is clearer than string interpolation isn't it 👍

misaka: In this case. I wanted to do `"%s (%s)" % [vaccine.brand, t(vaccine.type, scope: "vaccines")]" but Rubocop got pushy about it ... I think it has a chip on it's shoulder about looking too much like Python.

tvararu (approved):

spec/features/manage_vaccines_spec.rb
tvararu: LGTM

PR #1365, opened 2024-05-31
Delete phase banner

tvararu:
This makes side-by-side comparisons with the prototype easier.

The consensus for on the [NHS Design System issue for the Phase Banner](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/89) appears to be in favour of not using phase banners, though some research has been done with regards to using the pattern as a way to ask for feedback, with good results.


misaka:
There's also bits in `controllers/application_controller.rb` that sets `@phase_banner_text` which is mostly Pilot-related. Do we want remove that too?

tvararu:
@misaka 🪓

misaka (approved):

PR #1366, opened 2024-05-31
Use rufo to format erb files

tvararu:
We already use `prettier` for Ruby, JS/TS, (S)CSS, plain HTML, markdown, and yml, but unfortunately it doesn't support `.erb` which is a significant portion of our codebase.

[`rufo`](https://github.com/ruby-formatter/rufo) does however and it has good editor support.

It's not as deterministic as `prettier` and there are some edge cases where it "changes its mind" and formats identical input in two different ways. But for the most part it seems to work well enough.

Second commit is really long, probably not worth reviewing. We can always manually tweak awkward styling in follow-up PRs.

### DIFF
--- a/bin/lint
+++ b/bin/lint
@@ -2,6 +2,7 @@
 
 bin/bundle exec rubocop --autocorrect-all $*
 yarn prettier --write --list-different --ignore-unknown ${*:-'**/*'}
+bin/bundle exec rufo --check ${*:-app}
 if [ -z "$*" ]
 then
    bin/bundle exec brakeman --quiet --no-summary --no-pager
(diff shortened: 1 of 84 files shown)


bin/lint
```diff
@@ -2,6 +2,7 @@
 
 bin/bundle exec rubocop --autocorrect-all $*
 yarn prettier --write --list-different --ignore-unknown ${*:-'**/*'}
+bin/bundle exec rufo --check app
```
misaka, line 5: Does rufo take cmdline args? Can we add `$*`?

misaka (approved):
🚢

bin/lint
tvararu: Yeah

PR #1367, opened 2024-06-01
Move all Gillick assessment-related data onto own model

benilovj:
Currently, the data related to Gillick assessment is stored on `PatientSession`. This works fine, but it means

* the Gillick assessment recording wizard has only one place to record the data – it's much harder to have a "draft" which can be discarded if the wizard is abandoned before confirmation.
* the `PatientSession` model is more complex than it needs to be

This change:

* introduces a separate `GillickAssessment` that belongs to `PatientSession`, and rewires all flows to use it. 
* adds a `GillickAssessment#recorded_at` which allows draft functionality

In a subsequent PR, recording a Gillick assessment will become its own wizard flow (in alignment with the prototype) – this will be much easier after this refactoring.


app/components/app_gillick_card_component.html.erb
```diff
@@ -8,13 +8,14 @@
           summary_list.with_row do |row|
             row.with_key { "Assessment" }
             row.with_value do
-              "Assessed by #{@patient_session.gillick_competence_assessor.full_name}" +
-              tag.br +
-              "#{@consent.recorded_at || @consent.created_at}.to_fs(:nhsuk_date)"
+              safe_join([
+                "Assessed by #{@assessment.assessor.full_name}",
+                (@assessment.recorded_at || @assessment.created_at).to_fs(:nhsuk_date),
+              ], tag.br)
```
tvararu, line 14: Thanks, was there an injection here without the `safe_join`? 🤔 

I did this change recently as part of the Rufo formatting work

tvararu (approved):

app/components/app_gillick_card_component.html.erb
benilovj: Well this bit definitely had a bug in it:

```
"#{@consent.recorded_at || @consent.created_at}.to_fs(:nhsuk_date)"
```

but I wouldn't worry about this component because I think it's going to go away very soon (it's been designed out).

PR #1389, opened 2024-06-07
Fix triage snags

tvararu:
See commits, these were more snags found on the way to implementing the `Update triage outcome` link.


app/controllers/triage_controller.rb
```diff
@@ -94,7 +83,7 @@ def process_triage
     @triage.assign_attributes triage_params.merge(user: current_user)
     if @triage.save(context: :consent)
       @patient_session.do_triage!
-      send_triage_mail(@patient_session, @consent)
+      @patient.consents.each { send_triage_mail(@patient_session, _1) }
```
benilovj, line 86: Is this going to email the same parent multiple times if they've submitted duplicate consents?

tvararu: A parent should only ever have 1 consent... The way we model it is that a consent is a parent's latest decision. We update them in place rather than creating new records if they change their mind

If a patient has multiple consent objects from the same parent that's a separate bug, and we shouldn't work around it at this step

tvararu: Carded this up, lest we forget

PR #1391, opened 2024-06-07
Make consent recording in Mavis more generic

benilovj:
## User changes

* Allow staff to choose how the consent was obtained (by phone, online, in person, paper)
* Always show the "Get consent" button, even if it's there is already consent
* Remove the ability to contact refusers – staff can now do this by recording a new consent
* Remove the ability to update consents where the parent couldn't be reached (no response) – instead, staff will recording a new consent

## Technical changes

* Remove cloning functionality from `ManageConsentsController` (no longer needed)
* Remove `not null` constraint from `consents.route` so it can be set during the wizard
* Add validation to ensure that `Consent#route` is set for non-draft consents
* Bug fix: always set `recorded_by` for draft consents

## Patient page

image

## How was the response given?

image

## Error

image


app/components/app_consent_component.html.erb
```diff
@@ -30,30 +30,13 @@
...
-            session,
-            patient,
-            consents.first.id,
-            :agree,
-            section: @section,
-            tab: @tab,
-          ),
-          secondary: true,
-        ) %>
+<p body="govuk-body">
```
tvararu, line 35:
```suggestion
<p>
```

tvararu (approved):

app/components/app_consent_component.html.erb
benilovj: 👍

PR #1392, opened 2024-06-08
Move the patient's parent info to a separate Parent model

benilovj:
There are currently 3 places where parent data is stored:

* the patient record, imported during cohorting
* the online consent form filled out by the parent themselves
* the consent recorded by a member of staff (picked from existing parent contact details, or a new contact recorded)

The parent fields are almost identical in all 3 places. Keeping the data all on the same `Parent` model and dealing with `parent_id`s would:

* reduce the number of copies of PII stored
* make it easier to avoid creating and keeping around duplicate parent contacts
* make it easier to determine "the same parent" in cases of consent conflicts

This PR is step 1, which introduces the `Parent` model and removes those fields off `Patient`. Subsequent PRs will do the same for `ConsentForm` and `Consent`.


tvararu (approved):
🙌

PR #1393, opened 2024-06-08
Add vaccines dose

misaka:
Doses are assumed to be `ml` until we see otherwise.

image


tvararu (approved):
As long as we won't need to perform arithmetic with dose values, this should be fine 👍

paulrobertlloyd:
> As long as we won't need to perform arithmetic with dose values, this should be fine

For nasal Flu vaccine, if only one sprayed up one nostril, you would record that as half a dose. Perhaps the same is true with a partial injection. Which is to say, there _might_ be arithmetic.

misaka:
> As long as we won't need to perform arithmetic with dose values, this should be fine 👍

What's the issue with doing arithmetic?

misaka:
> > As long as we won't need to perform arithmetic with dose values, this should be fine 👍
> 
> What's the issue with doing arithmetic?

Sorted in Slack, AR will turn `decimal` types into `BigDecimal` in Ruby so there shouldn't be any issues.

PR #1395, opened 2024-06-10
Move Consent parent info onto Parent model

benilovj:
Follow-up to #1392.


app/models/consent.rb
```diff
@@ -190,6 +192,16 @@ def reasons_triage_needed
   end
 
   def self.from_consent_form!(consent_form, patient_session)
+    parent =
```
tvararu, line 151: Can you wrap this in a transaction, to prevent the creation of orphaned parent records in case the consent creation fails later?

benilovj: Yup – added in 4534c03c80036f35b57cfee0f9c6c0f3ed3f02c0

tvararu (approved):

PR #1396, opened 2024-06-10
Delete obsolete example campaign JSON files

benilovj:
The data is no longer being mantained and does not align with the data model. Going forward, we can use factories for pentest and user testing exercises.


tvararu (approved):

tvararu:
Thanks, wanted to do this the other day but forgot!

PR #1404, opened 2024-06-12
Add CIS2 integration

misaka:
Adds a dev-designed button to the login page that allows the user to login with CIS2. As part of the login process a user record is created in Mavis for the user that's signed in.

image

The user creation flow is pretty barebones, it simply creates a user with a random password, and then ensures that user's email and name are updated whenever they login, presuming that the Care Identity service is the single source of truth for that info. However, this begs the question of how do we manage local accounts. Can a user have both (e.g. as a backup in case CIS2 is down), and if so does this mean we reuse the same user record for either CIS2 or local auth? If we don't re-use the same record, we have a unique constraint on the email field which will need to be addressed.

Also, a user isn't added to a team when they login. We can get an ODS code from CIS2 (if a user has multiple roles / orgs they'll have to choose one in CIS2 before they get redirected back to Mavis), so this may be as easy as looking it up the team that has that ODS code and then assigning them to that team. What happens if multiple teams have the same ODS code, is that possible? The National RBAC system has a concept of workgroups that may address this. Also, what do we do if their ODS code isn't found? Or should we require onboarding to Mavis so we don't do automatic user creation? 🥾 🥫 🪱 

Tasks/issues remaining:
- [x] Rollback the rollback of `process_timeout` increase, I ~don't~ think this is the right fix
- [x] Add to README - probably should be done this PR
- [x] Test fiddling with turbo, do we really need it? 🪱 
- [x] Sprinkle more `notify_safe_email` validationisms 🪱 
- [ ] Get nurse registration number from userinfo 🥾 🪱 
- [ ] Add proper feature flag 🥾 🪱 
- [ ] Add new users to a team when their account is created from cis2 login. 🥾 🥫 🪱 
- [ ] Actually use the role or activities information to authorise what the user can do. 🥾 🪣 🪱


spec/rails_helper.rb
```diff
@@ -17,7 +17,8 @@
     app,
     inspector: ENV["DEBUG_TESTS"],
     js_errors: true,
-    window_size: [1200, 800]
+    window_size: [1200, 800],
+    process_timeout: 30
```
misaka, line 21: This is the timeout error generated by the ferrum when it was unable to start the Chrome process. 

image

I'm a little suspicious of this fix because it was the same test that had this issue both times, and I would've expected this error to be triggered randomly by different tests. I only tried re-running the job twice, and both times it failed it took 2m30s (vs <1m normally) so it's possible something else was going on on GitHub's end.

app/views/users/sessions/new.html.erb
```diff
@@ -1,7 +1,8 @@
+<%= h1 "Sign in", size: "xl" %>
```
misaka, line 1: ???

spec/rails_helper.rb
misaka: After chat with Theo, this test actually uses JS so does try to load up the browser, which is very likely what's triggering this issue. Adjusting the `process_timeout` here does potentially make sense, and as this does happen around 5% of the time we'll try to increase it and see if it helps.

app/views/users/sessions/new.html.erb
misaka: This lines up with other forms that put the `h1` above the `form`, and the errors continue to render above the heading.

app/views/layouts/application.html.erb
```diff
@@ -6,8 +6,8 @@
 
     <title><%= page_title(@service_name) %></title>
 
-    <%= stylesheet_link_tag "application", "data-turbo-track": Rails.env.development? ? "" : "reload" %>
-    <%= javascript_include_tag "application", "data-turbo-track": Rails.env.development? ? "" : "reload", defer: true %>
+    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
+    <%= javascript_include_tag "application", "data-turbo-track": "reload" %>
```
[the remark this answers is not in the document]
misaka: Nope. Restored.

Gemfile
```diff
@@ -25,6 +25,8 @@ gem "jsbundling-rails"
 gem "jsonb_accessor"
 gem "mail-notify"
 gem "okcomputer"
+gem "omniauth_openid_connect"
+gem "omniauth-rails_csrf_protection"
```
tvararu, line 29: [This gem](https://github.com/cookpad/omniauth-rails_csrf_protection) says it mitigates [CVE-2015-9284](https://nvd.nist.gov/vuln/detail/CVE-2015-9284)

`CVE-2015-9284` affects the Ruby `omniauth` gem until version 2.0.0 excluding, but we're using 2.1.2

Do we actually need it? 🤔

tvararu (approved):

Gemfile
misaka: iirc I added it because something wasn't working without it. Don't remember what, though, I'll try removing it.

tvararu: 👍

PR #1406, opened 2024-06-14
Parent model improvements

benilovj:
Some minor alignments of the `Parent` model, ahead of introducing it into the parental consent flow.

### DIFF
--- a/spec/models/parent_spec.rb
+++ b/spec/models/parent_spec.rb
@@ -70,4 +70,16 @@ RSpec.describe Parent do
       expect(subject.phone).to eq(nil)
     end
   end
+
+  describe "#email=" do
+    it "strips whitespace and downcases the email" do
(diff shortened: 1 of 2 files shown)


misaka (approved):

spec/models/parent_spec.rb
```diff
@@ -53,4 +53,33 @@
...
+      expect(subject.phone).to eq("01234567890")
+    end
+
+    it "leaves nil as nil" do
+      subject = build(:parent, phone: nil)
+      expect(subject.phone).to eq(nil)
+    end
+
+    it "sets the phone number to nil if it's blank" do
+      subject = build(:parent, phone: "")
```
misaka, line 69: I would maybe test for whitespace here, but ¯\_(ツ)_/¯

benilovj: Yup, I've updated the spec.

PR #1407, opened 2024-06-16
Refactor parental consent (ConsentForm) to store parent info on Parent model

benilovj:
Final bit of refactoring, following on from #1392 and #1395.


app/controllers/parent_interface/consent_forms/edit_controller.rb
```diff
@@ -39,14 +39,25 @@ def update
...
+                          @consent_form
+                        )
+          )
+        end
+
+        # rename keys, taking parent_ out of the key
+        parent_params =
+          update_params
+            .except(:form_step)
+            .transform_keys { |key| key.to_s.gsub("parent_", "") }
```
misaka: Sorry to nitpick, but:

```suggestion
            .transform_keys { |key| key.to_s.sub("parent_", "") }
```

app/controllers/parent_interface/consent_forms/edit_controller.rb
```diff
@@ -39,14 +39,25 @@ def update
...
+        (@consent_form.parent || @consent_form.build_parent).assign_attributes(
+          parent_params
         )
```
misaka: Just for the sake of readability, could it be worth doing this:

```suggestion
        consent_object = @consent_form.parent || @consent_form.build_parent
        consent_object.assign_attributes(parent_params)
```

app/controllers/parent_interface/consent_forms/edit_controller.rb
```diff
@@ -16,13 +17,30 @@ def show
     end
 
     def update
+      model = @consent_form
```
misaka, line 20: What's the point of this assignment? I looks like it always gets overwritten on line 29.

misaka: Sorry, nvm, I didn't think this method would be so long, didn't realise the `render model` at the end was part of this same method!

app/models/parent.rb
```diff
@@ -45,6 +45,7 @@ class Parent < ApplicationRecord
             length: {
               maximum: 300
             }
+  validates :contact_method_other, presence: true, if: :contact_method_other?
```
misaka, line 48: This isn't confusing at all. 😵  Not your fault, but this line does highlight the unfortunate naming we have of the `contact_method` enum.

misaka (approved):
Quite intricate but looking good. Ship and roll fixes forward, I say. 🚢

app/controllers/parent_interface/consent_forms/edit_controller.rb
benilovj: ah, sorry, this bit gets refactored away to something much simpler in a later commit, so safe to ignore.

app/controllers/parent_interface/consent_forms/edit_controller.rb
benilovj: Ya, this got refactored out as well in a later commit; now the parent is loaded or built in `EditController#set_parent` and just used where it's needed.

app/models/parent.rb
benilovj: mm yeah not ideal.

PR #1415, opened 2024-06-19
PDS Integration

misaka:
This is the simplest possible way of getting a patient from PDS. Wrapped up into a model so we can change the implementation later.

The model itself is also quite simple, whereas the FHIR model may potentially have multiple names (in case the patient has changed their name), etc. We need to better understand how Mavis uses this data before we expand our own model, so consider this a proof-of-concept.

### DIFF
--- a/app/models/pds/patient.rb
+++ b/app/models/pds/patient.rb
@@ -1,0 +1,31 @@
+class PDS::Patient
+  include ActiveModel::Model
+
+  attr_accessor :nhs_number, :given_name, :family_name, :date_of_birth
+
+  class << self
+    def find(nhs_number)
+      response =
+        JSON.parse(
+          Net::HTTP.get(
+            URI(
+              "https://sandbox.api.service.nhs.uk/personal-demographics/FHIR/R4/Patient/#{nhs_number}"
+            ),
+            "X-Request-ID": SecureRandom.uuid
+          )
+        )
+      from_pds_fhir_response(response)
+    end
+
+    private
+
+    def from_pds_fhir_response(response)
+      new(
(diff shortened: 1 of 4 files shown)


app/models/pds/patient.rb
```diff
@@ -0,0 +1,31 @@
+class PDS::Patient
+  include ActiveModel::Model
+
+  attr_accessor :nhs_number, :given_name, :family_name, :date_of_birth
+
+  class << self
+    def find(id)
```
benilovj, line 7:
```suggestion
    def find(nhs_number)
```

misaka: I'm of two minds on this one. On one hand, yes, this is the NHS Number, but in terms of FHIR API, this is an "id". You can see it's actually returned as "id" in the JSON response with no specific mention of "nhs_number", which is why I wasn't too bothered which one it was called here.

Still feel like it's more intuitive to call it "nhs_number" here?

benilovj: > but in terms of FHIR API, this is an "id"

I made the suggestion because upon reading the code, my first thought was – "what is the id actually". My 2p: it's fair enough that it's the ID in the FHIR API, but I think it'd be more expressive to abstract away from that and just call it what it is within our codebase.

misaka: It's hard to gauge after working with it for a while, useful to have someone's initial reaction. 🙏

tvararu: New code reads better 👍

tvararu: New code reads better 👍

tvararu (approved):

PR #1417, opened 2024-06-21
Remove routes flagged by `rails routes --unused`

benilovj:
This will further break offline support, but I suspect that stopped working a long time ago and would need to be revived separately, in any case.


tvararu (approved):
🎉

PR #1422, opened 2024-06-24
Tweak default batch functionality

tvararu:
See commits.

image

### DIFF
--- a/app/views/vaccinations/batch.html.erb
+++ b/app/views/vaccinations/batch.html.erb
@@ -16,15 +16,12 @@
         caption: { text: @session.location.name, size: "l" },
         legend: { size: "l", tag: "h1", text: page_title },
       ) do %>
-    <% @batches.each_with_index do |batch, index| %>
-      <%= f.govuk_radio_button(
-            :id,
-            batch.id,
-            label: {
-              text: "#{batch.name} (expires #{batch.expiry.to_fs(:nhsuk_date)})",
-            },
-            link_errors: index == 0,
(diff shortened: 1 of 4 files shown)


benilovj (approved):

app/views/vaccinations/batch.html.erb
```diff
@@ -16,15 +16,12 @@
...
-            label: {
-              text: "#{batch.name} (expires #{batch.expiry.to_fs(:nhsuk_date)})",
-            },
-            link_errors: index == 0,
-          ) %>
+    <% @batches.each_with_index do |batch, idx| %>
+      <% label = proc do %>
+        <span class="app-u-monospace"><%= batch.name %></span>
+        (expires <%= batch.expiry.to_fs(:nhsuk_date) %>)
+      <% end %>
```
misaka, line 23: Do you need a proc here? I tested and I think this could also work with:

```
      label = begin %>
        <span class="app-u-monospace"><%= batch.name %></span>
        (expires <%= batch.expiry.to_fs(:nhsuk_date) %>)
      <% end %>
```

misaka (approved):

app/views/vaccinations/batch.html.erb
tvararu: Tried it but `rufo` formats it weirdly... it tries to immediately end the `begin` 🤔

misaka: Shame. :shipit:

PR #1423, opened 2024-06-24
Provision production environment

tvararu:
Tweak the documentation and update the SSL certificate.

See commits.


tvararu:
This is ready to review but isn't really ready to merge before I confirm that the non-www domains work in practice.

misaka (approved):

PR #1429, opened 2024-06-25
Update dashboard page to match designs

thomasleese:
This updates the cards on the dashboard page to match the design in the prototype. It also refactors the page titles of a number of sections to use locale strings.

I've also sorted the locale string file to reduce churn and the size of diffs in the future, but if we're not confident on this change I can remove it from this PR as it does make the number of lines changed very big!

### Old

Screenshot 2024-06-25 at 17 22 33

### New

Screenshot 2024-06-25 at 17 22 15


tvararu (approved):
🙌

benilovj (approved):

PR #1431, opened 2024-06-26
Add "or" divider to triage form options

tvararu:
This adds an "or" divider to triage form options.

It also updates the legend styling to match the prototype. It should be bolded on the patient page, and on the update triage outcome page it should not be. For the latter, we need some custom CSS as the `govuk_design_system_formbuilder` fieldset component always specifies a size which sets the font weight to bold and adds a margin.

Also removes some unused code and simplifies the arguments to `AppTriageFormComponent` by removing the `section` and `tab` kwargs, instead passing in the whole URL for the form.

#### Patient page / New triage

Screenshot 2024-06-26 at 09 07 08

#### Updating a triage outcome

image


app/views/manage_consents/triage.html.erb
```diff
@@ -17,11 +17,28 @@
...
+        ) %>
+    <%= f.govuk_radio_button(
+          :status, :delay_vaccination,
+          label: { text: Triage.human_enum_name(:status, :delay_vaccination) },
+        ) %>
+    <%= f.govuk_radio_button(
+          :status, :needs_follow_up,
+          label: { text: Triage.human_enum_name(:status, :needs_follow_up) },
+        ) %>
+  <% end %>
```
thomasleese, line 41: This looks the same to me as the fields in `app_triage_form_component.html.erb`, is there a reason we wouldn't use that component or can we extract this to somewhere shared across both the component and this view?

tvararu: Yup, but I felt like the code looked uglier/harder to understand than without...

Here's something similar to what I tried:

```erb
<%= f.govuk_radio_button(*radio_options(:ready_to_vaccinate)) %>
<%= f.govuk_radio_divider %>
<%= f.govuk_radio_button(*radio_options(: do_not_vaccinate)) %>
<%= f.govuk_radio_button(*radio_options(: delay_vaccination)) %>
<%= f.govuk_radio_button(*radio_options(: needs_follow_up)) %>
```

Where `radio_options` is a helper that returns out `[:status, :ready_to_vaccinate, { label: ... }]`.

However, Ruby is weird about how it handles splatting and keyword args. The code above would error due to passing in 3 arguments when 2 are required. This is because the final argument in my `radio_buttons` helper is a hash, but the radio expects kwargs, which are subtly different. The hash argument needs a double splat `**` but there's no easy way to do that.

So this would work, but it's really ugly:

```erb
<%= f.govuk_radio_button(*radio_options(:ready_to_vaccinate)[0..2], **radio_options(:ready_to_vaccinate).last) %>
```

After 30 minutes spent trying to golf it I decided the duplication isn't so bad yet...

If you can suggest a different way, happy to try it 👍

tvararu: Ah, the reason not to use the component is because the forms are different I guess. I didn't really consider trying to make `AppTriageFormComponent` fit this usecase. I'll look into that now.

tvararu: I've pushed up a commit, I massaged the `AppTriageFormComponent` to make it more flexible to fit this usecase. 💪

thomasleese (approved):
Looks good to me!

misaka (approved):

misaka (approved):

PR #1434, opened 2024-06-26
Enable frozen string literals

thomasleese:
This modifies the Rubocop rules to enforce a frozen string literal comment at the top of every Ruby file, and then runs Rubocop with autocorrect to add one to each file.

Frozen string literals was going to be default in Ruby 3.0, but this didn't happen in the end due to concerns about incompatibility, however I think this is good practice for new projects. An alternative to the comment is using the `--enable=frozen-string-literal` flag, however this applies to third party Gems and the standard library, which may then introduce the compatibility issues.

Aside from not needing to manually call `.freeze` on string literals, this approach introduces a small performance increase (although unlikely to be particularly noticable). However, it does reduce the chance of bugs happening related to mutable string literals.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # Puma can serve each request in a thread from an internal thread pool.
 # The `threads` method setting takes two numbers: a minimum and maximum.
 # Any libraries that use thread pools should be configured to match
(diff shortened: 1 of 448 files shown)


config/puma.rb
```diff
@@ -20,10 +22,10 @@
 end
 
 # Specifies the `environment` that Puma will run in.
-environment ENV.fetch("RAILS_ENV") { "development" }
+environment ENV.fetch("RAILS_ENV", "development")
```
misaka, line 25: ❤️

misaka (approved):
:shipit:

config/puma.rb
thomasleese: Interesting, not sure why `rubocop -A` decided to fix that now 🤔

PR #1435, opened 2024-06-27
Move CIS2 login button to start page

misaka:
This changes the login flow in a couple ways. By putting the Care Identity button on the start page we hide the username/password login from users when in CIS2-mode. That page is still accessible for the time being, though, for dev and testing purposes. The other big change is when a user visits a page that requires authentication, we can no longer just redirect them to the existing login page, since there's no button to use CIS2 there. Also, for security reasons we cannot simply redirect users to the `cis2` omniauth endpoint, as that requires a POST request (see https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284), so the user is redirected to the Start page and needs to click the Care Identity login button there.

The Care Identity login is placed behind a feature flag so it won't get in the way of development.

image


app/views/pages/start.html.erb
```diff
@@ -10,4 +10,17 @@
...
+<%= govuk_button_to(user_cis2_omniauth_authorize_path, class: "app-button--cis2 nhsuk-u-margin-bottom-4") do %>
+  <svg class="nhsuk-logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 80 32" height="27" aria-hidden="true" focusable="false">
+    <path fill="currentColor" d="M80 0v32H0V0h80ZM69 2.2c-5.8 0-11.6 2-11.6 8.8 0 7.4 10.2 5.8 10.2 10 0 2.6-3.4 3-5.6 3-2.2 0-5-.6-6.4-1.4L54 28c2.2.8 5.4 1.4 8 1.4 6.2 0 12.8-1.8 12.8-9 0-7.8-10.2-6.6-10.2-10.2 0-2.2 2.2-2.6 5-2.6 2.6 0 4.4.6 5.8 1.2L77 3.4c-1.8-.8-4.8-1.2-8-1.2ZM16.6 3H7.8L2.2 29h6.6l3.6-18h.2L18 29h8.6l5.6-26h-6.6L22 21h-.2L16.6 3Zm25.2 0h-7.2l-5.2 26h6.8l2.4-11.2h8.2L44.6 29h7L57 3h-7l-2.2 9.8h-8l2-9.8Z"></path>
+  </svg>
+
+  <span>Care Identity</span>
+<% end %>
+
+<p class="nhsuk-body">
+  <a class="nhsuk-link" href="https://manage-care-identities.care-identity-service2.nhs.uk/#/self-service/unlock-authenticator">
```
thomasleese: This goes to a 404 page for me, is the URL incorrect or is it not live yet?

spec/features/login_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  def given_the_cis2_feature_flag_is_disabled
+    Flipper.disable(:cis2)
+  end
+
+  def when_i_go_to_the_service
+    visit "/"
+  end
+
+  def then_i_should_see_the_cis2_login_button
+    expect(page).to have_button "Care Identity"
```
thomasleese:
```suggestion
    expect(page).to have_button "Log in with my Care Identity"
```

Should we check for this to ensure the content from the CSS is coming through as well?

app/controllers/application_controller.rb
```diff
@@ -26,6 +26,17 @@ class UnprocessableEntity < StandardError
 
   private
 
+  def authenticate_user!
+    unless user_signed_in? ||
+             request.path.in?(
+               [new_user_session_path, destroy_user_session_path]
+             )
+      store_location_for(:user, request.fullpath) if request.path != start_path
+      flash[:info] = "You must be logged in to access this page"
```
thomasleese, line 35:
```suggestion
      flash[:info] = "You must be logged in to access this page."
```

Would we put a full stop here?

app/views/pages/start.html.erb
misaka: I'm getting this page, which is an error but not a 404. tbh I'm not sure exactly how this is meant to work, but I thought we could 🥾🥫 until we better understood this stuff.

image

misaka:
@thomasleese Changes made (existing commits redone). Can you double check the 404, and if so can we have a look together?

app/views/pages/start.html.erb
thomasleese: Turns out this is a country thing, looks like that page only works in the UK, it worked when I switched on my VPN. I've attached what it looks like outside the UK.

Screenshot 2024-06-27 at 13 16 47

thomasleese (approved):

spec/features/login_spec.rb
misaka: Ugh. That's right, I did try this, and it did blow up, and I did ¯\\\_(ツ)_/¯ and did move on. _le sigh_

image

I'm not sure this is worth the effort to figure out, but if you know of a way to make this work happy to implement.

tvararu (approved):

app/views/pages/start.html.erb
misaka: Ah, interesting. The dev CIS2 worked when I was out of the country recently, and I was wondering if production was going to be region-restricted. Glad to see it is.

tvararu (approved):

PR #1437, opened 2024-06-27
Don't specify form builder in views

thomasleese:
We've set `GOVUKDesignSystemFormBuilder::FormBuilder` as the default form builder for forms so we don't need to specify this elsewhere.


thomasleese:
https://viewcomponent.org/known_issues.html#forms-dont-use-the-default-formbuilder 🤦🏻

PR #1440, opened 2024-06-28
Standardise date and time formats

thomasleese:
This standardises the date and time formats across the service to ensure that, where possible, we use a consistent format everywhere. It's based on [the standard date and time formats](https://railsdatetimeformats.com/) that is provided by Rails and then slightly modified to match the NHS service manual (in general, this means months are written in full).

See individual commits for more details on the changes. Overall this change has minimal impact on the frontend, but there are a number of places currently where months are written truncated which this PR has changed to write them out in full.

This was inspired from a conversation in https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1439#pullrequestreview-2145769187.

### DIFF
--- a/app/components/app_activity_log_component.html.erb
+++ b/app/components/app_activity_log_component.html.erb
@@ -1,7 +1,7 @@
 <% events_by_day.each do |day, events| %>
   <h2 class="nhsuk-heading-xs nhsuk-u-secondary-text-color
              nhsuk-u-font-weight-normal">
-    <%= day %>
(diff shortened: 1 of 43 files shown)


misaka (approved):
🎉 Nice one, brings some much-needed clarity to our time & date formats. I _did_ kind of like Theo's suggestion that we name the time and dates formats after the format (e.g. `YYYYMMDD`, but there are good reason (communicated here) to stick with standard names. 💯

tvararu (approved):
:shipit:

PR #1441, opened 2024-06-28
Add bin/test script

thomasleese:
This adds a script, similar to bin/lint, which automatically runs all the tests in one go. It's meant to be used for convenience when needing to run all the test, but won't replace CI or the running RSpec directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,12 @@ $ bin/dev
 
 ### Testing
 
+To run all the test:
+
+```bash
+bin/test
+```
+
 To run the Rails tests:
 
 ```bash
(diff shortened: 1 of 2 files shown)


benilovj:
Given that:

1. We don't have any running playwright tests anymore
2. We don't really make many changes to the JS at the moment

we **could** theoretically just run `bundle exec rspec` and leave the other two to run on CI (which would shave a few secs off each run)

thomasleese:
@benilovj Thanks, yeah, we came to the same conclusion in Slack, I'll close this as unneeded.

PR #1444, opened 2024-06-28
Fix spec warning

benilovj:
Fixes the following:

> WARNING: `around(:context)` hooks are not supported and behave like `around(:example). Called from spec/components/app_triage_notes_component_spec.rb:22:in `block (2 levels) in <top (required)>'.


tvararu (approved):

benilovj:
Already fixed by #1443.

PR #1453, opened 2024-07-02
Implement patient table collapsing on mobile

tvararu:
Adds the `.nhsuk-table-responsive` class and the `__heading` span inside each individual cell.

Also adds `.app-button-group` styles that will be used in a future piece of work that adds the attendance buttons.

### After (mobile)

Screenshot 2024-07-02 at 11 11 52


misaka (approved):
It's probably ok, but is there a specific reason to add `app-button-group` now before it's being used?

tvararu:
@misaka I started adding it, realised we don't need it yet, then said let's just keep it in as we'll need it eventually I think

PR #1455, opened 2024-07-02
Mavis consent form: improvements to 'parent details' step

benilovj:
## Before

screencapture-localhost-4000-sessions-1-consents-no-consent-patients-2-consents-66-who-2024-07-02-19_54_16

## After

screencapture-localhost-4000-sessions-1-consents-no-consent-patients-2-consents-65-parent-details-2024-07-02-19_53_42

## Changes

* rename the step from `who` to `parent-details` (`who` can be used in future to pick between existing or new parent)
* allow the `email` to be edited
* reorder the fields for consistency with the parent consent form
* create a new Parent record only when it's necessary, not when the Consent record is created; this will allow reusing an existing Parent record, instead of always creating a new one
* Add a new feature spec: verbal consent from a parent not on the system (all existing verbal consent specs always go with the pre-populated existing contact)


app/controllers/manage_consents_controller.rb
```diff
@@ -40,9 +43,20 @@ def update
...
     when :confirm
       handle_confirm
-    when :who
+    when :parent_details
       model = @consent.draft_parent
-      handle_who
+      handle_parent_details
+
+      # @consent.draft_parent might not be persisted at this point,
+      # so we need to save using
```
tvararu, line 51: I'm not sure I understand this comment 🤔

tvararu (approved):

app/controllers/manage_consents_controller.rb
benilovj: I think I did a not-very-good job of explaining why we can't just rely on `render_wizard` to do the right thing here, like in the other cases:

* if the user is going thru this flow for the first time, `consent.parent_id` is `nil` in the DB and `consent.parent_id` is a new record (set in `set_parent`) so you'd need to call `consent.save` to persist it in addition to saving `parent`
* if the user is going back to edit this page before confirming, `consent.parent_id` is already defined and it would be sufficient to just persist any changes on the `parent` without needing to touch `consent`

Maybe the comment is superfluous?

tvararu: Yeah, I don't think we need the comment, there are some other cases where we can't rely on `render_wizard`

PR #1456, opened 2024-07-04
Check that a user belongs to a team on login.

misaka:
> [!NOTE]
> When authenticating with CIS2, the wording used for choosing a role at an organisation is "role selection". But a "role" is it's own entity independent of organisations when not connected to a user, so this  term is overloaded. In Care Identity Management, the "position" entity joins the role(s) to organisations and is assigned to the user.

Users need to select which role they want to use when they login. Roles are connected to organisations, so Mavis uses the selected role (aka position) to know which team they are a part of. Is the selected organisation doesn't have a team we show them this error page and provide a button they can use to choose a different role.

image

The flow for logging in when the user selects the wrong role looks like this:

```mermaid
flowchart TB;
    subgraph Mavis
        direction TB;
        A[Start Page]
        C[Team not found]
        E[Dashboard]
    end
    subgraph CIS2
        direction TB;
        B[Login]
        D[Select role]
    end
    A -->|Start| B;
    B --> C;
    C -->|Change role| D;
    D --> E;
```


app/controllers/dev_controller.rb
```diff
@@ -46,6 +46,9 @@ def reset
             user:
           )
         end
+
+      Team.find_by(ods_code: "Y51") ||
+        FactoryBot.create(:team, name: "NMEPFIT SAIS Team", ods_code: "Y51")
```
benilovj, line 51: Is it better not to seed this?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -5,8 +5,30 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+    team = Team.find_by(ods_code: nrbac_role["org_code"])
+    if team.nil?
+      @selected_org =
+        raw_info["nhsid_user_orgs"].find do
+          _1["org_code"] == nrbac_role["org_code"]
+        end
+
+      flash[:org_name] = @selected_org["org_name"]
+      flash[:org_code] = @selected_org["org_code"]
+      flash[:has_other_roles] = raw_info["nhsid_nrbac_roles"].length > 1
```
tvararu, line 25: I guess as long as we don't assign to `flash[:info]` or `:success` or `:warning` this is fine... it shouldn't leak to the UI in the flash banner component and it won't persist past the next request.

Should we pass a hash along instead of 3 separate things?

app/controllers/errors_controller.rb
```diff
@@ -22,4 +23,16 @@ def too_many_requests
   def internal_server_error
     render "internal_server_error", status: :internal_server_error
   end
+
+  def team_not_found
```
tvararu, line 27: Should this live somewhere else, in a different controller (probably in the Users namespace)? My intention with `errors_controller` was for it to only handle the fallback/generic 404/422/429/500 scenarios. It feels like this functionality doesn't fit that model.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -5,8 +5,30 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
-    @user =
-      User.find_or_create_user_from_cis2_oidc(request.env["omniauth.auth"])
-    sign_in_and_redirect @user, event: :authentication # this will throw if @user is not activated
+    user_info = request.env["omniauth.auth"]
+    raw_info = user_info["extra"]["raw_info"]
+    selected_role = raw_info["selected_roleid"]
+    nrbac_role =
+      raw_info["nhsid_nrbac_roles"].find do
+        _1["person_roleid"] == selected_role
+      end
```
tvararu, line 14: I feel like the controller action would be skinnier if we extracted this assignment logic, and perhaps the flash-setting bits further down, to private methods... It would introduce some indirection, but would make it easier to see at a high level what the action is doing.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -5,8 +5,30 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+        raw_info["nhsid_user_orgs"].find do
+          _1["org_code"] == nrbac_role["org_code"]
+        end
+
+      flash[:org_name] = @selected_org["org_name"]
+      flash[:org_code] = @selected_org["org_code"]
+      flash[:has_other_roles] = raw_info["nhsid_nrbac_roles"].length > 1
+
+      redirect_to team_not_found_path
+      false
```
tvararu, line 28: I guess this `false` return is needed?

app/views/errors/team_not_found.html.erb
```diff
@@ -0,0 +1,14 @@
+<%=  h1 "Your organisation is not using this service yet" %>
+
+<p class="nhsuk-body">
```
tvararu, line 3:
```suggestion
<p>
```

config/initializers/devise.rb
```diff
@@ -286,30 +286,42 @@
...
+      {
+        setup:
+          lambda do |env|
+            if env.dig("rack.request.form_hash", "change_role")
+              env["omniauth.strategy"].options["scope"] << :changedrole
+              env["omniauth.strategy"].options["prompt"] = :login
+            else
+              env["omniauth.strategy"].options["scope"] << :selectedrole
+            end
+          end,
```
tvararu, line 300: I'd consider moving this to a variable above like `redirect_uri`, I think it'd be slightly cleaner

spec/features/user_authentication_with_cis2_spec.rb
```diff
@@ -59,11 +61,19 @@ def cis2_auth_mock
           ],
           "given_name" => "Nurse",
           "family_name" => "Flo",
+          "nhsid_user_orgs" => [
+            {
+              "org_name" => "LAVENDER TOWN HEALTH",
```
tvararu, line 66: 👍 for Pokemon reference

spec/features/user_authentication_with_cis2_spec.rb
```diff
@@ -97,6 +109,43 @@ def cis2_auth_mock
...
+
+      given_the_cis2_feature_flag_is_enabled
+      when_i_go_to_the_start_page
+      then_i_should_see_the_cis2_login_button
+
+      when_i_click_the_cis2_login_button
+      then_i_see_the_team_not_found_error
+      and_there_is_no_change_role_button
+    end
+  end
```
tvararu, line 147: I don't think we should have 3 scenarios in one feature spec. I think we should split these up into separate tests... It's probably more effort rewriting some of the steps but for consistency we try to stick to 1 scenario per file in all except for one case

app/controllers/users/omniauth_callbacks_controller.rb
misaka: Sure, a hash can be tidier.

app/controllers/errors_controller.rb
misaka: Ya, it's not a perfect fit. I think it could fit well in `Users::SessionsController` (fwiw I also considered and discarded putting it into `PagesController`)

misaka: Oh ya, this is why I didn't do it, and decided to 🥾🥫 it into `ErrorsController`

image

app/controllers/users/omniauth_callbacks_controller.rb
misaka: Yup, will tidy up.

app/controllers/users/omniauth_callbacks_controller.rb
misaka: Er, I was convinced it was, but I must've been smoking a red herring or something. Whatever error I was getting before isn't showing now. ✔️

app/controllers/errors_controller.rb
misaka: Found a way. I moved this to `users/accounts_controller` ... it could also have gone to `users/sessions_controller` but that requires a separate `devise_scope` as the above error indicates, and I think the accounts controller is as good a place as sessions.

spec/features/user_authentication_with_cis2_spec.rb
misaka: I've gone through and split this up into 3 files, each with a couple of scenarios. I'm not sure being really strict about only one scenario per file is right ... it becomes quite tedious to name the files and thinking about how they fit together, imo. Now that I've split this up, I actually think that having all the CIS2 auth stuff in one file was alright, but I think this is a reasonable middle-ground too.

Let me know what you think.

app/controllers/users/omniauth_callbacks_controller.rb
misaka: I've done a bit of work here including adding a concern and closing up another edge-case where someone could visit pages without having their team signed up (albeit without seeing any data). Please re-review.

spec/fixtures/cis2_auth_info.rb
tvararu, line 1: Nice

tvararu (approved):

spec/features/user_authentication_with_cis2_spec.rb
tvararu: Yeah, I think this feels a bit better... let's see how it works out as we iterate

app/controllers/dev_controller.rb
misaka: The issue is that the dev controller's `/reset` truncates the `teams` table. The example campaign creates a team so it's not  as simple as changing `/reset` to leave it, I think we should 🥾🥫 this and worry about it later.

PR #1457, opened 2024-07-04
Add NIVS reporting test data and ImmunisationImport model

tvararu:
This will be used to validate and process NIVS uploads by parsing each row and creating necessary vaccination, patient, location, and other records.

This is only the first PR that adds the initial model, a basic test, and serves as a starting point from which to keep building the rest of the functionality. It doesn't integrate it in any way with the rest of the app, the model is tested in isolation.

See commits.


app/models/immunisation_import.rb
```diff
@@ -0,0 +1,74 @@
...
+      {
+        "Left Thigh" => :left_thigh,
+        "Right Thigh" => :right_thigh,
+        "Left Upper Arm" => :left_arm_upper_position,
+        "Right Upper Arm" => :right_arm_upper_position,
+        "Left Buttock" => :left_arm, # TEMPORARY
+        "Right Buttock" => :left_arm, # TEMPORARY
+        "Nasal" => :left_arm # TEMPORARY
+      }[
+        @row["ANATOMICAL_SITE"]
```
benilovj, line 66: It should be more robust to be blanket case-insensitive...

thomasleese (approved):
Looks good to me!

PR #1472, opened 2024-07-09
Add initial immunisation import flow

thomasleese:
This adds a controller and some views to support a basic immunisation import upload flow, where the user can upload a CSV file and it will be stored in the `ImmunisationImport` model.

The next steps after this will be to process the CSV file, provide feedback to the user on success/failures and add error validation against invalid files.

This is a dev-lead design based on the pilot cohort upload as we don't have proper designs for this yet, but we can make the necessary changes when those designs come through.

## Screenshots

Screenshot 2024-07-09 at 15 16 26

Screenshot 2024-07-09 at 15 20 49

### DIFF
--- a/app/controllers/immunisation_imports_controller.rb
+++ b/app/controllers/immunisation_imports_controller.rb
@@ -1,9 +1,17 @@
 # frozen_string_literal: true
 
 class ImmunisationImportsController < ApplicationController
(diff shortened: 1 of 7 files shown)


tvararu (approved):
Clean 👍

PR #1473, opened 2024-07-09
Capture vaccine on vaccination record, use it to remove hard-coded logic in the UI

benilovj:
## Background

Prior to this change, there's lots of implicit domain logic (e.g. relating to delivery sites and delivery methods) that's hard-coded into views and helpers. Lots of those assumptions break as soon as we introduce anything other than HPV, which we're currently doing as part of the importer work. For me, the key part of this PR is the change to the `Vaccine` class and being able to articulate the domain logic where it belongs.

Take, for instance, the changes to `AppVaccinateFormComponent`. That form implicitly assumes that the user is recording an injection (since the defaults are left arm and right arm), and at that point in its lifecycle, the `VaccinationRecord` hasn't even been persisted yet.

This PR just makes that assumption explicit by pre-determining the vaccine (that the UI already assumes anyway). That makes it possible to defer at least some of the logic to the `Vaccine` class. I get that the UI might be different for recording different vaccines, but grouping logic this way will make it much easier to do sensible validations (e.g. catch invalid records like nasal flu administered into the buttocks as we build the reporting import). 

There's also a bugfix here, whereby you can only pick a batch for the vaccine that you're recording, not just all batches for all the vaccines in the campaign.

## To dos

- [x] `VaccinationRecord`: validate that the `vaccine` matches `batch.vaccine` for non-draft records


misaka:
@benilovj I don't quite understand why `vaccine` has to be a separate association on `VaccinationRecord`. The only place I could see where it mattered was the final commit where the vaccine is set on the draft vaccination_record, and the batch is selected from there, but that seems like a pretty drastic change just to get that one benefit.

Is it really necessary? If it is, might be worth adding as a comment to the first commit.

benilovj:
> @benilovj I don't quite understand why `vaccine` has to be a separate association on `VaccinationRecord`. The only place I could see where it mattered was the final commit where the vaccine is set on the draft vaccination_record, and the batch is selected from there, but that seems like a pretty drastic change just to get that one benefit.
> 
> Is it really necessary? If it is, might be worth adding as a comment to the first commit.

Prior to this change, there's lots of implicit domain logic (e.g. relating to delivery sites and delivery methods) that's hard-coded into views and helpers. Lots of those assumptions break as soon as we introduce anything other than HPV, which we're currently doing as part of the importer work. For me, the key part of this PR is the change to the `Vaccine` class and being able to articulate the domain logic where it belongs.

Take, for instance, the changes to `AppVaccinateFormComponent`. That form implicitly assumes that the user is recording an injection (since the defaults are left arm and right arm), and at that point in its lifecycle, the `VaccinationRecord` hasn't even been persisted yet. 

This PR just makes that assumption explicit by pre-determining the vaccine (that the UI already assumes anyway). That makes it possible to defer at least *some* of the logic to the `Vaccine` class. I get that the UI might be different in different cases, but grouping logic this way will make it much easier to do sensible validations (e.g. catch invalid records like nasal flu administered into the buttocks). There's also a bugfix here, whereby you can only pick a batch for the vaccine that you're recording, not just all batches for all the vaccines in the campaign.

tvararu (approved):

PR #1474, opened 2024-07-10
Add validations to ImmunisationImport

tvararu:
This adds a few of the basic validations that we'll need for the NIVS upload journey.


app/models/immunisation_import.rb
```diff
@@ -21,16 +21,32 @@
...
+    return if invalid?
+
+    self.data ||= CSV.parse(csv, headers: true, skip_blanks: true)
+  rescue CSV::MalformedCSVError
+    self.csv_is_malformed = true
+  ensure
+    csv.close if csv.respond_to?(:close)
+  end
+
   def process!(patient_session:)
```
thomasleese, line 49: Should this `raise` or something similar if we haven't previously called `load_data!`, i.e. `data` is `nil`?

spec/models/immunisation_import_spec.rb
```diff
@@ -30,8 +30,45 @@
 
   it { should validate_presence_of(:csv) }
 
+  describe "#load_data!" do
+    describe "with malformed CSV" do
+      let(:file) { "malformed.csv" }
+
+      it "is invalid" do
+        subject.load_data!
```
thomasleese, line 38: This could go in a `before` block since every test does it.

thomasleese (approved):
Looks good to me, just a couple of comments, but not blocking.

app/models/immunisation_import.rb
tvararu: What about calling `load_data!` if `data.nil?`?

thomasleese: Yeah that would work too, although it would still then need to check the data was loaded correctly.

PR #1475, opened 2024-07-10
Check user's role on cis2 login

misaka:
Users must have certain roles to be able to access the system. If they do not have an appropriate role they will get the page below. The list of valid roles is still naively implemented, we will refine that as we develop authorisation flows to match the designs.

image

### DIFF
--- a/app/views/users/accounts/role_not_found.html.erb
+++ b/app/views/users/accounts/role_not_found.html.erb
@@ -1,0 +1,20 @@
+<%= h1 "You do not have permission to use this service" %>
+
+<% if @cis2_info[:has_other_roles] %>
+  <%= govuk_button_to "Change role", user_cis2_omniauth_authorize_path, params: { change_role: true } %>
+<% end %>
+
+<h2 class="nhsuk-heading-m">Contact your registration authority</h2>
+
+<p>
+  If you think you should have permission to use this service, contact your
+  registration authority, Acme Community Hospital NHS Foundation Trust (ACM).
+  Ask for one of the following roles to be added to your Care Identity:
+</p>
+
+<%= govuk_list([
+      "Nurse Access Role (R8001)",
+      "Medical Secretary Access Role (R8006)",
+      "Admin/Clinical Support Access Role (R8008)",
+      "Receptionist Access Role (R8009)",
+    ], type: "bullet") %>
--- a/app/views/users/accounts/role_not_found.html.erb
+++ b/app/views/users/accounts/role_not_found.html.erb
@@ -8,7 +8,15 @@
 
 <p>
   If you think you should have permission to use this service, contact your
-  registration authority, Acme Community Hospital NHS Foundation Trust (ACM).
+  registration authority:
+</p>
+
+<%= govuk_inset_text do %>
+  <%= @cis2_info[:selected_org][:name] %><br />
+  (<%= @cis2_info[:selected_org][:code] %>)
+<% end %>
+
+<p>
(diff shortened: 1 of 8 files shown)


app/views/users/accounts/role_not_found.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= h1 "You do not have permission to use this service" %>
+
+<% if @cis2_info[:has_other_roles] %>
+  <%= govuk_button_to "Change role", user_cis2_omniauth_authorize_path, params: { change_role: true } %>
+<% end %>
+
+<h2 class="nhsuk-heading-m">Contact your registration authority</h2>
+
+<p class="nhsuk-body">
```
tvararu, line 9:
```suggestion
<p>
```

app/views/users/accounts/role_not_found.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  registration authority, Acme Community Hospital NHS Foundation Trust (ACM).
+  Ask for one of the following roles to be added to your Care Identity:
+</p>
+
+<ul class="nhsuk-list nhsuk-list--bullet">
+  <li>Nurse Access Role (R8001)</li>
+  <li>Medical Secretary Access Role (R8006)</li>
+  <li>Admin/Clinical Support Access Role (R8008)</li>
+  <li>Receptionist Access Role (R8009)</li>
+</ul>
```
tvararu, line 20: I recently learned that there is a `govuk-components` helper for `govuk_list`: https://govuk-components.netlify.app/helpers/list/

```suggestion
<%= govuk_list(["Nurse Access Role (R8001)",
                "Medical Secretary Access Role (R8006)",
                "Admin/Clinical Support Access Role (R8008)",
                "Receptionist Access Role (R8009)"], type: :bullet) %>
```

Non-blocking suggestion, code is nearly the same with/without.

spec/fixtures/cis2_auth_info.rb
tvararu, line 1: I feel like we could GPT this file to be:

```rb
CIS2_AUTH_INFO = {
  provider: :cis2,
  uid: "123456789012",
  info: {
    name: "Nurse Test",
    ...
  }
}.with_indifferent_access.freeze
```

Which is slightly cleaner looking, without the extra "" and =>

Also does it need the Ruby shebang on line 1?

tvararu (approved):

app/views/users/accounts/role_not_found.html.erb
misaka: Ya, you commented this on the last PR, but this is copied from the design. Is it extraneous there too?

spec/fixtures/cis2_auth_info.rb
misaka: > Which is slightly cleaner looking, without the extra "" and =>

I was tempted to do this, but I also wanted this file to provide a kind of reference to what data is provided by CIS2. The counter-argument is that the CIS2 docs should do that, but iirc they don't provide a very good example of what the whole response looks like.

> Also does it need the Ruby shebang on line 1?

Urp. No, that's just emacs being "helpful", missed that.

app/views/users/accounts/role_not_found.html.erb
tvararu: Yup, it's a CSS class that isn't defined and has no rules. cc @paulrobertlloyd

misaka:
Argh. Looks like an intermittent error. Replicating it locally with the same seed.

tvararu (approved):

PR #1476, opened 2024-07-11
Add immunisation import row validation

thomasleese:
This adds row-level validation to the `ImmunisationImport` model, allowing us to check the CSV data that comes in before performing the import. This doesn't present the errors back to the user yet, but that will be the next steps after this.


spec/models/immunisation_import_spec.rb
```diff
@@ -21,7 +21,7 @@
 require "rails_helper"
 
 describe ImmunisationImport do
-  subject { described_class.create!(csv:, user:) }
+  subject(:immunisation_import) { described_class.create!(csv:, user:) }
```
tvararu, line 24: Don't think I've seen this before, cool

spec/fixtures/immunisation_import/invalid_rows.csv
```diff
@@ -0,0 +1,2 @@
+ORGANISATION_CODE,SCHOOL_URN,SCHOOL_NAME,NHS_NUMBER,PERSON_FORENAME,PERSON_SURNAME,PERSON_DOB,PERSON_GENDER_CODE,PERSON_POSTCODE,VACCINATED,DATE_OF_VACCINATION,VACCINE_GIVEN,BATCH_NUMBER,BATCH_EXPIRY_DATE,ANATOMICAL_SITE,PERFORMING_PROFESSIONAL_FORENAME,PERFORMING_PROFESSIONAL_SURNAME,REASON_NOT_VACCINATED,CONSENT_TYPE,LOCAL_PATIENT_ID,LOCAL_PATIENT_ID_URI
+R1L,120026,shaftesbury junior school ,7420180008,Chyna,Pickle,20120912,Not Specified,LE3 2DA,Yes,20240514,AstraZeneca Fluenz Tetra LAIV,123013325,20220730,Unknown,Vaccinator1,Name1,,Parental Consent,LocalPatient1,www.LocalPatient1
```
tvararu, line 2: It's not very clear from glancing at this file why the row is invalid

Non-blocking, but perhaps we could make it look more `OBVIOUSLY BROKEN,`... 🤔

tvararu (approved):

spec/models/immunisation_import_spec.rb
thomasleese: Yeah I like it to make them more explicit while allowing the use the short-hand syntax `it { should }`.

spec/fixtures/immunisation_import/invalid_rows.csv
thomasleese: Yeah I thought about this too, I figure this fixture will get more and more obviously broken as we add more validation, but I'll change it to make clearer here too.

PR #1478, opened 2024-07-11
Associate VaccinationRecords to ImmunisationImport 

tvararu:
This adds ActiveRecord associations to and from ImmunisationImport to the VaccinationRecords that are created from one. This is useful for identifying how certain records were created and will most likely be used in the upload UI.


app/models/immunisation_import.rb
```diff
@@ -24,6 +24,9 @@ class ImmunisationImport < ApplicationRecord
   attr_accessor :csv_is_malformed, :data, :rows
 
   belongs_to :user
+  has_many :vaccination_records,
+           dependent: :nullify,
```
thomasleese, line 28: Do we definitely want this? I imagine if we tried to delete an immunisation import that had created some vaccination records we'd want to stop it from being deleted to keep a record of it?

tvararu: Yeah, I thought the same, but thought adding something to prevent deletion is more complex/requires some sort of hook

However, I just discovered `:restrict_with_exception` which seems like a better fit here: https://guides.rubyonrails.org/association_basics.html#options-for-has-one-dependent

thomasleese (approved):

PR #1482, opened 2024-07-12
Import organisation code from CSV file

thomasleese:
This ensures that we import the `ORGANISATION_CODE` from the CSV files, and matches it against the ODS code of the current signed in SAIS team. See individual commits for more details.


app/models/immunisation_import.rb
```diff
@@ -77,7 +77,14 @@ class Row
     validates :administered, inclusion: [true, false]
     validates :delivery_method, presence: true, if: :administered
     validates :delivery_site, presence: true, if: :administered
-    validates :organisation_code, presence: true, length: { maximum: 5 }
+    validates :organisation_code,
+              presence: true,
+              length: {
+                maximum: 5
+              },
+              inclusion: {
```
misaka, line 85: Why `inclusion` and not `comparison` with `equal_to`? Do we expect to have multiple valid ODS codes at some point?

thomasleese: Ah nope, that's me not noticing `equal_to` when I read the documentation the first time!

misaka (approved):

tvararu (approved):

PR #1483, opened 2024-07-12
Make locations independent of teams

tvararu:
Locations were designed to belong to a specific team during the pilot. We now think making locations global is the better choice, as it improves data consistency.

Patients will be separated from Locations in a follow-up PR.


app/models/cohort_list_row.rb
```diff
@@ -125,7 +124,7 @@ def parent_relationship_hash
   end
 
   def school_urn_is_valid
-    team.locations.find_by!(urn: school_urn)
+    Location.find_by!(urn: school_urn)
```
tvararu, line 127: Theoretically this should use `policy_scope` but it's inside a model, not a controller, so we don't have the pundit helper.

But even if we did that, the Location policy currently has no constraints on it, so it's equivalent to this.

Also, this is part of the pilot cohort upload code, which likely won't be re-used and is kept around for posterity.

misaka (approved):
Ya, LGTM. 🚢

PR #1484, opened 2024-07-15
Use db:seed:replant in reset script

tvararu:
`replant` was added in Rails 6 and seems like it does what we want in this case:

```sh
bin/rails db:seed:replant
    Truncate tables of each database for current environment and load the seeds
```

This also aligns the seeds to populate the `Y51` team.

Notable differences:

- We don't reset the identity counter anymore, so IDs will continue incrementing after resets from where they left off. This is fine; we used to rely on predictable IDs for the Playwright tests, but it isn't necessary anymore.
- We opted out of resetting feature flags before, which made local development easier on new features that relied on us keeping a feature flag around. In that situation, I think we could include the feature flags into `seeds` as well. `seeds` are already configured not to run in prod-like environments, so this should be safe.

### DIFF
--- a/app/controllers/dev_controller.rb
+++ b/app/controllers/dev_controller.rb
@@ -2,54 +2,16 @@
 
 class DevController < ApplicationController
   skip_before_action :authenticate_user!
+  skip_before_action :store_user_location!
   skip_after_action :verify_policy_scoped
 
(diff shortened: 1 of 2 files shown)


thomasleese (approved):
Nice improvement!

PR #1490, opened 2024-07-16
Create locations when importing NIVS data

tvararu:
We need to provision a location so that we can create a session, patient, patient_session, and finally to associate the vaccination records with that patient session.

This also makes URNs unique in the database and adds a model validation for it.

Location importing errors are not yet handled, that can come in a later PR.


app/models/immunisation_import.rb
```diff
@@ -68,6 +68,13 @@ def process!(patient_session:)
     parse_rows! if rows.nil?
     return if invalid?
 
+    rows
+      .map(&:to_location)
+      .uniq(&:urn)
+      .reject(&:invalid?)
+      .map(&:attributes)
+      .tap { Location.create! _1 }
```
tvararu, line 76: Just realised we should set `imported_from` and add the field to locations.

app/models/immunisation_import.rb
```diff
@@ -68,6 +68,13 @@ def process!(patient_session:)
...
+
```
misaka, line 77: If I read this right, the locations are created separately from the vaccination records, i.e. all the locations are created first and then all the vaccination records. This feels a little counter-intuitive to me, although maybe it's fine really, except that if one of the vaccination records can't be created then I think the locations and all the vaccination records up until that one stay. So a couple questions:

- Should we wrap this up as a transaction?
- If not, are we happy to have all the locations created? Thinking of a case where the vaccination record causes an error, maybe because it's completely wrong, then there's a chance the associated location shouldn't have been created?
- Do we have a sense of how many different locations we would expect to see in an uploaded file? This might be moot if it should only ever be one, but then, mistakes do happen and maybe we should check that.

app/models/immunisation_import.rb
misaka: Are we waiting for this to be done?

tvararu: Sorted now

app/models/immunisation_import.rb
tvararu: You're right, wrapped in a transaction. That's the simplest solution for now. We should revisit this and make the validation/error recovery more in-depth as necessary

app/models/immunisation_import.rb
```diff
@@ -74,6 +77,7 @@ def process!(patient_session:)
         .uniq(&:urn)
         .reject(&:invalid?)
         .map(&:attributes)
+        .map { _1.merge(imported_from: self) }
```
misaka, line 80: Why not create locations more like the vaccination_records below? That seems less surprising to me than instantiating a `Location` object, converting it to attributes, and then creating a new `Location` object from that.

tvararu: Then again creating a Location object only to throw it away and do it again later probably isn't fast either so anyway, operational reasoning...

I can switch to `.map(&:save!)` or do assignments more in line with the way we do the records.

misaka: Just had a look at how `create!` this is implemented, and I'm not sure there's any magic here that speeds it up, especially since it uses `save!` instead of `create!` when the input isn't an array:

```ruby
      def create!(attributes = nil, &block)
        if attributes.is_a?(Array)
          attributes.collect { |attr| create!(attr, &block) }
        else
          object = new(attributes, &block)
          object.save!
          object
        end
      end
```

https://github.com/rails/rails/blob/36c1591bcb5e0ee3084759c7f42a706fe5bb7ca7/activerecord/lib/active_record/persistence.rb#L50

misaka (approved):
🚀

PR #1494, opened 2024-07-17
Create patient records when processing import

tvararu:
This creates patients that are associated to the previously created locations.


spec/models/immunisation_import_spec.rb
```diff
@@ -80,13 +80,15 @@
...
+    it "creates locations, patients, and vaccination records" do
       # TEMPORARY: Pass in a dummy patient session. We will iterate this out.
       patient_session = create(:patient_session, user:)
       expect { immunisation_import.process!(patient_session:) }.to change(
         immunisation_import.vaccination_records,
         :count
-      ).by(11).and change(immunisation_import.locations, :count).by(4)
+      ).by(11).and change(immunisation_import.locations, :count).by(
+              4
+            ).and change(immunisation_import.patients, :count).by(11)
```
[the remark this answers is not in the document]
tvararu: Okay, found that `# stree-ignore` makes Prettier ignore the next line! https://github.com/prettier/plugin-ruby#ignoring-code

app/models/immunisation_import.rb
```diff
@@ -116,9 +134,18 @@ def initialize(data:, team:)
...
-      Location.new(
-        name: @data["SCHOOL_NAME"].strip,
-        urn: @data["SCHOOL_URN"].strip
+      Location.new(name: school_name, urn: school_urn)
+    end
+
+    def to_patient
+      return unless valid?
+
+      Patient.new(
```
thomasleese, line 143: Do we want to do a "update or create" here?

Currently this won't work if we import the same patient twice across two different files as the NHS number unique constraint will fail.

app/models/immunisation_import.rb
```diff
@@ -175,6 +202,30 @@ def recorded_at
...
+    def patient_date_of_birth
+      @data["PERSON_DOB"]
+    end
+
+    def patient_nhs_number
+      @data["NHS_NUMBER"]
+    end
+
+    def school_name
+      @data["SCHOOL_NAME"]&.strip
```
thomasleese, line 222: Should we be calling `strip` on all string values if we need it here?

tvararu: Possibly... I'm mainly doing it here in an attempt to clean up the data as the example CSV contains spaces

Screenshot 2024-07-18 at 09 19 14

thomasleese: If random whitespace can appear in the examples, I'd be tempted to apply it to all our string fields just in case.

tvararu: Yeah, I'll push something up

app/models/immunisation_import.rb
tvararu: Added a `.reject(&:invalid?)` instead which is consistent to how we handle Locations

We could instead branch here and return `nil` if the record exists and `.compact` instead back in the `process!` method, I think either approach is fine

thomasleese: Does this introduce a risk that we'll ignore these records for a different validation error (to uniqueness), and that'll then get swallowed and we won't know why the import fails?

app/models/immunisation_import.rb
```diff
@@ -175,6 +203,30 @@ def recorded_at
...
+      @data["PERSON_FORENAME"]
+    end
+
+    def patient_last_name
+      @data["PERSON_SURNAME"]
+    end
+
+    def patient_date_of_birth
+      @data["PERSON_DOB"]
+    end
```
thomasleese, line 216: Should we add some validation on this to ensure this is in the right format?

app/models/immunisation_import.rb
```diff
@@ -167,12 +200,36 @@ def delivery_method
...
+      @data["PERSON_SURNAME"]&.strip
+    end
+
+    def patient_date_of_birth
+      @data["PERSON_DOB"]&.strip
+    end
+
+    def patient_nhs_number
+      @data["NHS_NUMBER"]&.strip
+    end
```
thomasleese, line 225: Can we guarantee that this will contain no superfluous spaces?

I notice [we remove them in the Patient model](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/55e53ac325fc1620dfc4ea1bd531be069f37cd8f/app/models/patient.rb#L75) but that only happens at the point of saving, so here we could end up in a situation where we try and create multiple patients with the same NHS number if one has some spaces in and the other doesn't.

tvararu: Yeah, I thought it would encounter the validation at the point of saving and stop it there.

I've replaced the `.strip` with the gsub call instead 👍

thomasleese (approved):

PR #1495, opened 2024-07-17
Create Session and PatientSession records during import

tvararu:
This allows us to finally replace the stubbed patient_session.

Reviewing the individual commits might be easier.

### DIFF
--- a/app/models/immunisation_import.rb
+++ b/app/models/immunisation_import.rb
@@ -82,6 +82,7 @@ class ImmunisationImport < ApplicationRecord
     ActiveRecord::Base.transaction do
       rows
         .map(&:to_location)
+        .reject(&:persisted?)
         .uniq(&:urn)
         .reject(&:invalid?)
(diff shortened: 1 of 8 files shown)


app/models/immunisation_import.rb
```diff
@@ -90,6 +92,7 @@ def process!(patient_session:)
 
       rows
         .map(&:to_patient)
+        .reject(&:persisted?)
```
thomasleese, line 95: If we're rejecting the ones that are already persisted does that mean we're intentionally not using the data in the import to update existing records with more recent data?

tvararu: Yup, I think that's the right assumption for now...

thomasleese: Cool, yeah, happy with that I just wanted to check it was intentional.

thomasleese (approved):

PR #1499, opened 2024-07-18
Update immunisation import designs to match prototype

thomasleese:
This makes a number of changes to the pages related to immunisation imports to match the prototype. See individual commits for more details. It's a relatively large PR, and I could extract some of the changes to separate PRs if that's preferred.

As we don't currently process the import and generate vaccination records, there will always be 0 records shown, but once we process the import we can update the tests to ensure the records come through.

## Screenshots

Screenshot 2024-07-18 at 14 20 44
Screenshot 2024-07-18 at 14 20 40
Screenshot 2024-07-18 at 14 20 36
Screenshot 2024-07-18 at 14 20 30

### DIFF
--- a/app/controllers/immunisation_imports_controller.rb
+++ b/app/controllers/immunisation_imports_controller.rb
@@ -6,6 +6,7 @@ class ImmunisationImportsController < ApplicationController
   layout "two_thirds", only: :new
 
   def index
+    @immunisation_imports = @campaign.immunisation_imports.order(:created_at)
   end
 
   def new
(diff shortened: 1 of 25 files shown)


app/controllers/immunisation_imports_controller.rb
```diff
@@ -3,8 +3,14 @@
...
   before_action :set_campaign
 
+  def index
+    @immunisation_imports = @campaign.immunisation_imports.order(:created_at)
+  end
+
   def new
     @immunisation_import = ImmunisationImport.new
+
+    render layout: "two_thirds"
```
tvararu, line 13: `layout :two_thirds, only: %i[new]` is the more common pattern I think in the rest of our code

Fine with this though

tvararu (commented):
Nice work 👍 I skimmed through the HTML because I assume we're not doing anything too fancy there other than following the designs

tvararu (approved):
Nice work 👍 I skimmed through the HTML because I assume we're not doing anything too fancy there other than following the designs

app/controllers/immunisation_imports_controller.rb
thomasleese: I'll update this to match the existing code.

PR #1501, opened 2024-07-19
Import postcodes

thomasleese:
This ensures that postcodes are imported from the CSV immunisation import and then displayed as part of the patient records.


misaka (approved):
:shipit:

PR #1524, opened 2024-07-23
Fix date formats and add recorded_date to dps export

misaka:


spec/models/dps_export_row_spec.rb
```diff
@@ -26,7 +26,9 @@
     end
 
     it "has person_dob" do
-      expect(array[3]).to eq vaccination_record.patient.date_of_birth
+      expect(array[3]).to eq vaccination_record.patient.date_of_birth.strftime(
```
benilovj, line 29: I wonder whether it's better to set up specific values (timestamps etc) and assert on calculated values, rather than the implementation? Shouldn't be much more work and gives an example of what the output is like...

misaka: Done

tvararu (approved):

PR #1542, opened 2024-07-25
Add dose sequence column

thomasleese:
This adds a new column to vaccination records that will capture the dose sequence. This is required for the DPS export where we need to include the dose sequence.

I've decided to make the field non-optional for data integrity, although it does mean that we have to (until an interface is built) assume it's always a value of 1. For the NIVS import a follow up PR will be raised to correctly get this value from the CSV file. For the interface [we've got a design ticket in the backlog](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1542).

I've decided to do this explicitly in the code rather than with a default value to ensure that we aren't accidentally creating vaccination records elsewhere in the system and not recording the dose sequence.

An alternative to this could be to make this field `null`, but it introduces a number of data integrity risk:

- it would be possible to create and store vaccination records in the database without a dose sequence
- for Flu (or any vaccine where we only have one dose sequence) we would want to ensure we're not storing a combination of `null` and `1` values


thomasleese:
After discussing with @benilovj we think this will need to be `nullable` in the database (to handle draft records) so I'll amend the PR (temporarily moving it back to draft).

thomasleese:
> Does the fact that the tests didn't catch this mean we're missing coverage? Or do we not have the functionality that creates a vaccination record with a null dose sequence yet?

We don't have the functionality to record this yet in the service. There [is a design ticket in the backlog](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1542) although this [has now been added to the prototype](https://vaccsservice.slack.com/archives/C055LFS48RM/p1721927039512209).

@benilovj It looks like the design for the prototype assumes a dose sequence of 1, unless the user changes it before saving the vaccination record, or when editing an existing record. Therefore, in the end I think it is safe for this field to be `not nullable` - what do you think?

tvararu (approved):

benilovj:
@thomasleese 

> Therefore, in the end I think it is safe for this field to be not nullable - what do you think?

Yes I agree

PR #1550, opened 2024-07-26
Add vaccines initializer

tvararu:
This is a translation to `.yml` of the vaccine configuration table in the Mavis reporting spec.

Immunisation import rows contain a reference to a vaccine name which can roughly be constructed or linked to one of these records.

By adding this to the application, we can use it to:

- Validate immunisation import rows and map them to the canonical product codes
- Present a list of vaccine options to choose from when creating a campaign
- Improve our factories and seed data with more realistic vaccines

The vaccines are keyed by product_code which is also reflected in the values of each, which could come in handy.

### Example

```bash
$ rails runner 'pp VACCINES'
{27114211000001105=>
  {"name"=>"Fluenz Tetra - LAIV",
   "programme_type"=>"Flu",
   "methods"=>["nasal"],
   "brand"=>"AstraZeneca",
   "dose"=>0.2,
   "product_code"=>27114211000001105,
   "product_term"=>
    "Fluenz Tetra vaccine nasal suspension 0.2ml unit dose (AstraZeneca UK Ltd) (product)",
   "procedure_code"=>822851000000102,
   "procedure_term"=>"Seasonal influenza vaccination (procedure)"},
 12238911000001100=>
  {"name"=>"Cervarix",
   "programme_type"=>"HPV",
   "methods"=>["intramuscular", "subcutaneous"],
   "brand"=>"GlaxoSmithKline",
   "dose"=>0.5,
   "product_code"=>12238911000001100,...
```

### DIFF
--- a/config/initializers/vaccines.rb
+++ b/config/initializers/vaccines.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+VACCINES = YAML.load_file(Rails.root.join("config/vaccines.yml"))
--- a/config/vaccines.yml
+++ b/config/vaccines.yml
@@ -1,0 +1,157 @@
+27114211000001105:
+  name: Fluenz Tetra - LAIV
+  programme_type: Flu
+  methods:
+    - nasal
+  brand: AstraZeneca
+  dose: 0.2
+  product_code: 27114211000001105
+  product_term: Fluenz Tetra vaccine nasal suspension 0.2ml unit dose
+    (AstraZeneca UK Ltd) (product)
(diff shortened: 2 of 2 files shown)


thomasleese:
This looks good, I can see how this would be useful.

I'm wondering though if this is effectively a seeds file for the vaccine table? Should we instead be populating the table rather than storing this separately?

I'm wondering how we'll handle changes when we've got data stored in the database and in this YAML file?

thomasleese:
Replaced by #1570

PR #1551, opened 2024-07-28
Import dose sequence from HPV files

thomasleese:
For HPV, it's valid to have a dose sequence and therefore we need to ensure that it's imported correctly. This is an optional field, as Flu files won't contain a dose sequence.

To determine whether we need to import the dose sequence or not we use the maximum dose sequence on the vaccine.

This depends on the vaccine being imported, which hasn't yet been implemented.

### DIFF
--- a/app/models/immunisation_import_row.rb
+++ b/app/models/immunisation_import_row.rb
@@ -6,6 +6,12 @@ class ImmunisationImportRow
   validates :administered, inclusion: [true, false]
   validates :delivery_method, presence: true, if: :administered
   validates :delivery_site, presence: true, if: :administered
+  validates :dose_sequence,
+            presence: true,
+            comparison: {
+              greater_than_or_equal_to: 1,
+              less_than_or_equal_to: :maximum_dose_sequence
+            }
   validates :organisation_code,
             presence: true,
             length: {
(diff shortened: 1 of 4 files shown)


app/models/immunisation_import_row.rb
```diff
@@ -140,6 +146,16 @@ def delivery_method
     end
   end
 
+  def dose_sequence
+    return 1 if vaccine.maximum_dose_sequence == 1
```
misaka, line 150: A part of me is feeling paranoid over this bit. It means that there's a codepath where we could overwrite what is in the import file, and I don't think we should ever do that. Sure, theoretically we shouldn't ever hit this since we should only get the `DOSE_SEQUENCE` field when `maximum_dose_sequence` is > 1, but why even allow for the possibility? Is there a benefit to doing the check like this?

I can think of two alternatives that would make me less paranoid:
1. `return 1 if @data.key?("DOSE_SEQUENCE") && vaccine.maximum_dose_sequence == 1` ... a bit lengthy, so there's also the option ...
2. put this into the `rescue`

@benilovj am I being too paranoid?

thomasleese: Yeah I think I agree with you, if the file contains `DOSE_SEQUENCE` values we should try and import them even if they can only be one value (`1`).

misaka (approved):
👍

PR #1552, opened 2024-07-29
Match patients against fields other than NHS number

thomasleese:
This updates how we create patients as part of the import to match against existing patients on fields beyond just the NHS number. This is because the NHS number is optional, and may not exist in the individual rows.

To support this functionality I've also had to refactor the import code as it currently de-duplicates based on NHS numbers, which doesn't work when they can be `nil`.


misaka:
Should `ImmunisationImport::Row` be broken out into a separate file maybe? It's getting a bit unwieldy in `immunisation_import.rb`.

misaka (approved):
Nice work, looks good. I do feel that `to_patient`, `to_session`, etc, might read better if they was just `patient`, `session`, etc, especially now that they are memoised, but maybe that can be looked if/when `Row` is broken out to another file.

🚢

thomasleese:
> Nice work, looks good. I do feel that `to_patient`, `to_session`, etc, might read better if they was just `patient`, `session`, etc, especially now that they are memoised, but maybe that can be looked if/when `Row` is broken out to another file.

I was wondering about that too actually, and since you've suggested it too, I will include that as part of the refactor to split the file out.

PR #1556, opened 2024-07-29
Add job to send DPS export using MESH

misaka:
This adds an `ActiveJob` to send reports to DPS via MESH, and some rake tasks to manage it. It doesn't include scheduling the job, that's to come when we have a better understanding of what's needed; this PR is more about getting the MESH side working.

Instructions on how to do work with this locally have been added to the README.


README.md
```diff
@@ -245,6 +245,12 @@ govuk_notify:
 You should set it to `enabled: false` when you're done testing Notify locally,
 because it's easier to work offline without it.
 
+### Reply-To
+
+GOVUK Notify can store reply-to email addresses and use them when sending mail.
```
thomasleese, line 250:
```suggestion
GOV.UK Notify can store reply-to email addresses and use them when sending mail.
```

README.md
```diff
@@ -245,6 +245,12 @@ govuk_notify:
...
+Once you've added the reply-to email in GOVUK Notify, get the UUID and add it to
```
thomasleese, line 251:
```suggestion
Once you've added the reply-to email in GOV.UK Notify, get the UUID and add it to
```

app/jobs/dps_export_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class DPSExportJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    dps_export =
+      DPSExport.new(VaccinationRecord.where(exported_to_dps_at: nil)).export_csv
```
thomasleese, line 8:
```suggestion
      DPSExport.new(VaccinationRecord.recorded.where(exported_to_dps_at: nil)).export_csv
```

I know there's a default scope on the model, but I prefer things to be explicit. When I first read this code I thought it would be sending draft records to DPS.

lib/tasks/mesh_ack_message.rake
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+desc "Acknowledge message MESH, removing it from inbox"
+task "mesh:ack_message", [:message] => :environment do |_, args|
```
thomasleese, line 6: Would it be better to use a namespace here? https://ruby.github.io/rake/doc/rakefile_rdoc.html#label-Namespaces

lib/tasks/mesh_ack_message.rake
```diff
@@ -0,0 +1,28 @@
...
+  conn =
+    Faraday.new do |faraday|
+      faraday.ssl[:verify] = false if Rails.env.development?
+      faraday.headers = headers
+    end
+
+  response =
+    conn.put(
+      "#{base_url}/messageexchange/#{mailbox}/inbox/#{message}/status/acknowledged"
+    )
```
thomasleese, line 26: I'm wondering if there's a way we can define a shared Faraday connection instance for MESH using the `base_url` which can then be used in multiple places, perhaps in the `MESH` module.


```rb
def self.connection
  @connection ||= Faraday.new(Settings.mesh.base_url, headers: {
    "Accept" => "application/vnd.mesh.v2+json",
    "Authorization" => "#{MESH::SCHEMA} #{authorisation}"
  }, ssl: { verify: Rails.env.production? })
end
```

Then in this task all you'd need is:

```rb
response = MESH.connection.put("/messageexchange/#{mailbox}/inbox/#{message}/status/acknowledged")
```

And it would reduce the duplication we've got across the `MESH` module, and these Rake tasks.

lib/tasks/mesh_ack_message.rake
misaka: Ya, I considered it, but then ¯\_(ツ)_/¯ and didn't do it. I don't mind though, will move.

lib/tasks/mesh_ack_message.rake
misaka: Ya, this makes sense. 👍

thomasleese (approved):

PR #1567, opened 2024-07-30
Update vaccine modelling

thomasleese:
This makes a number of changes to the vaccine model to support our upcoming work related to importing vaccine information:

- Add validation and `not null` to most fields
- Add unique constraints on globally unique fields
- Rename supplier to manufacturer

Most of the changes here relate to changes I had to make to the tests/factories to support these changes, which I've included as a separate commit.


thomasleese:
> It must have been frustrating to make as many alterations to the specs as you've had to make in [e10b527](https://github.com/nhsuk/manage-vaccinations-in-schools/commit/e10b527dee6baa30acb9563c42805541fc316d6e), and I wonder whether it's something that the factories could handle better, but let's roll forward.

I'm hoping I've left the factories in a better state than I started, but yeah, I think part of the problem is that a lot of factories end up creating new campaigns, and from there we get a whole new set of vaccines, etc. I was wondering if there was a way we could get factory bot to do a "find or create" on vaccines, but it didn't seem officially supported, and I didn't like the implicit nature of that.

I think the work in #1550 might be relevant here in terms of how we manage the vaccine data.

PR #1570, opened 2024-07-31
Seed vaccines from an initialiser configuration 

thomasleese:
This builds on top of #1550 to use the configuration in a number of ways:

- automatically define factory traits that we can use in our tests
- populate vaccines as part of running `db:seed`
- provide a Rake task (`vaccines:seed`) that can be used to populate only the vaccines in deployed environments


tvararu (approved):

tvararu:
Closes #1550

PR #1576, opened 2024-08-02
Bigfix: set session time of day during imms import

benilovj:
Also validate that it's set on for non-draft sessions.


misaka:
Are there any tests for this?

tvararu (approved):
Yeah, I forgot to do this!

@misaka I'm happy to just ship this, testing this feels like minutiae

benilovj:
> Are there any tests for this?

I've now added a spec to `ImmunisationImportSpec` to prevent regressions.

PR #1581, opened 2024-08-07
Fewer crashes on sparse sessions

benilovj:
Follow-up to #1576.

### DIFF
--- a/app/components/app_session_details_component.rb
+++ b/app/components/app_session_details_component.rb
@@ -29,14 +29,20 @@ class AppSessionDetailsComponent < ViewComponent::Base
   end
 
   def consent_requests
-    "Send on #{@session.send_consent_at.to_fs(:long_day_of_week)}"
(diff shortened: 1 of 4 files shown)


app/components/app_session_details_component.rb
```diff
@@ -29,14 +29,20 @@ def cohort
...
 
   def reminders
-    "Send on #{@session.send_reminders_at.to_fs(:long_day_of_week)}"
+    if @session.send_reminders_at.present?
+      "Send on #{@session.send_reminders_at.to_fs(:long_day_of_week)}"
+    end
   end
 
   def deadline_for_responses
+    return nil unless @session.close_consent_at.present?
```
tvararu, line 44:
```suggestion
    return nil if @session.close_consent_at.blank?
```

tvararu (approved):
I've suggested the linter fix 👍

PR #1583, opened 2024-08-07
Vaccinations page: disambiguate dose (dose volume), show dose number for non-seasonal 

benilovj:
## Flu: before

image

## Flu: after

image

## HPV: before

image

## HPV: after

image

### DIFF
--- a/app/components/app_vaccination_record_details_component.rb
+++ b/app/components/app_vaccination_record_details_component.rb
@@ -9,6 +9,28 @@ class AppVaccinationRecordDetailsComponent < ViewComponent::Base
     @batch = vaccination_record.batch
   end
 
+  def dose_number
+    return nil if @vaccine.nil? || @vaccine.seasonal?
+
+    numbers_to_words = {
+      1 => "First",
+      2 => "Second",
+      3 => "Third",
+      4 => "Fourth",
+      5 => "Fifth",
+      6 => "Sixth",
+      7 => "Seventh",
+      8 => "Eighth",
+      9 => "Ninth"
+    }.freeze
+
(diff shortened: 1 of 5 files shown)


app/components/app_vaccination_record_details_component.rb
```diff
@@ -9,6 +9,28 @@ def initialize(vaccination_record)
...
+      4 => "Fourth",
+      5 => "Fifth",
+      6 => "Sixth",
+      7 => "Seventh",
+      8 => "Eighth",
+      9 => "Ninth"
+    }.freeze
+
+    if @vaccination_record.dose_sequence <= 9
+      numbers_to_words[@vaccination_record.dose_sequence]
```
misaka, line 28: Did you remember/consider `ordinalize`?

```ruby
1.ordinalize     # => "1st"
2.ordinalize     # => "2nd"
```

app/components/app_vaccination_record_details_component.rb
```diff
@@ -9,6 +9,28 @@ def initialize(vaccination_record)
...
+    else
+      @vaccination_record.dose_sequence
```
misaka, line 30: Even if you prefer the words above, `.ordinalize` could work here.

```suggestion
      @vaccination_record.dose_sequence.ordinalize
```

app/components/app_vaccination_record_details_component.rb
benilovj: I did, but the design in the prototype explicitly used words and I don't think Rails has anything out of the box for that, and bringing in a gem just for the above was overkill 🤷

misaka (approved):

PR #1590, opened 2024-08-09
Introduce `dev_tools` feature flag

benilovj:
* Replace the `make_session_in_progress_button` feature flag with `dev_tools`
* Put DPS export reset functionality behind it

### DIFF
--- a/app/views/immunisation_imports/index.html.erb
+++ b/app/views/immunisation_imports/index.html.erb
@@ -63,11 +63,9 @@
 </div>
 
 <% content_for :after_main do %>
-  <% if Flipper.enabled? :dev_tools %>
-    <%= render(AppDevToolsComponent.new) do %>
-      <%= govuk_button_to "Download DPS export", dps_export_campaign_reports_path(@campaign) %>
+  <%= render(AppDevToolsComponent.new) do %>
+    <%= govuk_button_to "Download DPS export", dps_export_campaign_reports_path(@campaign) %>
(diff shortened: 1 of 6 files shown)


app/views/immunisation_imports/index.html.erb
```diff
@@ -63,9 +63,11 @@
 </div>
 
 <% content_for :after_main do %>
-  <%= render(AppDevToolsComponent.new) do %>
-    <%= govuk_button_to "Download DPS export", dps_export_campaign_reports_path(@campaign) %>
+  <% if Flipper.enabled? :dev_tools %>
```
thomasleese, line 66: Could the feature flag check be part of the `AppDevToolsComponent` component? Would we ever want the component to render without the dev tools flag being enabled?

We [already have a `render?` method on that component that checks that we're not running in production](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/940f20284bd622ad5382ab58884cc7202c4e6601/app/components/app_dev_tools_component.rb#L18-L20).

benilovj: @thomasleese like this? [828f0bb](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1590/commits/828f0bbf985bf1ff644c77b8015f1cfb1944a520)

thomasleese (approved):

misaka (approved):

PR #1594, opened 2024-08-12
Add MESH configuration for staging deployed env

misaka:
This should allow Mavis to connect to MESH in the staging environment. This is also dependent on secrets being updated in AWS Session Manager's Parameter Store, which has been updated but still needs to be fully tested, which I will do as soon as this is deployed to `test.mavistesting.com`. These changes have been tested locally.


config/settings/development.yml
```diff
@@ -7,6 +7,7 @@ cis2:
 
 mesh:
   base_url: https://localhost:8700
+  disable_ssl_verification?: true
```
tvararu, line 10: It's a little bit weird that this has a ? in it since it could be an ENV var but I guess it works!

tvararu (approved):

config/settings/development.yml
misaka: Hrm, ya, a little weird. I'll remove it just in case.

PR #1597, opened 2024-08-12
Prevent detection of NHS numbers as telephone numbers

paulrobertlloyd:
Currently, in certain scenarios (notably iOS Safari, perhaps other user agents), NHS numbers get detected as being phone numbers and are linked to accordingly. Trying to copy an NHS number, or accidentally clicking on a linked NHS number may initiate a phone call.

We can prevent numbers from being detected as phone numbers by adding the following meta tag:

```html
<meta name="format-detection" content="telephone=no">
```

However, that prevents actual telephone numbers from being linked as well, which is useful. We could revert this behaviour by adding a link with the `tel:` protocol to telephone numbers, but linked phone numbers may not make sense on desktop browsers where there is less likelihood of there being a telephony function.

What we need to do is opt-out NHS numbers from being detected as phone numbers.

In my experimentation, replacing a space (`U+0020`) with any space-like character doesn’t prevent format detection from happening. However, adding a ZWJ (zero-width joiner, `U+200D`) does.

Replacing a space with a ZWJ removes the space between number groups; adding a non-breaking space plus the ZWJ recreates the previous appearance. (It doesn’t need to be a non-breaking space, but as the number shouldn’t wrap where there’s a space, this makes sense to use).

Am using `&zwj;` as this entity was introduced in HTML 4.0, so hopefully has near universal browser support.

💅

> [!NOTE]
> I mostly got the tests to pass, but am having trouble getting Capybara’s `have_content` method detect the special characters correctly. I tried using HTML entities and Unicode references, and `have_text`, but none of these seemed to work).

### DIFF
--- a/app/helpers/patients_helper.rb
+++ b/app/helpers/patients_helper.rb
@@ -1,10 +1,15 @@
 # frozen_string_literal: true
 
+# Replace each space in NHS number with a non-breaking space and
+# zero-width word joiner to prevent telephone format detection
 module PatientsHelper
   def format_nhs_number(nhs_number)
     if nhs_number.present?
       tag.span(class: "app-u-monospace") do
-        nhs_number.to_s.gsub(/(\d{3})(\d{3})(\d{4})/, "\\1 \\2 \\3")
+        nhs_number
+          .to_s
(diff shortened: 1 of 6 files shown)


tvararu (approved):

tvararu:
@paulrobertlloyd I've fixed the tests, going to merge this, thank you!

PR #1606, opened 2024-08-13
Add MESH mailbox validation

misaka:
This adds a function to validate the MESH mailbox, and enables it in the `staging` environment so that we can test it.

> [!NOTE]
> This job requires the feature flag `mesh_jobs` to be enabled. For now this will be disabled, we can enable it when we are ready to do testing.


spec/jobs/mesh_validate_mailbox_job_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe MESHValidateMailboxJob do
+  it "calls MESH.validate_mailbox" do
+    Flipper.enable(:mesh_jobs)
+    allow(MESH).to receive(:validate_mailbox)
```
thomasleese, line 8: This line could go in a `before` block as it's used in both tests.

app/jobs/mesh_validate_mailbox_job.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class MESHValidateMailboxJob < ApplicationJob
+  queue_as :default
```
thomasleese, line 4: We've got a number of different kinds of jobs now that all go in the same queue, do we need to think about prioritisation here? For example, I imagine we wouldn't want this job to get stuck behind loads of consent reminders.

misaka: Ya. Sounds reasonable to create a queue for MESH

spec/jobs/mesh_validate_mailbox_job_spec.rb
misaka: Amended

app/jobs/mesh_validate_mailbox_job.rb
misaka: Added the queue

thomasleese (approved):

PR #1616, opened 2024-08-14
Associate DPS exports and vaccination records 

tvararu:
This adds a has and belongs to many relationship instead of the previous method, and removes the reliance on the exported_to_dps_at timestamp.


thomasleese (approved):

app/models/dps_export.rb
```diff
@@ -27,26 +27,36 @@ class DPSExport < ApplicationRecord
...
+  end
+
   def export!
-    csv.tap do
-      unexported_vaccination_records.update_all(
-        exported_to_dps_at: Time.zone.now
-      )
-    end
+    csv
   end
```
misaka, line 56: Can we drop this method?

app/models/dps_export.rb
```diff
@@ -27,26 +27,36 @@ class DPSExport < ApplicationRecord
...
 
-        unexported_vaccination_records.each { csv << DPSExportRow.new(_1).to_a }
+        vaccination_records.each { csv << DPSExportRow.new(_1).to_a }
       end
   end
 
+  def populate!
+    self.vaccination_records =
+      VaccinationRecord
+        .includes(:batch, :location, :patient, :session, :team, :vaccine)
```
thomasleese, line 44: Is the `includes` here in the assignment so when we call `vaccination_records` again later in `export!` it's already fetched from the database?

I wonder if there's a way of making that clearer, or supporting the case where `populate!` hasn't been called first.

For example doing `DPSExport.find(...).csv` would have N+1 issues as it wouldn't have associated included.

misaka: Would it make sense to set `vaccination_records` in the `initialize`?

tvararu: Have updated this and pushed. I'm really stumped why the test is failing though. It says it violates the uniqueness constraint, but I don't get why that's getting triggered.

misaka (approved):

PR #1617, opened 2024-08-14
Add academic year to campaigns

thomasleese:
This adds an academic year column to campaigns which will be required when creating a new campaign, and is then displayed in a number of places in the service.

Academic years span two real years, the field will store the first real year, for example the value 2024 means the academic year 2024/25.

## Screenshots

Screenshot 2024-08-14 at 19 11 56

Screenshot 2024-08-14 at 19 12 02


benilovj (approved):

app/views/campaigns/sessions.html.erb
```diff
@@ -8,6 +7,8 @@
                                         ]) %>
 <% end %>
 
+<%= render "name_and_academic_year" %>
```
misaka, line 10: Have you considered making this a component? Theo and I laid out a game plan where we would progressively adopt components for these types of things, one advantage is we can keep the helper logic and the display logic in the same file (see `app/components/app_timestamped_entry_component.rb` for example).

thomasleese: I did consider it but for something small like this I decided a shared view was simpler. If the plan is to aim to use components then I'm happy to refactor this and move it in to a component.

I don't think it's quite as simple as moving the academic year formatting logic in to a component because we show the academic year in a number of different places, unless you're suggesting we have a component that effectively [replaces the logic in the helper](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/e85924d71190b0922798e35e62519d6f924ac79c/app/helpers/campaigns_helper.rb#L4-L8) and another which renders the name and academic year?

misaka: Ya, that's fair. I hadn't seen that `campaign_academic_year` was being used in a few places, make sense. 👍

thomasleese: I do think there's a discussion to be had about helpers vs components though, I think ideally we'd pick one approach and stick with it, but sometimes components can heavily quite heavy for something as simple as formatting a piece of text as in this example.

tvararu: 👍

PR #1618, opened 2024-08-15
Add start and end dates to campaigns

thomasleese:
This adds a start date and end date column to campaigns with some validation to ensure the values are reasonable for the academic year. This values will need to be set as part of the creation of a campaign, and will be used to valid the vaccinations records belong to the campaign.


benilovj (approved):

app/models/campaign.rb
```diff
@@ -37,4 +37,28 @@ class Campaign < ApplicationRecord
               greater_than_or_equal_to: 2000,
               less_than_or_equal_to: Time.zone.today.year + 5
             }
+
+  validates :start_date,
+            comparison: {
+              greater_than_or_equal_to: :first_possible_start_date,
+              if: :academic_year,
```
misaka, line 44: I think `academic_year` is NOT NULL, so would this have an effect?

thomasleese: If the campaign hasn't been saved to the database yet it's possible to create it with a `nil` value (for example `Campaign.new(academic_year: nil)`). Calling `.valid?` on that object would then raise an error from trying to call `first_possible_start_date` with a `nil` `academic_year`.

misaka (approved):

PR #1627, opened 2024-08-16
Add interface for creating campaigns

thomasleese:
This adds a new step-by-step interface for creating new campaigns based on the designs in the prototype. I've introduced a new controller which handles all the logic related to the steps using the Wicked Gem that we're using in other places.

## Screenshots

Screenshot 2024-08-16 at 15 13 08

Screenshot 2024-08-16 at 10 11 22

Screenshot 2024-08-16 at 14 37 47

Screenshot 2024-08-16 at 14 46 38

### DIFF
--- a/app/controllers/campaigns/new_controller.rb
+++ b/app/controllers/campaigns/new_controller.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+class Campaigns::NewController < ApplicationController
+  include Wicked::Wizard
+
+  before_action :set_campaign
+  before_action :set_steps
+  before_action :setup_wizard
+
+  skip_after_action :verify_policy_scoped
+
+  layout "two_thirds"
+
+  def show
+    render_wizard
+  end
+
+  def update
+    params = send("#{current_step}_params")
+
+    @campaign.assign_attributes(form_step: current_step, **params)
+
+    render_wizard(@campaign)
+  end
+
+  private
+
+  def set_campaign
+    @campaign =
+      policy_scope(Campaign).where(
+        active: params[:id] == Wicked::FINISH_STEP
+      ).find(params[:campaign_id])
+  end
+
+  def set_steps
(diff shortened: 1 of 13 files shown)


spec/features/create_campaign_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    and_i_click_continue
+    then_i_should_see_the_dates_page
+
+    when_i_fill_in_the_dates
+    and_i_click_continue
+    then_i_should_see_the_confirm_page
+
+    when_i_confirm_the_campaign
+    then_i_should_see_the_campaigns_page
+    and_i_should_see_the_flu_campaign
```
benilovj, line 26: I wonder whether it's worth adding coverage here (and for the HPV scenario) that only active vaccines are added to the campaign?

benilovj (approved):

app/controllers/campaigns/new_controller.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+class Campaigns::NewController < ApplicationController
```
tvararu, line 3: We usually call the other ones `*EditController` rather than `NewController` but this is okay too

tvararu (approved):

spec/features/create_campaign_spec.rb
thomasleese: Yeah I think it's probably a good idea, I'll amend it.

app/controllers/campaigns/new_controller.rb
thomasleese: Yeah I did notice that, I decided to go with new because in the prototype there's a separate edit flow which I thought could be implemented with an edit controller in the future. Do you think it's worth renaming this here?

tvararu: Nope, that makes sense!

PR #1630, opened 2024-08-19
Validate vaccination record administered at date

thomasleese:
This validates the `administered_at` column to ensure that it's always within the start and end date of the campaign. We need this to ensure that we're only recording vaccinations that are part of the campaign.

### DIFF
--- a/spec/models/vaccination_record_spec.rb
+++ b/spec/models/vaccination_record_spec.rb
@@ -39,25 +39,50 @@
 #
 require "rails_helper"
 
-describe VaccinationRecord do
-  it "validates that the vaccine and the batch vaccines match" do
-    patient_session = create(:patient_session)
-    vaccine = patient_session.campaign.vaccines.first
-    different_vaccine = create(:vaccine)
-    batch = create(:batch, vaccine: different_vaccine)
-
-    subject =
-      build(
-        :vaccination_record,
-        administered: true,
-        vaccine:,
-        batch:,
-        patient_session:
-      )
-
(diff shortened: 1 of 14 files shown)


spec/models/vaccination_record_spec.rb
```diff
@@ -39,25 +39,50 @@
...
 
-    subject =
-      build(
-        :vaccination_record,
-        administered: true,
-        vaccine:,
-        batch:,
-        patient_session:
+  describe "validations" do
+    it do
```
misaka, line 62: Can we give this spec a description? I think it makes sense to skip a description if the spec all fits on a line, but this one spans a few lines.

thomasleese: Yeah no problem, I've added one.

misaka:
@thomasleese Just to understand, why do we validate the dates for `VaccinationRecord` in the model but for `Session` it's in the importer?

thomasleese:
> @thomasleese Just to understand, why do we validate the dates for `VaccinationRecord` in the model but for `Session` it's in the importer?

Effectively we have the same validation in two places. The validation in the importer is so we capture the problem at a higher level and show the errors to the user before trying to even create a `VaccinationRecord`. The validation in the model itself is to capture when we create vaccination records outside the importer, via the normal interface in the app, it's more of a last defence because I imagine the interface wouldn't even allow you to get to that page if the campaign is finished.

misaka (approved):

PR #1665, opened 2024-08-22
Update dps mesh mailbox in staging

misaka:
Update this mailbox to one we have access to so we can see what gets sent to it for testing purposes.


misaka:
> I trust that's the right combination of characters!

🤞

PR #1666, opened 2024-08-22
Update dev functions for DPS exports

misaka:
Broken with an earlier change, this is for dev only and will either be enhanced or removed altogether at a later time.

### DIFF
--- a/app/controllers/vaccination_records_controller.rb
+++ b/app/controllers/vaccination_records_controller.rb
@@ -13,14 +13,15 @@ class VaccinationRecordsController < ApplicationController
   end
 
   def export_dps
-    send_data(dps_export.export!, filename: dps_export.filename)
+    send_data(dps_export.csv, filename: dps_export.filename)
   end
 
   def reset_dps_export
-    dps_export.reset!
+    campaign.dps_exports.each { _1.vaccination_records = [] }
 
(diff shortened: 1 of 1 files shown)


app/controllers/vaccination_records_controller.rb
```diff
@@ -13,14 +13,15 @@ def show
   end
 
   def export_dps
-    send_data(dps_export.export!, filename: dps_export.filename)
+    send_data(dps_export.csv, filename: dps_export.filename)
   end
 
   def reset_dps_export
-    dps_export.reset!
+    campaign.dps_exports.each { _1.vaccination_records = [] }
```
thomasleese, line 20: I don't think this will commit the change to the database, you'll likely need to call `save` somewhere.

misaka: This actually saves the change immediately. I did double-check this, as I wasn't quite 100% sure myself. I have this vague notion that maybe this is configurable somewhere, but also this semantic kind of makes sense since there's no change directly on the `DPSExport` record to save.

image

thomasleese: Oh interesting, I wouldn't have expected that!

thomasleese (approved):

PR #1672, opened 2024-08-23
Override `nhsuk-` brand prefix on notification banner and panel components

paulrobertlloyd:
Trialing the change to `govuk-components` proposed at https://github.com/x-govuk/govuk-components/pull/552 (see https://github.com/x-govuk/govuk-components/issues/547)

This allows us to use `govuk-` brand prefix for components that do not have associated styles in `nhsuk-frontend`, and thus not need to extend the `govuk-` prefixed classes to apply to components that have been forced to use `nhsuk-` prefix given the global `brand` setting for this gem.

There may be a corresponding update the GOV.UK Form Builder gem; that will allow us to review and remove styles for the password input and file upload form fields in a similar fashion.

Will merge once these upstream updates have been published. /cc @peteryates


tvararu:
Thanks @paulrobertlloyd!

thomasleese (approved):

PR #1673, opened 2024-08-23
Show immunisation import statistics

thomasleese:
If there are duplicate records or not administered records, we want to include this information in a warning callout at the top of the page.

The warning callout component is part of the NHS design system, but not part of GOV.UK's so we don't have a component in govuk-components for it. As this is the first time we're using it, I decided against creating a component for it, but we might want to do add one in the future.

## Screenshots

Screenshot 2024-08-23 at 16 54 15


tvararu (approved):

app/models/immunisation_import.rb
```diff
@@ -177,4 +186,16 @@ def rows_are_valid
...
     end
   end
+
+  def counts_are_nil_or_present
+    if processed?
+      COUNT_COLUMNS.each do |column|
+        errors.add(column, :blank) if send(column).nil?
+      end
+    else
+      COUNT_COLUMNS.each do |column|
```
misaka, line 196: When would this path ever be taken? Do we really need to account for it?

misaka (commented):
Sorry, this comment got caught in the rebase and turned into a review comment. Not a blocker, but just wondering if we really need to accommodate that code path.

app/models/immunisation_import.rb
thomasleese: Yeah I was in two minds about whether it was worth doing, I put it in because it is a requirement for the stats to have been calculated if it's processed, but you're right that there's only one code paths which processes these. I figured there was no harm in having it in there, and best case it'll spot perhaps a manual update to the records in the future.

misaka: Would it make sense to use something that would raise here? That would give us an alert in Sentry which might be a clearer indication of our intention here.

But ¯\\\_(ツ)_/¯ ... :shipit: first

PR #1677, opened 2024-08-26
Refactor form views

thomasleese:
This makes a number of small refactors to the form views to ensure we're rendering them consistently, the only change on the frontend should be that the error summary component appears at the top of the page, as it should do according to the design system.


tvararu (commented):
🧹 🧹 🧹

tvararu (approved):

PR #1679, opened 2024-08-26
Add ability to edit vaccination record date/time

thomasleese:
This adds the ability to edit the date and time an existing vaccination record.

I've decided to add these actions to a new controller as it's likely we'll need to edit different parts of the vaccination records in the future. I've also set up the controller in a similar way to how we handle wizards as I imagine we'll likely need that functionality in the future.

The `govuk_form_builder` Gem doesn't currently support time fields (and there's no mention of how they should work in the design system) so I've had to re-create it myself in the form.

If the user is on this page we should expect the date/time to always be set, whereas we allow the field to be optional on the model. One way to solve this could be to use a form object, however for this page I've decided to add some validation logic to the controller. We could refactor this in to a form object in the future.

I've opened this PR to discuss whether a form object would be suitable here, I will also need to integrate with the `DateParamsValidator` before marking this as ready for review.

This PR doesn't implement the feature exactly as designed, where multiple changes can be made to vaccination records and then committed together in one go, this could be implemented later as we add more fields to change.

## Screenshots

Screenshot 2024-08-26 at 17 21 20
Screenshot 2024-08-26 at 17 21 16

### DIFF
--- a/app/controllers/vaccination_records/edit_controller.rb
+++ b/app/controllers/vaccination_records/edit_controller.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+class VaccinationRecords::EditController < ApplicationController
+  before_action :set_vaccination_record
+  before_action :set_campaign
+
+  def edit_date_and_time
+    render :date_and_time
+  end
+
+  def update_date_and_time
+    @vaccination_record.assign_attributes(vaccination_record_params)
+
+    if @vaccination_record.administered_at.nil?
+      @vaccination_record.errors.add(:administered_at, :blank)
+      render :date_and_time, status: :unprocessable_entity
+    elsif @vaccination_record.save
+      redirect_to campaign_vaccination_record_path(
+                    @campaign,
+                    @vaccination_record
+                  )
+    else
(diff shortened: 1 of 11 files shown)


misaka (approved):
🚀

app/controllers/vaccination_records/edit_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+class VaccinationRecords::EditController < ApplicationController
+  before_action :set_vaccination_record
+  before_action :set_campaign
+
+  def edit_date_and_time
+    render :date_and_time
+  end
+
+  def update_date_and_time
```
tvararu, line 11: Couldn't these be `show` and `update`? I guess the idea is that we'll add more later?

tvararu (approved):

app/controllers/vaccination_records/edit_controller.rb
thomasleese: Yeah, that was the idea basically, I can rename them if you'd prefer?

tvararu: Hmm if we add more later it's fine

PR #1685, opened 2024-08-27
Remove CSV data after 30 days

thomasleese:
This adds a background job which automatically deletes the CSV data of immunisation imports after 30 days to avoid the service holding this data for longer than it needs to (which would contain personally identifiable information).

This makes no attempt to sanitize the data, it simply sets the value to `nil`, however I have added the ability to capture the filename of the upload so if we needed access to the raw data after 30 days we would have something to go on.


misaka (approved):
👍

PR #1686, opened 2024-08-27
Switch to ActiveSupport::Testing::TimeHelpers

thomasleese:
This replaces all uses of Timecop with the built-in time helpers provided by Rails directly, to reduce the number of third party libraries.


misaka (approved):

tvararu (commented):
Awesome!

PR #1688, opened 2024-08-27
Add and populate patient.pending_changes

tvararu:
This will be used to implement draft-like functionality on patients when importing vaccination records.

During the import, we compare the attributes on the patient versus the ones on the row and update the `pending_changes` field with the new values. These can then be displayed to the user so they can decide which ones to keep.


app/models/immunisation_import_row.rb
```diff
@@ -99,18 +99,12 @@
...
-          imported_from:,
-          last_name: patient_last_name,
-          nhs_number: patient_nhs_number,
-          school:
-        )
+      if (existing_patient = find_existing_patients.first)
+        existing_patient.stage_changes(patient_attributes)
+        existing_patient
+      else
+        Patient.create!(patient_attributes)
```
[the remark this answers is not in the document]
tvararu: We can't encrypt date_of_birth because AR encryption only works on string-y fields.

app/models/immunisation_import_row.rb
```diff
@@ -402,4 +396,18 @@ def zero_or_one_existing_patient
...
+    {
+      address_postcode: patient_postcode,
+      date_of_birth: patient_date_of_birth,
+      first_name: patient_first_name,
+      gender_code: patient_gender_code,
+      home_educated:,
+      imported_from:,
+      last_name: patient_last_name,
+      nhs_number: patient_nhs_number,
+      school:
```
thomasleese, line 410: How would we handle `school` (and `imported_from`) since these are foreign keys? Do we need to add something to `stage_changes` to store the ID in the `pending_changes` field?

tvararu: Good point. Maybe we should silently omit these keys from the `stage_changes` call?

thomasleese: I agree about `imported_from`, but I think for `school` it actually works quite nicely if we could support it as part of this feature, if a patient has moved schools between imports then it would be good to flag it like the other changed fields.

tvararu: @thomasleese yeah, fair point, let me update it to work with schools and add a test 👍

thomasleese (approved):

PR #1690, opened 2024-08-29
Temporarily limit locations when creating sessions

tvararu:
There are 51k+ locations in the database after running the GIAS import. We can't display all of them at the same time, and instead should use something like an autocomplete that fetches locations from the server.


thomasleese (approved):
Is it worth adding a `# TODO` comment to the line?

thomasleese (approved):

PR #1691, opened 2024-08-30
Add support for NHS API platform and PDS

misaka:
This allows us to connect to PDS using the NHS API. It adds authorisation code, with a rake task to test retrieving the access token and a task to retrieve patients using their NHS number.

This has been tested against the INT environment, and credentials for tnat env have been added to the app secrets in AWS. There are also instructions included to help replicate testing. What's been punted to later PRs (to support PDS integration):

- support for using the PDS search endpoint
- integration with the app to retrieve NHS numbers for children without them


tvararu (commented):
⛵

tvararu (approved):

PR #1692, opened 2024-08-30
Handle duplicate patient records during immunisation import

tvararu:
This implements the design that allows users to apply the `pending_changes` that are added to patient records when matching records that match on NHS Number are uploaded a second time.

TODO in follow-up PRs:

- [ ] Implement the diffing correctly
- [ ] Implement the form to keep or overwrite changes
- [ ] Vaccination record review
- [ ] Breadcrumbs
- [ ] Send an updated record to MESH if the user chooses to use the duplicate record
- [ ] If the duplicate records aren't reviewed and resolved before uploading the batch, then they appear in the "Upload issues" area (see prototype)
- [ ] Test 2+ patients, not just 1 record

### Screenshots

(Font is wrong because these are taken from tests)

image

image

### DIFF
--- a/app/controllers/immunisation_imports/patients_controller.rb
+++ b/app/controllers/immunisation_imports/patients_controller.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class ImmunisationImports::PatientsController < ApplicationController
+  before_action :set_campaign
+  before_action :set_immunisation_import
+  before_action :set_patient
+
+  def show
+    render layout: "full"
+  end
+
+  def update
+  end
+
+  private
+
+  def set_campaign
+    @campaign =
+      policy_scope(Campaign)
+        .active
+        .includes(:immunisation_imports)
+        .find(params[:campaign_id])
+  end
+
+  def set_immunisation_import
+    @immunisation_import =
+      @campaign.immunisation_imports.find(params[:immunisation_import_id])
+  end
+
+  def set_patient
+    @patient =
+      Patient
+        .joins(patient_sessions: :vaccination_records)
+        .where(
+          patient_sessions: {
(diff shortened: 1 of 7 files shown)


app/controllers/immunisation_imports/patients_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      @campaign.immunisation_imports.find(params[:immunisation_import_id])
+  end
+
+  def set_patient
+    @patient =
+      @immunisation_import
+        .vaccination_records
+        .map(&:patient)
+        .find(params[:id])
+        .first
```
thomasleese, line 36: Do we need the `.first` here? I'd expect `find` to return a single object.

tvararu: It returned an enumerator initially but I'll look at this code again. I'm also not pleased that I have to `map(&:patients)` and iterate in Ruby instead of AR, so I'd like to golf this to something else.

tvararu: I've updated this with an AR-based version; it doesn't look too great, but it's faster

app/controllers/immunisation_imports/patients_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def set_patient
+    @patient =
+      Patient
+        .joins(patient_sessions: :vaccination_records)
+        .where(
+          patient_sessions: {
+            vaccination_records: {
+              imported_from_id: @immunisation_import.id
```
thomasleese, line 37:
```suggestion
              imported_from: @immunisation_import
```

Small improvement, but no big deal.

thomasleese (approved):

PR #1706, opened 2024-09-04
Display highlighted changes to patients 

tvararu:
Refactor the duplicate patient summary to use a component.

Add a `with_pending_changes` method to patients that assigns all the attributes, allowing us to check whether they're changed using [ActiveModel::Dirty](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html) built-ins.

### Screenshots

image

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -118,6 +118,19 @@ class Patient < ApplicationRecord
     update!(pending_changes:) if pending_changes.any?
   end
 
+  def with_pending_changes
+    return self if pending_changes.blank?
+
+    dup.tap do |patient|
+      patient.clear_changes_information
+      pending_changes.each do |attr, value|
+        if patient.respond_to?("#{attr}=")
+          patient.public_send("#{attr}=", value)
+        end
+      end
+    end
+  end
+
   private
 
   def remove_spaces_from_nhs_number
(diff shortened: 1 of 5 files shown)


app/models/patient.rb
```diff
@@ -118,6 +118,19 @@ def stage_changes(attributes)
...
+  def with_pending_changes
+    return self if pending_changes.blank?
+
+    dup.tap do |patient|
+      patient.clear_changes_information
+      pending_changes.each do |attr, value|
+        if patient.respond_to?("#{attr}=")
+          patient.public_send("#{attr}=", value)
+        end
+      end
```
thomasleese, line 130: One for later perhaps, but I wonder if we should then clear the `pending_changes` field for this object?

It would allow us to do `patient.with_pending_changes.save!` in one go when we want to apply the pending changes.

I really like this approach though, works well!

thomasleese (approved):

app/models/patient.rb
tvararu: One issue with that is that this returns a `.dup` so the `id` field is blank; we couldn't call `.save!` on it as it will create a new record.

However, yeah that would be nice.

thomasleese: Oh right, I didn't realise `dup` cleared the `id`.

PR #1711, opened 2024-09-06
Rename campaign to programme

thomasleese:
This renames the campaign model to programme, which is the term we're using across the user interface of the service. This is for consistency across the codebase and so we can stick to a single naming convention.

Most of these changes were done automatically (find and replace and refactor in RubyMine), however a number of additional fixes were done manually.

This should have no functional changes to the service itself.


tvararu (approved):
Skimmed through, looks good!

PR #1714, opened 2024-09-08
Handle duplicate patient review form

tvararu:
Add logic to the #update method in the controller to apply or discard pending changes.

TODO:

- [x] Implement `apply_`/`discard_pending_changes`
- [x] Handle 422 and display error
- [x] Fix gender code bug
- [x] Consider https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1706#discussion_r1745054924

### Screenshots

image

### DIFF
--- a/app/controllers/immunisation_imports/patients_controller.rb
+++ b/app/controllers/immunisation_imports/patients_controller.rb
@@ -44,11 +44,7 @@ class ImmunisationImports::PatientsController < ApplicationController
   end
 
   def set_form
-    apply_changes =
-      params
-        .fetch(:patient_changes_form, {})
-        .permit(:apply_changes)
-        .fetch(:apply_changes, nil)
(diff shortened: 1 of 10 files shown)


app/controllers/immunisation_imports/patients_controller.rb
```diff
@@ -30,4 +42,14 @@ def set_immunisation_import
   def set_patient
     @patient = @immunisation_import.patients.find(params[:id])
   end
+
+  def set_form
+    apply_changes =
+      params
+        .fetch(:patient_changes_form, {})
+        .permit(:apply_changes)
+        .fetch(:apply_changes, nil)
```
thomasleese, line 51: Since you're only fetching one value, I think you can use a simpler form:

```suggestion
      params.dig(:patient_changes_form, :apply_changes)
```

thomasleese (approved):

PR #1717, opened 2024-09-09
Link imports with created records

thomasleese:
This updates the cohort and immunisation imports to create links between the import record (`CohortImport` and `ImmunisationImport`) with the objects that are getting imported. This doesn't necessarily mean that the import was the one that first created the object because of duplicates, but it does tell us whether the record was part of that import in some way.

This also adds a currently missing link between cohort imports and the patients which get imported from them.

This slightly changes the meaning of this relationship, but it introduces a number of improvements:

- We can simplify the patient and vaccination record controller for duplicate handling.
- It allows us to have a more complete picture of the import process, users can see if the same record was part of multiple imports, and trace how the object might have changed over time.
- We can still determine if an import was responsible for creating an object by comparing the `processed_at` and `created_at` times, or the `recorded_at` times, albeit without a direct relationship.


misaka (approved):
Way to blow up our tables count! 🚢

tvararu (approved):
Yeah, I was thinking this made sense while handling the duplicate import feature.

PR #1719, opened 2024-09-09
Move cohort upload link + pilot cleanup

benilovj:
* Move the cohort upload link under Programmes > [specific programme] > Cohort. This will need more work as the cohort management designs develop but it's now possible to actually navigate to this link.
* Remove the "Manage pilot" area completely
* Remove mentions of the pilot from a couple of feature specs


config/routes.rb
```diff
@@ -97,6 +89,10 @@
           controller: "vaccination_records/edit",
           action: "update_date_and_time"
     end
+
+    resource :cohort_list, as: :cohort, only: %i[new create] do
```
thomasleese, line 93:
```suggestion
    resource :cohort_import, as: :cohort, only: %i[new create] do
```

Could we name the controller to `CohortImportController` at the same time to match the renamed model? It's the last step to moving away from the old "cohort lists" naming.

thomasleese (commented):
Looks good, it's something I had on my to do list so nice to get it done! Just one comment about renaming the controller.

config/routes.rb
benilovj: done!

tvararu (approved):
Nice one!

PR #1728, opened 2024-09-10
Extend data model so that children can have more than one parent

benilovj:
This PR updates the data modelling so that `Patient` can have multiple `Parent` entities.

This brings changes to how notifications / comms work; specifically, the following are now sent to *every* parental contact for a child:

* consent requests
* consent reminders
* session reminders

Consent comms are still only sent to the parent who is on the consent.


tvararu (approved):

db/migrate/20240909161707_create_parents_patients_join_table.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class CreateParentsPatientsJoinTable < ActiveRecord::Migration[7.2]
+  def change
+    create_join_table :parents, :patients do |t|
```
thomasleese, line 5:
```suggestion
    create_join_table :parents, :patients, column_options: { foreign_key: true } do |t|
```

app/models/cohort_import.rb
```diff
@@ -77,7 +77,7 @@ def process_row(row)
 
     patient.save!
 
-    link_records(patient, patient.parent)
+    patient.parents.each { |parent| link_records(patient, parent) }
```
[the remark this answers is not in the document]
benilovj: fixed

db/migrate/20240909161707_create_parents_patients_join_table.rb
benilovj: thanks, updated.

PR #1729, opened 2024-09-10
Link parents with cohort imports

thomasleese:
This follows up from #1717 to add the same link to parents which was missed. The cohort imports create parents objects so we should link them to the imports in the same way.


app/models/parent.rb
```diff
@@ -25,13 +25,19 @@ class Parent < ApplicationRecord
...
   attr_accessor :parental_responsibility
 
   enum :contact_method, %w[text voice other any], prefix: true
   enum :relationship, %w[mother father guardian other], prefix: true
 
   encrypts :email, :name, :phone, :relationship_other
 
+  normalizes :phone,
+             with: ->(str) { str.blank? ? nil : str.to_s.gsub(/\s/, "") }
+  normalizes :email, with: ->(str) { str.nil? ? nil : str.to_s.downcase.strip }
```
tvararu, line 39: TIL https://www.bigbinary.com/blog/rails-7-1-adds-activerecord-base-normalizes

tvararu (approved):

app/models/parent.rb
thomasleese: Yeah, it's a nice feature!

PR #1731, opened 2024-09-10
Add check and confirm page for cohort imports

thomasleese:
This updates the various pages related to cohort imports to match the designs we have for immunisation imports, specifically added a two-stage process where the user has to "Check and confirm" the records before they are recorded.

The implementation is based on the existing designs for immunisation import, but I expect these will need to change slightly when the designs for cohorts are finalised. I think once we know more about the similarities and differences between cohort and immunisation imports we could refactor these views in to shared components.

### DIFF
--- a/app/controllers/cohort_imports_controller.rb
+++ b/app/controllers/cohort_imports_controller.rb
@@ -1,10 +1,7 @@
 # frozen_string_literal: true
 
 class CohortImportsController < ApplicationController
-  skip_after_action :verify_policy_scoped, only: %i[new create success]
-
-  before_action :set_team, only: %i[new create]
-  before_action :set_programme, only: %i[new create]
+  before_action :set_programme
 
   def new
     @cohort_import = CohortImport.new
@@ -38,15 +35,11 @@ class CohortImportsController < ApplicationController
 
   private
 
+  def set_programme
+    @programme = policy_scope(Programme).active.find(params[:programme_id])
+  end
+
   def cohort_import_params
     params.fetch(:cohort_import, {}).permit(:csv)
-  end
-
-  def set_team
-    @team = current_user.team
-  end
-
-  def set_programme
-    @programme = policy_scope(Programme).find(params[:programme_id])
   end
 end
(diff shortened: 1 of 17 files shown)


app/controllers/cohort_imports_controller.rb
```diff
@@ -28,25 +27,43 @@ def create
 
     @cohort_import.process!
 
-    @cohort_import.record!
+    if @cohort_import.processed_only_exact_duplicates?
+      render :duplicates
+      return
```
tvararu, line 32: If you want to be fancy:

```suggestion
      render :duplicates and return
```

app/controllers/cohort_imports_controller.rb
```diff
@@ -28,25 +27,43 @@ def create
...
-  def cohort_import_params
-    params.fetch(:cohort_import, {}).permit(:csv)
+  def set_programme
+    @programme = policy_scope(Programme).active.find(params[:programme_id])
   end
 
-  def set_team
-    @team = current_user.team
+  def set_cohort_import
+    # TODO: @programme.cohort_imports.find
```
tvararu, line 58: Follow-up?

app/controllers/cohort_imports_controller.rb
```diff
@@ -28,25 +27,43 @@ def create
...
   end
 
-  def set_programme
-    @programme = policy_scope(Programme).find(params[:programme_id])
+  def set_patients
+    @patients = @cohort_import.patients
+  end
+
+  def cohort_import_params
+    params.fetch(:cohort_import, {}).permit(:csv)
```
tvararu, line 67: Why can't we `.require(:cohort_import)` here? I have a similar issue on the ImmunisationImport flow and might have to do the same.

tvararu (approved):

app/controllers/cohort_imports_controller.rb
thomasleese: Yeah, although I'm going to wait until we've settled on the correct data model for cohorts. I think there will be a relationship between cohort imports and programmes, but not entirely sure yet.

app/controllers/cohort_imports_controller.rb
thomasleese: Good point, I think we probably can't use `require`.

PR #1734, opened 2024-09-11
Refactor email sending

thomasleese:
This refactors how the emails are sent to simplify the mailer classes and tests.

Specifically it adds a `app_template_mail` method which takes a template name and automatically gets the template ID for it, and updates the `be_sent_with_govuk_notify` matcher to again take a template name. It also adds more uses of the `expect_email_to` expectation helper.

I've also refactored the `opts` method to allow the user of `super` rather than needing to re-create functionality.

This is part one of a number of follow up PRs related to consolidating emails and text reminders, and then finally implementing text reminders.


benilovj (approved):

tvararu:
Nice refactor!

PR #1736, opened 2024-09-12
Add ParentRelationship model

thomasleese:
This represents the relationship between a parent and a patient with a particular type. This is needed so we can model the parent and patients as having a many to many relationship, rather than currently with a one to many.

This adds the model but doesn't yet connect it up to the parent or patient model as we first need to refactor the consent form model to no longer link to a parent directly.


spec/models/parent_spec.rb
```diff
@@ -17,7 +17,7 @@
 #  updated_at           :datetime         not null
 #
 
-describe Parent do
+describe Parent, type: :model do
```
tvararu, line 20: Isn't the type inferred from the file location?

tvararu (approved):

spec/models/parent_spec.rb
thomasleese: Oh, well, I've learnt something new today!

PR #1738, opened 2024-09-13
Add many-to-many relationship between patients and parents

thomasleese:
This follows up from #1736 and #1737 to finish the migration to have a many-to-many relationship between patients and parents. This removes the relationship fields from parents, and instead creates `ParentRelationship` records. The behaviour of the service should be the same before and after this change. See individual commits for more details.


db/migrate/20240912193815_remove_relationship_from_parents.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class RemoveRelationshipFromParents < ActiveRecord::Migration[7.2]
+  def change
+    change_table :parents, bulk: true do |t|
+      t.remove :relationship, type: :integer
+      t.remove :relationship_other, type: :string
```
tvararu, line 7: These are harmless but I don't think they do anything:

```suggestion
      t.remove :relationship
      t.remove :relationship_other
```

spec/factories/parents.rb
```diff
@@ -11,32 +11,19 @@
...
 #  recorded_at                  :datetime
-#  relationship                 :integer
-#  relationship_other           :string
 #  created_at                   :datetime         not null
 #  updated_at                   :datetime         not null
 #
 FactoryBot.define do
   factory :parent do
     transient do
-      random { Random.new }
```
tvararu, line 22: Think we had a reason to allow passing in a transient `random:` attribute...

tvararu (approved):

spec/factories/parents.rb
thomasleese: I did wonder about that, but then I couldn't see it was being used anywhere so I figured we could put it back if we needed to.

db/migrate/20240912193815_remove_relationship_from_parents.rb
tvararu: Oh fair

PR #1770, opened 2024-09-17
Show import issues

tvararu:
If a user does not address duplicate issues when uploading, we want to display the upload issues in a separate section so that they can action them at a later point.

TODO:

- [x] Add import issues section button to all secondary navigations
- [x] Show amount of import issues with badge
- [x] Implement show/update logic
- [x] Uncomment `.strict_loading`

#### New tab

image

#### With records

image

#### Reviewing

image

#### Submitting the form

image


app/components/app_programme_navigation_component.rb
thomasleese, line 1: Nice!

app/controllers/import_issues_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+class ImportIssuesController < ApplicationController
+  before_action :set_programme
+  before_action :set_vaccination_record, only: %i[show update]
+  before_action :set_form, only: %i[show update]
+
+  def index
+    @import_issues = @programme.import_issues
+
+    render layout: "full"
```
thomasleese, line 11: Here we can set `layout "full"` on the controller since every action uses this layout.

db/migrate/20240918084709_add_default_to_pending_changes.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class AddDefaultToPendingChanges < ActiveRecord::Migration[7.2]
+  def up
+    change_column_default :vaccination_records, :pending_changes, {}
+
+    VaccinationRecord.where(pending_changes: nil).update_all(
+      pending_changes: {
+      }
+    )
+  end
```
thomasleese, line 11: Can we make this `not null` at the same time?

thomasleese (approved):

PR #1775, opened 2024-09-18
Show cohorts

thomasleese:
This adds a new page for showing the cohorts attached to a particular programme (via the team) in a grid, allowing the user to then select each cohort and see the children within.

## Screenshot

Screenshot 2024-09-18 at 10 41 32


tvararu (approved):
You'll have conflicts for the secondary navigation items, but otherwise 👍

PR #1785, opened 2024-09-18
Remove parent_contact_method feature flag

benilovj:
The parental consent form now supports texts, so we can enable this step of the journey for all users.

### DIFF
--- a/spec/models/consent_form_spec.rb
+++ b/spec/models/consent_form_spec.rb
@@ -738,9 +738,33 @@ describe ConsentForm do
     expect(consent_form.health_answers).not_to be_empty
   end
 
-  it "removes the health questions when the parent refuses consent" do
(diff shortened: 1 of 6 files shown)


spec/models/consent_form_spec.rb
```diff
@@ -738,9 +738,33 @@
     expect(consent_form.health_answers).not_to be_empty
   end
 
-  it "removes the health questions when the parent refuses consent" do
```
benilovj, line 741: This spec started failing when the feature flag was removed; I'm a bit unclear how it was passing previously when the session (which is randomly flu or hpv, if I'm reading the factory code correctly) was a flu session.

I've split the spec so there is one for each type of programme.

benilovj: Ah, this spec is flaky for me on `main`.

tvararu (approved):

PR #1791, opened 2024-09-19
Immunisation import upload fixes

tvararu:
Fixes:

- [x] Fixes the review section breaking due to `strict_loading` and missing `includes(:patient_session)`
- [x] Adds `changed_record_count` to ImmunisationImport to track the number of records with `pending_changes`
- [x] Fixes layout on show action to be full width
- [x] Fix records with no pending changes appearing in Import issues
- [x] Fix title of uploads section (imports)
- [x] Miscellaneous others (see commits)


app/models/immunisation_import_row.rb
```diff
@@ -89,8 +89,8 @@ def to_vaccination_record
     if vaccination_record.persisted?
       vaccination_record.stage_changes(
         batch_id: batch.id,
-        delivery_method:,
-        delivery_site:
+        delivery_method: delivery_method.to_s,
+        delivery_site: delivery_site.to_s
```
thomasleese, line 93: I wonder if it makes sense to change the `delivery_method` and `delivery_site` methods themselves to return the value as a string so it's the same everywhere.

Since the model instance returns them as strings, this might avoid any similar issues in the future.

Gemfile
```diff
@@ -12,6 +12,7 @@ gem "awesome_print"
 gem "bootsnap", require: false
 gem "config"
 gem "cssbundling-rails"
+gem "csv"
```
thomasleese, line 15: Does this mean we can remove `require "csv"` across the app?

app/models/immunisation_import_row.rb
tvararu: Thought about doing that instead 👍

Gemfile
tvararu: Think so!

thomasleese (approved):

PR #1795, opened 2024-09-19
Ensure patients are part of current programme

thomasleese:
When importing patient records we want to be sure that the patient belongs to one of the current programme's year groups. This is necessary as we don't currently have a way of showing patients not attached to a current cohort.

We expect this is will be a temporary validation for the first release as it doesn't allow users to upload historical records, or work very well with cohorts being agonistic of programmes.

To support this I have refactored the `Cohort` model in a way that I think makes it easier to understand going forward.


tvararu (approved):
> being agonistic of programmes

agnostic*, though agonistic is a great word!

thomasleese:
> agnostic*, though agonistic is a great word!

I didn't even know that word exists!

PR #1797, opened 2024-09-20
Replace locations radio list with autocomplete

tvararu:
This replaces the list of radio buttons with a select that is progressively enhanced to an autocomplete when JS is enabled.

Screenshot 2024-09-23 at 11 58 52

### DIFF
--- a/app/controllers/sessions/edit_controller.rb
+++ b/app/controllers/sessions/edit_controller.rb
@@ -98,7 +98,11 @@ class Sessions::EditController < ApplicationController
 
   def set_locations
     # TODO: Use an autocomplete instead.
-    @locations = policy_scope(Location).order(:name).limit(100)
+    @locations =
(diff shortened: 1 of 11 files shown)


app/controllers/sessions/edit_controller.rb
```diff
@@ -98,7 +98,11 @@ def set_steps
 
   def set_locations
     # TODO: Use an autocomplete instead.
-    @locations = policy_scope(Location).order(:name).limit(100)
+    @locations =
+      policy_scope(Location)
+        .order(:name)
+        .limit(100)
```
thomasleese, line 104: Do we still need the `limit`?

thomasleese: Fair enough, I wondered if that was still the case.

thomasleese (approved):

PR #1817, opened 2024-09-24
Update ImmunisationImportDuplicateForm to support just patients

thomasleese:
To build the duplicate matching page for patients we need to update the duplicate form to handle both patients and vaccination records.


tvararu (approved):
Very cool use of locale strings!

PR #1822, opened 2024-09-24
Make cohort import upload quicker

tvararu:
Use batch processing via activerecord-import.


tvararu:
Closing for now but this PR will continue to be available to analyze/incorporate fixes from it in the future should we desire.

PR #1826, opened 2024-09-24
Simplify set_patients query

tvararu:
Instead of arel, we can use native AR queries.

Also two small unrelated changes.

### DIFF
--- a/app/controllers/sessions/edit_controller.rb
+++ b/app/controllers/sessions/edit_controller.rb
@@ -92,7 +92,7 @@ class Sessions::EditController < ApplicationController
   end
 
   def set_locations
-    # TODO: Use an autocomplete instead.
+    # TODO: Don't use limit.
     @locations =
       policy_scope(Location)
         .order(:name)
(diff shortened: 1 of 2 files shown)


app/controllers/sessions/edit_controller.rb
```diff
@@ -104,20 +104,17 @@ def set_patients
...
-            .joins(:patient_sessions)
-            .active
-            .where(programme: @session.programme)
-            .where("patient_sessions.patient_id = patients.id")
-            .arel
-            .exists
-        )
-        .sort_by(&:last_name)
+        .where.not(id: patients_in_active_sessions_and_same_programme)
+        .order(:last_name)
```
tvararu, line 117: @thomasleese I haven't tested this, I probably should, but is it possible that the reason we had `sort_by` is because `order` won't work due to encryption?

thomasleese: Hmm, yeah, that's very possible, I'm not actually sure myself why we would've used `sort_by` here and not `order`.

tvararu: Confirmed that `order` does not work.

PR #1832, opened 2024-09-25
Remove Session date column

thomasleese:
Sessions no longer have a single date, instead they can run over multiple dates. This change supports the underlying database change, while maintaining the current functionality.

Further work will be required to update the UI to allow the user to pick multiple dates, updating consent to handle multiple session dates, and updating the reminder and other emails.


app/models/immunisation_import_row.rb
```diff
@@ -118,17 +118,32 @@ def session
...
+          session =
+            Session.create!(
+              active: false,
+              team:,
+              location:,
+              time_of_day: :all_day
+            )
+          session.dates.create!(value: session_date)
+          session.programmes << @programme
+          session
```
thomasleese, line 144: I expect this will be rewritten as we work on sessions more, specifically we can probably add dates to existing sessions rather than creating new ones.

tvararu (approved):

PR #1837, opened 2024-09-25
Add accessibility end to end playwright test

tvararu:
This test logs into the app and checks a few pages for easy to find accessibility issues using Axe.


tvararu:
The hard-coded secret GitGuardian is referring to is `postgres://postgres:postgres@localhost:5432/manage_vaccinations_test`.

thomasleese (approved):

PR #1839, opened 2024-09-25
Remove Session#active

thomasleese:
Following up from #1838 this removes the concept of an "active" session and replaces it with sessions that have dates, and where the dates are within today's date.

This matches the new user interface where sessions implicitly exist (for each team school and programme that matches the year groups of that school) and only the dates need to be set before the sessions moved from the "Unplanned" tab to one of the others.


paulrobertlloyd:
I'd hold off on this for a bit. After a design catch up today, we decided to keep the previous concept of 'todays sessions', that is, one of the session dates is today (and I'm going to make it so that the prototype actually works like that, now we have multiple dates). Will be updating the prototype shortly.

tvararu (approved):

PR #1844, opened 2024-09-26
Create sessions from schools

thomasleese:
This updates the journey for creating sessions by removing the "Add a new session" button and replacing it with the "Unscheduled" tab. When a user clicks on one of these schools, an unscheduled session is automatically created for it and the user is redirected to the session page.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -2,19 +2,16 @@
 
 class SessionsController < ApplicationController
   before_action :set_session,
-                except: %i[index scheduled unscheduled completed create]
+                except: %i[new index scheduled unscheduled completed]
 
-  def create
-    skip_policy_scope
+  skip_after_action :verify_policy_scoped, only: :new
 
-    @session =
-      Session.create!(
-        academic_year: Date.current.academic_year,
-        team:,
-        programmes: team.programmes
-      )
+  def new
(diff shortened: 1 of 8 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -2,19 +2,16 @@
...
-    @session =
-      Session.create!(
-        academic_year: Date.current.academic_year,
-        team:,
-        programmes: team.programmes
-      )
+  def new
+    location = team.schools.find(params[:location_id])
+
+    session = Session.find_or_create_by!(team:, academic_year:, location:)
```
misaka, line 12: Surprised to see a create in a GET request. I suppose it's ok because it's idempotent?

thomasleese: Discussed in the office, decided to ship it like this, it's perhaps not ideal but keeps things simple for the other pages down the line related to sessions, and I will follow up this PR with a unique constraint to avoid race conditions creating multiple sessions for the same school.

misaka (approved):

app/controllers/sessions_controller.rb
thomasleese: Index added in https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1846.

PR #1857, opened 2024-09-27
Create TeamProgramme

thomasleese:
This adds a new model for representing the relationship between teams and programmes. We're going to change programmes to be global (to make it easier to compare vaccination records between different teams) so we need a join table to link programmes with teams.

The follow up work will be to remove the `team_id` column from programmes and use this join table instead.


tvararu (approved):

thomasleese:
It looks like we might not need this for the moment as we haven't come across any examples of a team not administering a programme, I'll leave it open for the time being while I work on making programmes global and see if a need arises.

PR #1866, opened 2024-09-27
Attach vaccines to programmes directly

thomasleese:
This adds a programme column to vaccines, to change the relationship between programmes and vaccines to be one to many rather than many to many.


tvararu (commented):
Can you also update `lib/tasks/add_health_questions.rake`?

thomasleese:
> Can you also update `lib/tasks/add_health_questions.rake`?

Done, it takes a programme type now instead of a team ID (this might need to change if we decide health questions need to be specific to teams in the future).

lib/tasks/add_health_questions.rake
```diff
@@ -6,24 +6,20 @@ desc <<-DESC
...
-     %i[team_id vaccine_id replace] => :environment do |_task, args|
-  team = Team.find(args[:team_id])
-  vaccine =
-    team
-      .programmes
-      .flat_map(&:vaccines)
-      .find { |v| v.id == args[:vaccine_id].to_i }
+     %i[programme vaccine_id replace] => :environment do |_task, args|
+  programme = Programme.find_by!(type: args[:programme])
+  vaccine = programme.vaccines.find_by(id: args[:vaccine_id])
```
tvararu, line 22: I think this breaks as there are references to `team` further down, and the `raise` below needs to be updated too as we don't pass in `team` anymore

thomasleese: Yep, I pushed a little too quickly there, should be working now.

PR #1882, opened 2024-10-01
Add dev tools endpoint for resetting a team's data

benilovj:
This will make it easier to run repeatable black box tests of the app.

I'm not convinced this is the best way to implement it – open to suggestions on alternatives.

### DIFF
--- a/app/controllers/dev_controller.rb
+++ b/app/controllers/dev_controller.rb
@@ -16,6 +16,63 @@ class DevController < ApplicationController
     redirect_to root_path
   end
 
+  def reset_team
+    team = Team.find_by(ods_code: params[:team_ods_code])
+
+    cohort_imports = CohortImport.where(team:)
+    cohort_imports.find_each do |cohort_import|
+      cohort_import.parent_relationships.clear
+      cohort_import.patients.clear
+      cohort_import.parents.clear
+
+      cohort_import.destroy!
+    end
+
+    immunisation_imports = ImmunisationImport.where(team:)
+    immunisation_imports.find_each do |immunisation_import|
+      immunisation_import.batches.clear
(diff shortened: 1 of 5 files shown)


app/controllers/dev_controller.rb
```diff
@@ -16,6 +16,63 @@ def reset
...
+      .joins(:cohort)
+      .where(cohorts: { team: })
+      .distinct
+      .find_each do |patient|
+        parent_ids = patient.parent_ids
+        patient.parents.clear
+        parent_ids.each { |parent_id| Parent.find(parent_id).destroy }
+
+        patient.save!
+        patient.destroy!
```
thomasleese, line 68: I wonder if it's worth setting up the `has_many`s in a way that allows us to do `patient.parents.destroy_all` and then `patient.destroy!`, I think that would read better here, but not a big deal.

benilovj: Like this? [200211e](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1882/commits/200211efeafbeab7b7bef47a7675990c4ede1323)

tvararu (approved):

app/controllers/dev_controller.rb
thomasleese: That works, although I was thinking we could do something like:

```rb
# models/patient.rb
has_many :parent_relationships, dependent: :destroy
```

```rb
patients.destroy_all
```

```rb
# Delete childless parents
Parent.joins(:patients).where(patients: { id: nil }).destroy_all
```

benilovj: I've left it as is for now just to keep things moving, but not precious about it if anyone changes it afterwards.

PR #1893, opened 2024-10-01
Display row errors on cohort import edit page

tvararu:
This moves the `parse_rows!` action to a background job and stores errors in a jsonb field to display them to the user at a later point.

Adds `status` and `serialized_errors` fields to Cohort and Immunisation imports to store the results of the background job.


thomasleese (approved):
Nice!

PR #1901, opened 2024-10-02
Remove unused state transition methods

thomasleese:
The state functionality was partially removed in
8200353266453e7728cbdc76cdb2e6a35c7bc05f but the transition methods weren't removed. They're not being used anywhere and don't do anything so we can safely remove them.


misaka (approved):

misaka:
🪓

PR #1902, opened 2024-10-02
Decouple consent forms from sessions

thomasleese:
This refactors the consent forms to decouple them from specific sessions, instead linking them to sessions via the location, programme and team.

This is necessary as we want to allow users to submit consent forms for a location different to the session they received the consent request for.

If we receive a consent form for a patient that we can match in the current session, but the parent says they go to a different school, we can move the patient across to the other school, and if appropriate, move them to a different vaccination session under the same programme and team.


tvararu (approved):
Good change. The school moving logic seems like a sensible start and can be iterated if it doesn't fit all requirements 👍

PR #1919, opened 2024-10-05
Support CIS2 backchannel logout

misaka:
This is required by CIS2 technical conformance. It adds an callback `/auth/cis2_logout` which CIS2 will use to logout a user session.

Testing this in a real environment is tricky, it's not even clear yet how to get this endpoint registered with CIS2.


app/lib/nhs/cis2.rb
```diff
@@ -0,0 +1,45 @@
...
+            Rails.cache.delete("cis2:jwks")
+          end
+          Rails
+            .cache
+            .fetch("cis2:jwks") do
+              @cache_last_update = Time.zone.now.to_i
+              jwks_hash = JSON.parse(Faraday.get(jwks_uri).body)
+              jwks = JWT::JWK::Set.new(jwks_hash)
+              jwks.select! { |key| key[:use] == "sig" }
+              jwks
```
thomasleese, line 22:
```suggestion
              jwks.select { |key| key[:use] == "sig" }
```

thomasleese (approved):
Just one small comment, but otherwise looks good.

app/lib/nhs/cis2.rb
misaka: Ya, agreed, that makes more sense.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -2,9 +2,11 @@
 
 class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   include AuthenticationConcern
+  include CIS2LogoutConcern
 
   skip_before_action :authenticate_user!
   skip_after_action :verify_policy_scoped
+  skip_before_action :verify_authenticity_token, only: [:cis2_logout]
```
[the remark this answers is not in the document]
misaka: Dismissed because we validate this request by verifying the logout JWT provided.

tvararu (approved):

PR #1945, opened 2024-10-08
Add sorting to session and cohort views

benilovj:
* sessions with dates are sorted by earliest date and location name
* sessions without dates are sorted by location name
* cohorts are sorted by patient first and last names

### DIFF
--- a/app/controllers/programmes_controller.rb
+++ b/app/controllers/programmes_controller.rb
@@ -13,9 +13,17 @@ class ProgrammesController < ApplicationController
   end
 
   def sessions
-    sessions_for_programme = policy_scope(Session).has_programme(@programme)
+    sessions_for_programme =
+      policy_scope(Session).has_programme(@programme).includes(
+        :dates,
+        :location
+      )
 
-    @scheduled_sessions = sessions_for_programme.scheduled
+    @scheduled_sessions =
+      sessions_for_programme.scheduled.sort_by do |session|
+        [session.dates.first.value, session.location&.name]
+      end
+
     @unscheduled_sessions = sessions_for_programme.unscheduled
     @completed_sessions = sessions_for_programme.completed
   end
--- a/app/controllers/programmes_controller.rb
+++ b/app/controllers/programmes_controller.rb
@@ -26,7 +26,10 @@ class ProgrammesController < ApplicationController
 
     @unscheduled_sessions =
       sessions_for_programme.unscheduled.order_by_location_name
-    @completed_sessions = sessions_for_programme.completed
+    @completed_sessions =
+      sessions_for_programme.completed.sort_by do |session|
+        [session.dates.first.value, session.location&.name]
+      end
   end
 
   private
--- a/app/controllers/programmes_controller.rb
+++ b/app/controllers/programmes_controller.rb
@@ -24,7 +24,8 @@ class ProgrammesController < ApplicationController
         [session.dates.first.value, session.location&.name]
       end
 
-    @unscheduled_sessions = sessions_for_programme.unscheduled
+    @unscheduled_sessions =
+      sessions_for_programme.unscheduled.order_by_location_name
     @completed_sessions = sessions_for_programme.completed
   end
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -10,7 +10,13 @@ class SessionsController < ApplicationController
   end
 
   def scheduled
-    @sessions = policy_scope(Session).scheduled
+    @sessions =
+      policy_scope(Session)
+        .includes(:dates, :location, :programmes)
+        .scheduled
+        .sort_by do |session|
+          [session.dates.first.value, session.location&.name]
+        end
 
     render layout: "full"
   end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -33,7 +33,13 @@ class SessionsController < ApplicationController
   end
 
   def completed
-    @sessions = policy_scope(Session).completed
+    @sessions =
+      policy_scope(Session)
+        .includes(:dates, :location, :programmes)
+        .completed
+        .sort_by do |session|
+          [session.dates.first.value, session.location&.name]
+        end
 
     render layout: "full"
   end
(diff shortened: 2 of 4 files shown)


app/controllers/programmes_controller.rb
```diff
@@ -14,9 +14,16 @@ def show
...
+    sessions_for_programme =
+      policy_scope(Session)
+        .has_programme(@programme)
+        .strict_loading
+        .includes(:dates, :location)
 
-    @scheduled_sessions = sessions_for_programme.scheduled
+    @scheduled_sessions =
+      sessions_for_programme.scheduled.sort_by do |session|
+        [session.dates.first.value, session.location.name]
```
thomasleese, line 25: The location can be `nil` if we import a session for an unknown school so I think we'll want to handle that here.

app/controllers/sessions_controller.rb
```diff
@@ -21,31 +21,55 @@ def new
   end
 
   def index
-    @sessions = policy_scope(Session).today
+    @sessions =
+      policy_scope(Session)
+        .strict_loading
+        .includes(:dates, :location, :programmes)
+        .today
+        .sort_by { |session| session.location.name }
```
thomasleese, line 29: Could we do the sorting at the database level to improve performance or maybe even have a default ordering?

benilovj: Do you mean something like this?

```
diff --git a/app/controllers/sessions_controller.rb b/app/controllers/sessions_controller.rb
index ba43c7d3..2249e100 100644
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -26,7 +26,6 @@ class SessionsController < ApplicationController
         .strict_loading
         .includes(:dates, :location, :programmes)
         .today
-        .sort_by { |session| session.location.name }

     render layout: "full"
   end
diff --git a/app/models/session.rb b/app/models/session.rb
index eec1212a..e0422f28 100644
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -50,7 +50,12 @@ class Session < ApplicationRecord
   scope :has_programme,
         ->(programme) { joins(:programmes).where(programmes: programme) }

-  scope :today, -> { has_date(Date.current) }
+  scope :today,
+        -> do
+          has_date(Date.current).left_joins(:location).order(
+            "locations.name ASC NULLS LAST"
+          )
+        end
```

thomasleese: Yeah something like that looks good to me!

app/controllers/programmes_controller.rb
benilovj: Updated in https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1945/commits/5806f935e4bc7c9a04b2de0fbb18a9f45074909f to handle `location = nil`

app/controllers/sessions_controller.rb
benilovj: 👍 Addressed in https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1945/commits/ee4ef13075fc6c3d08629b3b0e5a5868a4d7baec

tvararu (approved):

app/controllers/programmes_controller.rb
```diff
@@ -27,8 +34,12 @@ def sessions
           .map do |location|
             Session.new(team: current_user.team, location:, academic_year:)
           end
+    @unscheduled_sessions.sort_by! { |session| session.location.name }
```
thomasleese, line 37:
```suggestion
    @unscheduled_sessions.sort_by! { |session| session.location&.name }
```

app/controllers/programmes_controller.rb
```diff
@@ -27,8 +34,12 @@ def sessions
...
 
-    @completed_sessions = sessions_for_programme.completed
+    @completed_sessions =
+      sessions_for_programme.completed.sort_by do |session|
+        [session.dates.first.value, session.location.name]
```
thomasleese, line 41:
```suggestion
        [session.dates.first.value, session.location&.name]
```

app/controllers/sessions_controller.rb
```diff
@@ -21,31 +21,54 @@ def new
...
 
   def scheduled
-    @sessions = policy_scope(Session).scheduled
+    @sessions =
+      policy_scope(Session)
+        .strict_loading
+        .includes(:dates, :location, :programmes)
+        .scheduled
+        .sort_by do |session|
+          [session.dates.first.value, session.location.name]
```
thomasleese, line 40:
```suggestion
          [session.dates.first.value, session.location&.name]
```

app/controllers/sessions_controller.rb
```diff
@@ -21,31 +21,54 @@ def new
...
 
   def completed
-    @sessions = policy_scope(Session).completed
+    @sessions =
+      policy_scope(Session)
+        .strict_loading
+        .includes(:dates, :location, :programmes)
+        .completed
+        .sort_by do |session|
+          [session.dates.first.value, session.location.name]
```
thomasleese, line 70:
```suggestion
          [session.dates.first.value, session.location&.name]
```

app/controllers/sessions_controller.rb
```diff
@@ -21,31 +21,54 @@ def new
...
+        .strict_loading
+        .includes(:dates, :location, :programmes)
+        .unscheduled +
         team
           .schools
           .for_year_groups(team.year_groups)
           .has_no_session
           .map { |location| Session.new(team:, location:, academic_year:) }
 
+    @sessions.sort_by! { |session| session.location.name }
```
thomasleese, line 58:
```suggestion
    @sessions.sort_by! { |session| session.location&.name }
```

app/controllers/programmes_controller.rb
benilovj: Ugh, apologies for leaving the inconsistencies in.

Should be fixed now everywhere now.

thomasleese (approved):

app/controllers/programmes_controller.rb
benilovj: Fixed now. I debated whether to also push this sort into the database but held off for now for expediency. We can probably sweet it up when we revisit this code or if we discover it's not performant.

thomasleese: Yeah, that makes sense!

PR #1962, opened 2024-10-09
Include schools in patients table

tvararu:
Add an extra row and make the text slightly smaller.

Screenshot 2024-10-09 at 18 32 10


app/components/app_patient_table_component.html.erb
```diff
@@ -32,6 +35,11 @@
             <span class="nhsuk-table-responsive__heading">Postcode</span>
             <%= patient.address_postcode %>
           <% end %>
+
+          <% row.with_cell do %>
+            <span class="nhsuk-table-responsive__heading">School</span>
+            <%= patient.school.name %>
```
thomasleese, line 41:
```suggestion
            <%= patient_school(patient) %>
```

`school` can be `nil` if we don't know the school, or they're home schooled.

thomasleese (approved):

PR #1965, opened 2024-10-09
Bulk import records link optimisations

misaka:
Provides further performance optimisations to the cohort import process.

These should be usable by other classes that use CSVImportable, with a
little modification.

Also adds an index on `parents.email` that helps optimise performance.

Time to import 40k cohorts file on my local system:
- With only bulk linked records change: ~16m
- Add `parents.email` index: 6m20s

Still missing:
- [ ] specs for new `CSVImportable` methods


tvararu (approved):

thomasleese (approved):
Nice!

thomasleese:
Merging in to get the performance improvements in, we can add the tests and apply this to vaccination records later.

PR #1978, opened 2024-10-10
Refactor how patients are moved between sessions

thomasleese:
This refactors how we handle patients moving between schools to remove the logic out of a callback on the patient model and instead apply it explicitly where we need it. This has a number of advantages:

- It improves the performance of the import as we no longer need to run a callback for each patient and instead can use the `create_patient_sessions!` method which handles this in bulk.
- It fixes a bug where importing vaccination records and choosing a duplicate where the school is different leads to the patient being added to a upcoming session regardless of their vaccination status.
- It ensures the cohorts are updated correctly (this is a replacement to #1974).


tvararu (commented):
Nice!

tvararu (approved):

PR #1989, opened 2024-10-11
Add scripts to generate fake data

thomasleese:
This adds a number of scripts, initially built for the model office, but is useful for general testing as it allows us to generate large amounts of test data. I haven't written any tests for the scripts for now, but if we find we're using them more and more we can start to improve the quality of the code further.

I've also improved our postcode generation so we will now have random UK valid postcodes generated in our factories.


tvararu:
Instead of `bin`, could you create and add these to a `scripts` folder? This follows a new Rails 8 convention: https://blog.appsignal.com/2024/10/07/whats-new-in-ruby-on-rails-8.html

thomasleese:
> Instead of `bin`, could you create and add these to a `scripts` folder? This follows a new Rails 8 convention: https://blog.appsignal.com/2024/10/07/whats-new-in-ruby-on-rails-8.html

I've pushed up amended commits that puts the files in `script`.

benilovj (approved):

PR #1997, opened 2024-10-14
Simplify users

thomasleese:
This removes some functionality around users that we won't need since our users will be signing in via CIS2 and user management is handled by that service. See individual commits for what's been removed.


tvararu (approved):
:scissors:

@misaka can you review this as well?

misaka (approved):

PR #1999, opened 2024-10-14
Disable passwords if CIS2 is enabled

thomasleese:
If a user was created via CIS2 we don't need to set a password on the user, which makes it impossible to sign in as that user by guessing/bruteforcing the random password that was previously set on the user.


tvararu (approved):

app/models/user.rb
```diff
@@ -50,6 +49,15 @@ class User < ApplicationRecord
...
 
+  with_options unless: -> { Flipper.enabled?(:cis2) } do
+    validates :password,
+              presence: true,
+              confirmation: true,
+              length: {
+                within: 6..128
+              }
+  end
+
```
misaka, line 60: Would it make more sense to validate the password if the user didn't come from CIS2? Using the feature flag means that invalid users (without a password) can be created while CIS2 is enabled, and then if we disable the flag they'll be invalid. This might not happen in a production env, but in a dev / testing env it could be a nuisance.

thomasleese: Yep I think that's a good idea.

thomasleese: I've amended the commit with the logic you've suggested.

misaka (approved):

PR #2017, opened 2024-10-15
Accept CHILD_GENDER attribute for class and cohort imports

benilovj:
It matches the allowed values for `Patient#gender_code`

### DIFF
--- a/app/models/patient_import_row.rb
+++ b/app/models/patient_import_row.rb
@@ -8,6 +8,7 @@ class PatientImportRow
   validates :first_name, presence: true
   validates :last_name, presence: true
   validates :nhs_number, length: { is: 10 }, allow_blank: true
+  validates :gender_code, inclusion: { in: Patient.gender_codes.keys }
   validates :year_group,
             inclusion: {
               in: :year_groups
@@ -78,6 +79,7 @@ class PatientImportRow
       common_name:,
       date_of_birth:,
       first_name:,
+      gender_code:,
       home_educated:,
       last_name:,
       nhs_number:,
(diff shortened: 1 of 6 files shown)


app/models/patient_import_row.rb
```diff
@@ -8,6 +8,11 @@ class PatientImportRow
   validates :first_name, presence: true
   validates :last_name, presence: true
   validates :nhs_number, length: { is: 10 }, allow_blank: true
+  validates :gender_code,
+            inclusion: {
+              in: Patient.gender_codes.keys
+            },
+            allow_blank: true
```
thomasleese, line 15: I think we don't want to allow blank here, since we've got a default in the `def gender_code` method anyway it'll always have a value.

app/models/patient_import_row.rb
```diff
@@ -131,6 +137,10 @@ def date_of_birth
     nil
   end
 
+  def gender_code
+    @data["CHILD_GENDER"]&.strip&.downcase&.gsub(" ", "_") || "not_specified"
```
thomasleese, line 141:
```suggestion
    @data["CHILD_GENDER"]&.strip&.downcase&.gsub(" ", "_") || "not_known"
```

Should this default to `not_known` rather than not specified?

app/models/patient_import_row.rb
benilovj: Agreed, I've removed the `allow_blank`.

app/models/patient_import_row.rb
benilovj: Yes, looking at the definitions more closely, you're right. I've changed this.

thomasleese (approved):

PR #2018, opened 2024-10-15
Don't always remove patients from sessions

thomasleese:
If the patient session has been interacted with in some way (either consent received, or triaged, or even vaccinated) we can't safely remove the patient from the session as it's unclear what state they're in.

https://good-machine.sentry.io/issues/5990842901


tvararu (approved):

thomasleese:
We're not quite sure about the logic here so moving this to draft.

benilovj (approved):

PR #2028, opened 2024-10-15
Tighten up batch expiry validations in frontend and in the file import

benilovj:


app/models/batch.rb
```diff
@@ -40,6 +40,7 @@ class Batch < ApplicationRecord
   validates :expiry,
             presence: true,
             comparison: {
-              greater_than_or_equal_to: Date.new(2000, 1, 1)
+              greater_than: -> { Date.new(Date.current.year - 15, 1, 1) },
+              less_than: -> { Date.new(Date.current.year + 15, 1, 1) }
```
thomasleese, line 44: We might want an `on: :create` here because it'll make it impossible to update other things about batches in 15 years, but that's possibly not something to worry about...

thomasleese (approved):

app/models/batch.rb
benilovj: I think we'd want to also validate this on edit as well.
If it becomes a problem in 15 years, we can probably loosen the validation 😂

PR #2035, opened 2024-10-16
Patient page title: use patient initials instead of full name

benilovj:
This is to avoid the patient name appearing in browser histories

### DIFF
--- a/app/views/patients/show.html.erb
+++ b/app/views/patients/show.html.erb
@@ -5,7 +5,7 @@
       ) %>
 <% end %>
 
-<%= h1 page_title: @patient.full_name do %>
+<%= h1 page_title: @patient.initials do %>
   <span class="nhsuk-caption-l"><%= patient_school(@patient) %></span>
   <%= @patient.full_name %>
 <% end %>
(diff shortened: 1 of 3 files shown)


tvararu (approved):

app/models/patient.rb
```diff
@@ -131,6 +131,10 @@ def relationship_to(parent:)
     parent_relationships.find { _1.parent == parent }
   end
 
+  def initials
```
thomasleese, line 134: Doesn't really matter, but might be nice to put this in the `FullNameConcern` since that relates to formatting names (and although unlikely to be used allows this method to exist for consent forms and users as well).

app/views/patients/show.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :page_title, @patient.initials %>
```
thomasleese, line 1: On the `h1` line below you can change the `page_title:` attribute rather than needing this line.

app/models/patient.rb
benilovj: done

app/views/patients/show.html.erb
benilovj: done

thomasleese (approved):

PR #2045, opened 2024-10-16
Look up NHS numbers when matching consent forms

thomasleese:
When matching a consent form with a patient, we can look up an NHS number using the patient details provided in the consent form, which makes it more likely that we'll successfully match a consent form with a patient and with increased accuracy.

### DIFF
--- a/app/jobs/consent_form_matching_job.rb
+++ b/app/jobs/consent_form_matching_job.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ConsentFormMatchingJob < ApplicationJob
-  queue_as :default
+  include NHSAPIConcurrencyConcern
 
   def perform(consent_form)
     session = consent_form.scheduled_session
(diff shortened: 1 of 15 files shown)


app/jobs/consent_form_matching_job.rb
```diff
@@ -1,14 +1,21 @@
...
+  include NHSNumberLookupConcern
+
+  queue_as :consents
 
   def perform(consent_form)
     session = consent_form.scheduled_session
 
+    nhs_number = find_nhs_number(consent_form)
+
+    # TODO: What happens if we find an NHS number for a patient,
```
benilovj, line 13: I *think* this scenario is likely to come up when:

* child recently moved to this school, from another in-area school or from out-of-area
* they've not appeared on a class list before, or the school hadn't shared the class list with the SAIS team
* the child's parents got a link to the consent form via the school and filled in the consent

Let's check with the others tomorrow, but I think we should be flagging this person as a mover in, to be confirmed? Then if they're confirmed, then they would automatically be added to the right cohort / session, and the consent form would already be attached?

thomasleese: I've pushed up a commit that implements this logic, although without the flagging step, for now it follows the same automated logic as every other type of school move.

PR #2046, opened 2024-10-16
Fix crash when very invalid date given for batch expiry, add spec coverage

benilovj:
https://good-machine.sentry.io/issues/5995629672/


app/controllers/batches_controller.rb
```diff
@@ -11,7 +11,13 @@ def new
   end
 
   def create
-    @batch = Batch.new(batch_params.merge(team: current_user.team, vaccine:))
+    begin
+      @batch = Batch.new(batch_params.merge(team: current_user.team, vaccine:))
+    rescue ActiveRecord::MultiparameterAssignmentErrors
```
benilovj, line 16: There might be a more elegant way of solving this but it escapes me right now.

thomasleese: I think using form objects and only creating a `Date` object once the year/month/day has been validated is probably the best option, but that would require more changes. I reckon we're going to have this problem across every model that accepts dates, so I wonder how we could apply this across the board.

thomasleese (approved):

app/controllers/batches_controller.rb
benilovj: Just tested it, and the same scenario happens on creating session dates, for instance, but not on the consent form DOB 🤷 . Will investigate more.

tvararu: Handling dates is actually quite annoying and some amount of rescuing is to be expected.

tvararu (approved):

app/controllers/batches_controller.rb
benilovj: I'll merge this change in, but I appreciate it doesn't solve the same problem in other places. I've added follow-up technical work for a more generalised solution into the backlog.

PR #2056, opened 2024-10-17
Add basic global children views

thomasleese:
This adds some initial global children views loosely based on the designs in the prototype, but not fully implementing all the features. 

This is needed so we have a place to put the new "Remove from cohort" button that will be implemented next.

## Screenshots

Screenshot 2024-10-17 at 12 04 27

Screenshot 2024-10-17 at 12 13 55


app/components/app_patient_table_component.html.erb
```diff
@@ -18,7 +18,7 @@
         <% body.with_row do |row| %>
           <% row.with_cell do %>
             <span class="nhsuk-table-responsive__heading">Full name</span>
-            <%= patient.full_name %>
+            <%= govuk_link_to patient.full_name, patient_path(patient) %>
```
tvararu, line 21: `link_to` is configured to be the same in our app:

```suggestion
            <%= link_to patient.full_name, patient_path(patient) %>
```

tvararu (approved):

app/components/app_patient_table_component.html.erb
thomasleese: Oh that's cool, didn't realise that.

PR #2068, opened 2024-10-18
Handle restricted patients

thomasleese:
This adds functionality to handle when a patient is flagged as restricted (otherwise known as sensitive) which means that the system won't automatically send out any emails/texts to their parents, and will prevent the contact information of the patient and parents from being displayed to users.

## Screenshots

Screenshot 2024-10-18 at 11 06 53

Screenshot 2024-10-18 at 11 00 15


misaka (approved):
As discussed in Slack, there's a place where encapsulation might help (e.g. adding `PDS::Patient#restricted?`), but up to you if it's worth the effort here.

PR #2097, opened 2024-10-22
Transfer patients from other sessions

tvararu:
This updates the `Session#create_patient_sessions!` method so that instead of destroying the patient sessions for patients in this cohort that are in other sessions, it proposes their transfer to the new session.

A follow-up PR will implement the interface from the prototype on the actual sessions to surface and handle these `proposed_session`s.

### DIFF
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -4,16 +4,22 @@
 #
 # Table name: patient_sessions
 #
-#  id         :bigint           not null, primary key
-#  created_at :datetime         not null
-#  updated_at :datetime         not null
-#  patient_id :bigint           not null
-#  session_id :bigint           not null
+#  id                  :bigint           not null, primary key
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#  patient_id          :bigint           not null
+#  proposed_session_id :bigint
+#  session_id          :bigint           not null
 #
 # Indexes
 #
 #  index_patient_sessions_on_patient_id_and_session_id  (patient_id,session_id) UNIQUE
+#  index_patient_sessions_on_proposed_session_id        (proposed_session_id)
 #  index_patient_sessions_on_session_id_and_patient_id  (session_id,patient_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (proposed_session_id => sessions.id)
 #
 
 class PatientSession < ApplicationRecord
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -75,9 +75,11 @@ class PatientSession < ApplicationRecord
         end
 
   def draft_vaccination_record
-    # HACK: this code will need to be revisited in future as it only really works for HPV, where we only have one
-    # vaccine. It is likely to fail for the Doubles programme as that has 2 vaccines. It is also likely to fail for
-    # the flu programme for the SAIS teams that offer both nasal and injectable vaccines.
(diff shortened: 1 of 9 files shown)


app/models/session.rb
```diff
@@ -131,18 +131,21 @@ def create_patient_sessions!
         required_programmes.all? { |programme| patient.vaccinated?(programme) }
       end
 
-    # First we remove patients from any other upcoming sessions.
+    # Mark existing patient sessions for transfer
     unvaccinated_patients.each do |patient|
-      sessions_other_than_self = patient.upcoming_sessions.reject { _1 == self }
-      next if sessions_other_than_self.empty?
+      other_sessions = patient.upcoming_sessions.where.not(id:)
```
thomasleese, line 136: I think this'll cause a query to the database to run, which is why is originally gone with `reject`.

app/models/session.rb
```diff
@@ -131,18 +131,21 @@ def create_patient_sessions!
...
+      next if other_sessions.empty?
 
       patient
         .patient_sessions
-        .where(session: sessions_other_than_self)
-        .find_each(&:destroy_if_safe!)
+        .where(session: other_sessions)
```
thomasleese, line 141: Do we need to consider patient sessions here that have vaccination records or will that be handled when the transfer is confirmed?

Effectively we can't remove the patient from a session if they've been vaccinated there.

tvararu: I think I'll create `propose_transfer_if_safe!(session)` which handles this

thomasleese: In these cases I think we'd still want to move them across to the new session, but we can't remove them from the old session, effectively the patients would exist in both, at least that's how it works currently.

app/models/patient_session.rb
```diff
@@ -99,11 +108,16 @@ def able_to_vaccinate?
...
     vaccination_records.empty? && gillick_assessments.empty?
   end
+  alias_method :safe_to_transfer?, :safe_to_destroy?
 
   def destroy_if_safe!
     destroy! if safe_to_destroy?
   end
 
+  def propose_transfer_if_safe!(session)
+    update!(proposed_session: session) if safe_to_transfer?
```
thomasleese, line 118: I think we should always set a `proposed_session`, but when we come to build the UI to confirm the move, only if it's `safe_to_transfer?` do we _remove_ the patient from their existing session while also adding them to the proposed session.

thomasleese (commented):
Look good, I think we also need to add the same logic to when class lists are imported and [unknown patients are moved to the clinic session](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/c07b71e9ac73c48728879accc0c53a1bcb05a352/app/models/class_import.rb#L58-L66) (added in #2106), and [also maybe for consent as well](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/main/app/models/consent_form.rb#L368-L381)?

app/models/patient_session.rb
tvararu: Yeah, this makes sense!

app/models/session.rb
tvararu: Okay, dialled back to the original implementation and we can handle this on the moving UI side.

app/models/patient_session.rb
tvararu: Will be handled in the UI in a follow-up

thomasleese (approved):

PR #2102, opened 2024-10-22
Track when sessions are closed

thomasleese:
This adds a new column `closed_at` on sessions which tracks when a session has been closed, which will trigger the movement of any unvaccinated patients to a clinic and move the session over to the completed stage.

I've also refactored some other bits of code related to sessions at the same time, see individual commits for more details.


misaka (approved):

misaka:
:shipit:

PR #2104, opened 2024-10-22
Specify Sentry environment

thomasleese:
This allows us to make a distinction between the different deployed environments that all use the same `RAILS_ENV` when looking at issues in Sentry as they will show that they are coming from the correct environment.

Once deployed we'll need to update our alerting to send a message from these new environments rather than `staging` which will no longer exist.

### DIFF
--- a/copilot/webapp/manifest.yml
+++ b/copilot/webapp/manifest.yml
@@ -67,6 +67,7 @@ environments:
       rolling: recreate # Disables blue-green deployment for speed
     variables:
       RAILS_ENV: staging
+      SENTRY_ENVIRONMENT: accessibility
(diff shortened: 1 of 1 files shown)


copilot/webapp/manifest.yml
```diff
@@ -111,5 +115,6 @@ environments:
         - "give-or-refuse-consent-for-vaccinations.nhs.uk"
     variables:
       RAILS_ENV: production
+      SENTRY_ENVIRONMENT: production
```
thomasleese, line 118: Not technically needed as Sentry will take from `RAILS_ENV`, but left here for consistency with the others.

tvararu (approved):

PR #2106, opened 2024-10-22
Add home educated and unknown patients to the generic clinic

thomasleese:
This adds the functionality to add home educated and unknown patients to the generic clinic under the following circumstances:

- When importing the cohort
- When importing a class list and moving unknown patients out of the school
- When closing a session

### DIFF
--- a/app/models/class_import.rb
+++ b/app/models/class_import.rb
@@ -52,8 +52,10 @@ class ClassImport < PatientImport
 
     return if session.closed?
 
+    # Add newly imported patients to the session
     session.create_patient_sessions!
 
+    # Remove unknown patients from the session and school
     unknown_patients = session.patients - patients
 
     Patient.where(id: unknown_patients.map(&:id)).update_all(school_id: nil)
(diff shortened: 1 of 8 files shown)


app/models/class_import.rb
```diff
@@ -62,5 +64,14 @@ def postprocess_rows!
       .patient_sessions
       .where(patient: unknown_patients)
       .find_each(&:destroy_if_safe!)
+
+    # Add the unknown patients to the generic clinic
```
benilovj, line 68: Don't we want this to happen only when the change from <specific school> => <school unknown> has been confirmed via the UI?

benilovj (approved):
Looks good, just not sure whether we should be moving children immediately post-class list upload

app/models/class_import.rb
thomasleese: Yeah, we will want that. For now I made it automatic as with everything else because that UI hasn't been built yet, under the assumption that it would be changed to use that UI in the future.

benilovj (approved):

PR #2107, opened 2024-10-22
Split completed sessions in to completed and closed

thomasleese:
This partially reverts eabbb6c6c95eeecbe5fbe790d24d09a90caafdc4 by splitting the "Completed" state in to two separate "Completed" and "Closed" states. Completed means that there are no more upcoming dates, but only once someone has pressed the "Close session" button does it move to the closed state.

## Screenshots

Screenshot 2024-10-22 at 16 48 46

### DIFF
--- a/app/controllers/programmes_controller.rb
+++ b/app/controllers/programmes_controller.rb
@@ -19,9 +19,10 @@ class ProgrammesController < ApplicationController
         .includes(:dates, :location)
         .strict_loading
 
+    @closed_sessions = sessions_for_programme.closed.sort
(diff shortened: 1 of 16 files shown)


benilovj (approved):
I guess we don't have a "Close session" yet but once we do, it'd be good to cover some of the changes introduced in this PR with a feature spec.

thomasleese:
> I guess we don't have a "Close session" yet but once we do, it'd be good to cover some of the changes introduced in this PR with a feature spec.

Yeah I'll be adding that button in tomorrow and add a feature spec for it.

PR #2110, opened 2024-10-22
Make it easy to test permissions when username/password enabled

benilovj:


app/models/user.rb
```diff
@@ -82,15 +82,15 @@ def self.find_or_create_from_cis2_oidc(userinfo)
   end
 
   def is_medical_secretary?
-    return false unless Flipper.enabled?(:cis2)
+    return email.include?("admin") unless Flipper.enabled?(:cis2)
```
misaka, line 85: This is one of those times that I find `unless` kinda hard to read. Wish there was a `Flipper.disabled?`

misaka (approved):
🚀

app/models/user.rb
misaka, line 85: I feel like this was unlikely to be tested up until now. But now that we have a way to test, is it worth at least a feature test to show that admins can't record a vaccination?

benilovj: Added in [5971b22](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/2110/commits/5971b22c8e74b949b247d1ebcc31eb358a3b097a)

misaka (approved):
🚀

PR #2113, opened 2024-10-23
Add UI to handle proposed sessions 

tvararu:
This allows nurses to transfer patients from one session to another after they upload a class list.

TODO:

- [x] Handle scenario where session they're coming from has vaccination records (create new patient_session)
- [x] Count of moving out/in
- [x] End to end test

### Screenshots

Screenshot 2024-10-23 at 10 46 25
Screenshot 2024-10-23 at 10 46 38

Screenshot 2024-10-23 at 10 46 41

### DIFF
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -136,7 +136,12 @@ class PatientSession < ApplicationRecord
   def confirm_transfer!
     return unless pending_transfer?
 
-    update!(session: proposed_session, proposed_session: nil)
+    if safe_to_destroy?
+      update!(session: proposed_session, proposed_session: nil)
+    else
+      PatientSession.create!(patient:, session: proposed_session)
+      update!(proposed_session: nil)
+    end
   end
 
   def ignore_transfer!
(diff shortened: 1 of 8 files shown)


app/models/patient_session.rb
```diff
@@ -126,4 +128,23 @@ def latest_triage
...
   end
+
+  def pending_transfer?
+    proposed_session_id.present?
+  end
+
+  def confirm_transfer!
+    return unless pending_transfer?
+
+    if vaccination_records.any?
```
thomasleese, line 139: We also need to consider gillick assessments here, we should use `!safe_to_destroy?` I think.

app/models/patient_session.rb
```diff
@@ -126,4 +128,23 @@ def latest_triage
...
+      PatientSession.create!(patient:, session: proposed_session)
+      update!(proposed_session: nil)
+    else
+      update!(session: proposed_session, proposed_session: nil)
```
thomasleese, line 143: I'm wondering whether it's better to always create a new one and delete old ones? It means `created_at` can be relied on to mean when the patient was added to the session.

tvararu: Hmm, yeah that's a good idea

tvararu: Updated!

thomasleese (approved):

PR #2118, opened 2024-10-23
Generate class lists for model office

thomasleese:
This updates the scripts for the model office to generate class lists, and fixes a bug where importing the class list would remove the NHS numbers (and other attributes) of any patients that we already know about.

### DIFF
--- a/script/generate_model_office_data.rb
+++ b/script/generate_model_office_data.rb
@@ -43,20 +43,20 @@ def create_student(school, year_group, team)
 end
 
 def create_vaccination_record(
-  student,
+  patient_session,
   vaccine:,
-  session:,
   batch:,
   programme:,
   dose_sequence:
 )
-  print "+"
+  session = patient_session.session
+
   FactoryBot.create(
     :vaccination_record,
     :performed_by_not_user,
-    patient: student,
+    patient: patient_session.patient,
     session:,
-    patient_session: PatientSession.find_by(patient: student, session:),
+    patient_session:,
     programme:,
     vaccine:,
     administered_at: session.dates.first.value + rand(8..16).hours,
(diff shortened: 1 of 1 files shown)


script/generate_model_office_data.rb
```diff
@@ -241,11 +253,14 @@ def create_students_and_vaccinations_for(school:, team:, year_size_estimate:)
...
     vaccination_records += v
+
+    progress_bar.increment
   end
 
+  puts "Writing files"
+
   write_nominal_roll_to_file(students)
   write_vaccination_records_to_file(vaccination_records)
-  print "."
```
benilovj, line 248: aww my dots

thomasleese:
> Not sure you meant to add [8ddb224](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/2118/commits/8ddb224b721d9105b68c4f9753f35426203323e0) onto this PR?

Probably should be two separate PRs, but I spotted this bug while testing class list imports... I can split it out if you'd prefer?

benilovj (approved):
👍 

* might be worth sticking the class import row changes on another PR
* If i'm reading the code correctly, the class lists assume nobody's moved at all. It would be good to introduce a small bit of variability into the class lists to better simulate reality (e.g. 5% of children have moved to a different school in the area, for instance, and another 5% have left the area altogether). But this could be a follow-up PR.

thomasleese:
> If i'm reading the code correctly, the class lists assume nobody's moved at all. It would be good to introduce a small bit of variability into the class lists to better simulate reality (e.g. 5% of children have moved to a different school in the area, for instance, and another 5% have left the area altogether). But this could be a follow-up PR.

Yep definitely agree.

PR #2119, opened 2024-10-23
Don't create sessions for community clinics

thomasleese:
Although these location types exist in the service we don't currently handle them correctly and shouldn't be creating sessions for them.

### DIFF
--- a/lib/tasks/clinics.rake
+++ b/lib/tasks/clinics.rake
@@ -33,7 +33,7 @@ namespace :clinics do
 
     location =
       Location.create!(
-        type: :clinic,
+        type: :community_clinic,
         name:,
         address_line_1:,
         address_town:,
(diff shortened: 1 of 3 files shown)


lib/tasks/clinics.rake
```diff
@@ -33,7 +33,7 @@ namespace :clinics do
 
     location =
       Location.create!(
-        type: :clinic,
+        type: :community_clinic,
```
benilovj, line 36: does this change mean we're missing some test coverage? prob worth adding as this is an important bit of ops.

benilovj (approved):

lib/tasks/clinics.rake
thomasleese: I think at the moment we have no tests for our Rake tasks, which I agree is probably would adding.

PR #2121, opened 2024-10-23
Control cis2 with a setting vs feature flag

misaka:
This has wider reaching changesa as it means only one of database_authenticatable or omniauth_authenticatable is in effect at any one time, depending on whether CIS2 is on or off. The benefit is when in CIS2 mode Mavis should be more secure.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -30,11 +30,11 @@ class User < ApplicationRecord
 
   attr_accessor :sso_session
 
-  devise :database_authenticatable,
-         :trackable,
-         :timeoutable,
-         :omniauthable,
-         omniauth_providers: %i[cis2]
+  if Settings.cis2.enabled
+    devise :omniauthable, :trackable, :timeoutable, omniauth_providers: %i[cis2]
+  else
+    devise :database_authenticatable, :trackable, :timeoutable
+  end
 
   has_and_belongs_to_many :teams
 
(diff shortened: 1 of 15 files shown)


app/models/user.rb
```diff
@@ -82,15 +82,15 @@ def self.find_or_create_from_cis2_oidc(userinfo)
   end
 
   def is_medical_secretary?
-    return email.include?("admin") unless Flipper.enabled?(:cis2)
+    return email.include?("admin") unless Settings.cis2.enable
```
tvararu, line 85:
```suggestion
    return email.include?("admin") unless Settings.cis2.enabled
```

app/models/user.rb
```diff
@@ -82,15 +82,15 @@ def self.find_or_create_from_cis2_oidc(userinfo)
...
 
     role_codes = sso_session.dig("selected_role", "code")
     role_codes.include?("R8006")
   end
 
   def is_nurse?
     # All users are nurses if cis2 is disabled
-    return email.include?("nurse") unless Flipper.enabled?(:cis2)
+    return email.include?("nurse") unless Settings.cis2.enable
```
tvararu, line 93:
```suggestion
    return email.include?("nurse") unless Settings.cis2.enabled
```

tvararu (approved):

thomasleese (approved):

PR #2126, opened 2024-10-23
Propose move to generic clinic when uploading class list

tvararu:
When users upload a class list, instead of automatically moving patients to the generic clinic, propose their move to that session.

Also update the heading/content for the column to be accurate (before it was always "School joined from" in both tabs which is wrong).

Screenshot 2024-10-24 at 08 40 30

### DIFF
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -143,6 +143,10 @@ class PatientSession < ApplicationRecord
     PatientSession.transaction do
       PatientSession.create!(patient:, session: proposed_session)
 
+      school = proposed_session.location
+      school = nil if school.generic_clinic?
+      patient.update!(school:)
+
       safe_to_destroy? ? destroy! : update!(proposed_session: nil)
     end
   end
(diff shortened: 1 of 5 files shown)


app/models/patient_session.rb
```diff
@@ -141,9 +141,18 @@ def confirm_transfer!
...
-
-      safe_to_destroy? ? destroy! : update!(proposed_session: nil)
+      new_patient_session =
+        PatientSession.create!(patient:, session: proposed_session)
+
+      if safe_to_destroy?
+        school = new_patient_session.location
+        school = nil if school.generic_clinic?
+
+        patient.update!(school:)
```
thomasleese, line 151: I think we should always update the school even if we can't delete the old session, in the case of a move, we'd still want the new school to be reflected on the patient.

tvararu: Yeah, makes sense

thomasleese (approved):

PR #2127, opened 2024-10-23
Make users.email nullable

misaka:
When the account is coming from an SSO provider (CIS2) we don't need to have an email for it. In the case of CIS2, even, it's explicitly stated that the email address may not be valid and should not be relied on to work.


benilovj (approved):

app/models/user.rb
```diff
@@ -7,7 +7,7 @@
 #  id                  :bigint           not null, primary key
 #  current_sign_in_at  :datetime
 #  current_sign_in_ip  :string
-#  email               :string           default(""), not null
+#  email               :string           default("")
```
thomasleese, line 10: I'm wondering why we've got a `default("")` if we're allowing the column to be `null`? We could leave the emails as blank for these users rather than `null`.

misaka: Ya, I was _just_ thinking the same thing. Will change.

misaka: Changed

thomasleese (approved):

PR #2137, opened 2024-10-23
Send invitations to clinics

thomasleese:
This adds the functionality that sends invitations to clinics if the patient belongs to a community clinic, they haven't already been vaccinated, and we haven't received a refused consent from the parent.

Some outstanding questions:

- When should we be sending these?
- What happens if someone has booked in to a clinic? We may continue to send them emails between them booking in to a clinic and the patient getting vaccinated.


benilovj (approved):
Looks great!

I notice we're not sending texts yet for clinics - is that because we don't have the templates yet? 

Would be good to track this on a card, together maybe with making the timings configurable, and any other follow-on bits I might have missed....

thomasleese:
> I notice we're not sending texts yet for clinics - is that because we don't have the templates yet?

Yep that's right, I couldn't see any templates, I'll ask about it in Slack today.

> Would be good to track this on a card, together maybe with making the timings configurable, and any other follow-on bits I might have missed....

Yeah I agree.

PR #2148, opened 2024-10-24
Enable CIS2 in test

misaka:
Things to check:
- [x] cis2 config created
- [x] client_id and secret are configured


tvararu:
We decided not to move forward with this right now.

PR #2159, opened 2024-10-25
Support passing just patients to patient tables

thomasleese:
This refactors the `AppSessionPatientTableComponent` to support passing in a list of patients rather than a list of patient sessions.

This is needed to support the scenario where we may not have a session, specifically this will be used to render the new global consent response matching view which needs to display patients across all sessions, not just patients in a current session.

I reckon we could probably merge the `AppPatientTableComponent` and `AppSessionPatientTableComponent` in to a single component after this change, but I'll leave that for a later refactor.


tvararu (commented):
Nice

tvararu (approved):
Nice

PR #2160, opened 2024-10-25
Allow filtering on postcodes

thomasleese:
When displaying the session patient table we want to allow the users to be able to filter on postcodes when displaying the list for the global consent matching page.


thomasleese:
Included in #2161.

PR #2170, opened 2024-10-26
Handle CSV uploads with a UTF8 BOM prefix

benilovj:
This can be happen when exporting CSVs from Excel.


misaka (approved):

benilovj:
Original implementation: 226fdd47

tvararu (approved):

PR #2172, opened 2024-10-26
Fix for the manually-triggered consent request emails not appearing in the activity log

benilovj:
I was doing local testing of the new email and text logging introduced in #2169, and wasn't able to get the email logs to appear in the activity log using the "Send consent request" button.

* At first glance, it looked like the ActionMailer observer codepath wasn't being executed.
* I noticed that the registration of the observer seemed different from what was in the observers documentation[^1] so I tried using the initializer as suggested – still couldn't get it to work
* I then came across a post [^2] about the new `after_deliver` callback that was introduced in Rails 7.1 [^3]. This approached seemed slightly cleaner anyway, as `ApplicationMailer` seems a more obvious place to stumble upon the logging logic.
* I added test coverage to prevent regressions

[^1]: https://guides.rubyonrails.org/v7.2/action_mailer_basics.html#observing-emails
[^2]: https://www.bigbinary.com/blog/rails-7-1-adds-_-deliver-callbacks-to-action-mailer
[^3]: https://guides.rubyonrails.org/v7.2/action_mailer_basics.html#action-mailer-callbacks


thomasleese (approved):
Nice!

PR #2173, opened 2024-10-27
Date validation fixes

benilovj:
* Format dates in ActiveRecord validation error messages; show D MMM YYYY format, instead of YYYY-MM-DD. This is applied globally
* Some attribute-specific validation fixes on children's dates of birth on the consent form, and on session date validations

### Before

image

image

### After

image

image

### Before

image

image

### After

image

image

### DIFF
--- a/app/models/consent_form.rb
+++ b/app/models/consent_form.rb
@@ -176,9 +176,9 @@ class ConsentForm < ApplicationRecord
     validates :date_of_birth,
               presence: true,
               comparison: {
-                less_than: Time.zone.today,
-                greater_than_or_equal_to: 22.years.ago.to_date,
-                less_than_or_equal_to: 3.years.ago.to_date
+                less_than: -> { Time.zone.today },
+                greater_than_or_equal_to: -> { 22.years.ago.to_date },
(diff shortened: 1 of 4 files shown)


config/initializers/i18n_custom_interpolation.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# This custom interpolation is used to format interpolated dates in error messages
+# so they align to the NHS design system.
+module I18n
+  module CustomInterpolation
```
misaka, line 6: Since we're including it below, can we customise the name as it's specifically about reformatting the date? Or do you expect this to be used for other types of values too?

misaka (commented):
Looks good. Small question about naming, ready to approve once that's sorted.

config/initializers/i18n_custom_interpolation.rb
benilovj: Ya, fair. I've renamed it to `NHSDesignSystemDateInterpolation`.

misaka (approved):

PR #2177, opened 2024-10-27
Rename medical_staff to admin_staff

misaka:
This commit has been pulled out of a larger piece of work to make team-based authorisation work in CIS2-mode.

We generally use "admin" instead of "medical secretary", so this change brings it closer to that. But also, medical secretary may be only one of the NHS RBAC roles that qualifies for the "admin [staff]" role in Mavis.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -84,7 +84,7 @@ class User < ApplicationRecord
     user.tap(&:save!)
   end
 
-  def is_medical_secretary?
+  def is_admin?
     return email.include?("admin") unless Settings.cis2.enabled
 
     selected_role = cis2_info.dig("selected_role", "code")
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -9,15 +9,15 @@ class ApplicationPolicy
   attr_reader :user, :record
 
   def index?
-    user.is_nurse? || user.is_medical_secretary?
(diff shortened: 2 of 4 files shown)


app/models/user.rb
```diff
@@ -84,7 +84,7 @@ def self.find_or_create_from_cis2_oidc(userinfo)
     user.tap(&:save!)
   end
 
-  def is_medical_secretary?
+  def is_admin_staff?
```
thomasleese, line 87:
```suggestion
  def is_admin?
```

Could we also go with this? For consistency with `is_nurse?`.

thomasleese (approved):

app/models/user.rb
misaka: We could. It's nice and matches the length of "nurse", but also I was worried with this being mistaken with a "sysadmin", which is the other commonly used meaning. I was  hoping `admin_staff` would void confusion.

Do you think `admin` wouldn't be misleading?

misaka: @tvararu How do you feel about calling this `admin` vs `admin_staff`? ☝️

tvararu: I prefer `admin`

tvararu (approved):

app/models/user.rb
misaka: 👍 Let's do it

PR #2179, opened 2024-10-28
Turn on cis2 be default in tests

misaka:
As we are moving towards having CIS2 enabled by default, in deployed envs at least, I wanted to update the tests to work in CIS2 mode. With this change all the tests now run in CIS2 mode by default. Some tests that tested local-user access have been temporarily disabled, to run them you need  to disable CIS2 in `config/settings/test.yml` and then run rspec with the correct tag:

```
bundle exec rspec -t local_users
```


thomasleese (approved):

tvararu (approved):
https://github.com/nhsuk/manage-vaccinations-in-schools/pull/2177 needs to be merged first?

PR #2192, opened 2024-10-28
Log id token when cis2 auth errors

misaka:
We've had some issues where an exception is raised while we process smartcard logisn. This should help us debug them better.


thomasleese (approved):
Another option could be to sent this information to Sentry, but this works too.

misaka:
> Another option could be to sent this information to Sentry, but this works too.

Good idea. Do you remember how to do that offhand?

misaka:
`with_scope` seems like a good candidate. https://docs.sentry.io/platforms/ruby/guides/rack/enriching-events/scopes/#using-with-scope

thomasleese:
> `with_scope` seems like a good candidate. https://docs.sentry.io/platforms/ruby/guides/rack/enriching-events/scopes/#using-with-scope

Yeah, that's the one, and then you can add `context` here: https://docs.sentry.io/platforms/ruby/enriching-events/context/

PR #2193, opened 2024-10-28
Mark patients as invalid when NHS number is invalid

thomasleese:
If the NHS number itself is invalid we can mark the patient as invalid. This extends the definition of an invalid patient from the definition in PDS to include these cases.

An advantage of this is that we can avoid checking PDS for patients we know will never succeed by excluding all types of invalid patients. A potential disadvantage is that these patients will appear in the notices, although I’d argue they are still "invalid".

### DIFF
--- a/app/jobs/patient_update_from_pds_job.rb
+++ b/app/jobs/patient_update_from_pds_job.rb
@@ -5,7 +5,6 @@ class PatientUpdateFromPDSJob < ApplicationJob
 
   queue_as :pds
 
-  discard_on NHS::PDS::InvalidNHSNumber
   discard_on NHS::PDS::PatientNotFound
 
   def perform(patient)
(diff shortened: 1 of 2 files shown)


thomasleese:
> Could we hold off merging this? I'm not sure whether we will have thousands of these records in a given cohort and that could overwhelm the important notices view...

Yeah no problem.

benilovj:
Arguably this need is now met by https://github.com/nhsuk/manage-vaccinations-in-schools/pull/2292

thomasleese:
> Arguably this need is now met by #2292

I still think maybe it's worth flagging `NHS::PDS::InvalidNHSNumber` as invalid, this is where the NHS number is itself never going to return a person because it's invalid in some way (i.e. check digit is wrong), because we should never have a patient with an invalid NHS number.

But I agree `NHS::PDS::PatientNotFound` is probably covered by #2292.

benilovj:
> I still think maybe it's worth flagging `NHS::PDS::InvalidNHSNumber` as invalid, this is where the NHS number is itself never going to return a person because it's invalid in some way (i.e. check digit is wrong), because we should never have a patient with an invalid NHS number.

Agreed.

benilovj (approved):

PR #2229, opened 2024-10-30
Don't move patients when they have draft records

misaka:
Currently if a patient has a draft gillick assessment or vaccination record, when they are moved between schools an error is triggered. This is because the existing draft records cause a foreign key constraint error.

https://good-machine.sentry.io/issues/6028433066/

The fix in this PR is to treat draft versions of these associations the same as the recorded versions. This appears to be align with the spirit of PR #2018 which introduces the change to not move patients who have gillick assessments or vaccination records.

It's also possible the right fix is to remove these draft records, but that seems unlikely since as the #2018 states, it's unclear what actual state the patient record is in.

### DIFF
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -119,7 +119,8 @@ class PatientSession < ApplicationRecord
   end
 
   def safe_to_destroy?
-    vaccination_records.empty? && gillick_assessments.empty?
+    VaccinationRecord.where(patient_session: self).empty? &&
+      GillickAssessment.where(patient_session: self).empty?
   end
 
   def destroy_if_safe!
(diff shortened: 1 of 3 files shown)


benilovj (changes requested):
I'm not sure this is the right approach. If a Gillick assessment has been recorded, it means a child was seen by a nurse during a session. If a draft exists, it doesn't mean anything - it could just mean someone clicked on the option and then clicked away, like what happened on the test env. 

I think we should be discarding the draft here and proceeding with the move.

misaka:
Will sort tomorrow.

thomasleese (dismissed):

PR #2230, opened 2024-10-30
Add explicit cancel link on the 'Edit session dates' page

benilovj:
Without this, the only way off the page is to find the dinky 'Back' link right at the top of a potentially long page.

image


thomasleese (approved):

benilovj:
@paulrobertlloyd this isn't in the designs, it's a dev-design suggestion that I'm easy if we choose not to proceed with.

tvararu (approved):

paulrobertlloyd:
Works for me!

PR #2245, opened 2024-10-30
Remove draft records when moving patient

misaka:
Patients may have draft gillick assessments or vaccination recrods which need to be removed before they can be moved to a new session.

This replaces the previous PR (#2229) which would copy patient record instead of moving it if it had any draft records. (didn't realise renaming the branch would close it)

### DIFF
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -154,6 +154,9 @@ class PatientSession < ApplicationRecord
       school = nil if school.generic_clinic?
       patient.update!(school:)
 
+      draft_vaccination_records.destroy_all
+      draft_gillick_assessments.destroy_all
+
       safe_to_destroy? ? destroy! : update!(proposed_session: nil)
     end
   end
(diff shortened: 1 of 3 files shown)


app/models/patient_session.rb
```diff
@@ -144,6 +144,11 @@ def pending_transfer?
     proposed_session_id.present?
   end
 
+  def remove_unrecorded_records
+    VaccinationRecord.where(patient_session: self).draft.destroy_all
+    GillickAssessment.where(patient_session: self).draft.destroy_all
```
thomasleese, line 149:
```suggestion
    draft_vaccination_records.destroy_all
    draft_gillick_assessments.destroy_all
```

This could be a nicer alternative, the relationships are defined here: https://github.com/nhsuk/manage-vaccinations-in-schools/blob/23ffaf1d94a04e43725d721176f4631a0d629623/app/models/patient_session.rb#L38-L54

thomasleese (approved):

misaka:
@thomasleese If you gIve that a 👍 I'll squash that before merging.

PR #2247, opened 2024-10-30
Add basic programme patients view

tvararu:
This adds a link within a programme to view the children that are part of it. The table isn't aligned to the prototype yet, that's coming in a follow-up PR.

### Screenshots

Screenshot 2024-10-30 at 12 32 04

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -110,6 +110,8 @@ Rails.application.routes.draw do
 
     resources :cohorts, only: %i[index show]
 
+    resources :patients, only: %i[index], module: "programme"
+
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -110,6 +110,8 @@
 
     resources :cohorts, only: %i[index show]
 
+    resources :patients, only: %i[index], module: "programme"
```
thomasleese, line 113: Nice use of the module, I've been wondering about whether we should structure more controllers like this as we've got a few different views of the same data in different sections of the app.

thomasleese (approved):

PR #2253, opened 2024-10-30
Rename team to organisation

thomasleese:
This renames the team model to organisation to matching the naming convention we already use in the UI and in CIS2. This also allows us to introduce our own concept of a team which will sit between schools and organisations to allow custom contact details.


tvararu (approved):
I haven't reviewed this, but if the tests pass, it must be okay 👍

PR #2255, opened 2024-10-30
Create team model

thomasleese:
This introduces a new concept of a team (separate from organisation which was previously named team) to represent a way of grouping records, starting with schools.


tvararu (approved):
Teams are dead! Long live teams!

PR #2267, opened 2024-10-31
Require CIS2 users to be in the right workgroup

misaka:
This gives SAIS teams a way to limit who has access to Mavis within their organisation.

When a user is not in the correct workgroup this error page is shown:
image


misaka:
Just realised that the "Change Role" button is redundant, and probably also true on the "wrong team" error page. ¯\_(ツ)_/¯ ... we can fix this later maybe.

benilovj:
> Just realised that the "Change Role" button is redundant, and probably also true on the "wrong team" error page. ¯\_(ツ)_/¯ ... we can fix this later maybe.

If it's redundant in the sense that it does the same as the button in the header, that's fine as it's an obvious call to action.

app/controllers/concerns/authentication_concern.rb
```diff
@@ -35,6 +35,11 @@ def selected_cis2_org_is_registered?
       )
     end
 
+    def selected_cis2_workgroup_is_valid?
+      selected_cis2_nrbac_role.key?("workgroups") &&
+        "schoolagedimmunisations".in?(selected_cis2_nrbac_role["workgroups"])
```
benilovj, line 40: Should the magic string be in a constant at the top of the file, for obviousness?

benilovj (approved):

PR #2274, opened 2024-10-31
Use pentest not test for concurrency group

misaka:
When starting up in "test" we cancel out the other "test" deploy job.


misaka:
A little bit of a stab in the dark, but seems like a good guess as to why the other job was cancelled.

tvararu (approved):

PR #2277, opened 2024-10-31
Make Consent#name more robust

thomasleese:
This avoids trying to get the label for a `nil` parent.

https://good-machine.sentry.io/issues/6033335850/?environment=test&project=4505625073876992&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=2

### DIFF
--- a/app/models/consent.rb
+++ b/app/models/consent.rb
@@ -140,7 +140,9 @@ class Consent < ApplicationRecord
   delegate :restricted?, to: :patient
 
   def name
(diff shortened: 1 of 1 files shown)


benilovj:
I think the underlying bug is elsewhere? it looks like maybe a draft consent is being surfaced on the activity log when it shouldn't be?

PR #2286, opened 2024-11-01
Add disaster recovery instructions

tvararu:
These can be used to spin up a new database from a dump of a live database. Then we can run a new Rails server against it, or export data for offline or disaster scenario use.

### DIFF
--- a/docs/disaster-recovery.md
+++ b/docs/disaster-recovery.md
@@ -1,0 +1,156 @@
+# Disaster recovery ops manual
+
+## Getting a local dump of an Aurora DB
+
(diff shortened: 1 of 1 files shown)


docs/disaster-recovery.md
```diff
@@ -0,0 +1,156 @@
+# Disaster recovery ops manual
+
+## Getting a local dump of an Aurora DB
+
+You need Postgres 16.x to connect to the Aurora DB.
```
misaka, line 5: Just to mention, I used 17.0 so I think maybe it's 16+?

misaka (approved):

PR #2291, opened 2024-11-03
Parent form load test using k6

benilovj:


tests/parent-form-load-test.js
```diff
@@ -0,0 +1,317 @@
+import encoding from "k6/encoding";
+import http from "k6/http";
+import { check, sleep } from "k6";
+import { randomIntBetween } from "https://jslib.k6.io/k6-utils/1.2.0/index.js";
```
tvararu, line 4: I think I prefer this tbh:

```suggestion
function randomIntBetween(min, max) {
  return Math.floor(Math.random() * (max - min + 1)) + min;
}
```

PR #2293, opened 2024-11-04
Ensure pending patients are visible

thomasleese:
We limit which patients a user can seen, but we need to ensure that any patients with pending changes that brings them in to the organisation are visible to users so they can review and confirm the changes, as currently they won't be accessible and will raise an error.


misaka (approved):

misaka:
Did we want this on a `release-1a` branch perhaps?

PR #2298, opened 2024-11-04
Normalise parent phone numbers with Phonelib

benilovj:
* format national numbers with spaces
* save non-GB numbers in an international format
* handle leading zeros being stripped (when handling CSVs in Excel, for instance)


misaka (approved):
🚀

tvararu (approved):

PR #2309, opened 2024-11-05
Don't filter consents and triage on programme 

thomasleese:
This should improve performance by removing the need to have a reference to the patient session when performing the join, which prevents Rails from being able to run a single query to avoid N+1 issues. This does slightly break functionality, but since we're only running with one programme at the moment, and other functionality is broken with regards to multiple programmes, it seems like a reasonable fix for now.

### DIFF
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -41,27 +41,22 @@ class PatientSession < ApplicationRecord
            -> { draft },
            class_name: "GillickAssessment"
 
-  has_many :vaccination_records,
-           -> { recorded },
-           class_name: "VaccinationRecord"
+  has_many :vaccination_records, -> { recorded }
   has_many :draft_vaccination_records,
            -> { draft },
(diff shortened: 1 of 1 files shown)


thomasleese:
Locally this is a big performance improvement (from @benilovj):

```
before: 08:19:43 web.1  | Completed 200 OK in 39626ms (Views: 140.9ms | ActiveRecord: 2530.5ms (792 queries, 1 cached) | GC: 6163.8ms)

after: 08:20:56 web.1  | Completed 200 OK in 10149ms (Views: 95.0ms | ActiveRecord: 1837.5ms (23 queries, 1 cached) | GC: 1785.4ms)
```

In the future we'll likely need to look at caching these statistics values or patient session stats as we will need to filter on programmes, but it's also possible that the data model will change in a way that avoids the expensive queries in the future anyway.

benilovj (approved):

PR #2310, opened 2024-11-05
Small fixes to get the Cov scripts back up and running

benilovj:


script/set_up_coventry.sh
thomasleese, line 1: As mentioned in Slack we left with intentionally unexecutable, although if we do want it to be executable we'll likely want to add a shebang to it.

thomasleese (approved):

PR #2314, opened 2024-11-05
Refactor settings to make defaults prodlike

tvararu:
This moves most production and staging-specific config from staging/production.yml to the default settings.yml. Development and test are then configured to be different on purpose, which is easier to think about.

The goal is to have an empty `config/settings/production.yml`, and a near-empty `config/settings/staging.yml`.

Also removes the now redundant `allow_dev_phone_numbers: true` env setting for staging envs, but adds the `cis2_enabled: false` for all non-test envs.

CIS2/Mesh/NHSAPI are configured with the int environments but can be changed to prod ones later.


thomasleese (approved):
I like this approach!

PR #2316, opened 2024-11-05
Update URNs in seeds for school used by tests

misaka:
134522 was a closed school so we're replacing it with an open one.

Once this is merged I'll let the testers know of the new school names.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -194,13 +194,17 @@ unless Settings.cis2.enabled
   create_user(organisation:, email: "admin.hope@example.com")
 
   attach_sample_of_schools_to(organisation)
+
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -200,7 +200,7 @@ def create_imports(user, organisation)
   ) # needed for automated testing
   attach_specific_school_to_organisation_if_present(
     organisation:,
-    urn: "134522"
+    urn: "145377"
   ) # needed for automated testing
```
tvararu, line 204: Can you update these two comments as well?

tvararu (approved):

thomasleese (approved):

PR #2318, opened 2024-11-05
Reference more settings from credentials

tvararu:
This moves `govuk_notify.api_key` and two of the Mesh settings to credentials.

More settings will be added in follow-ups, but it's best to merge these changes frequently to more easily identify why things break if they do.


thomasleese (approved):

misaka (approved):

tvararu:
Playwright tests will be fixed in a follow-up.

PR #2321, opened 2024-11-05
Add "Edit vaccination record" page

thomasleese:
This adds a page that will include all the change links for various parts of a vaccination record that can be changed.

## Screenshot

Screenshot 2024-11-05 at 12 40 09


tvararu (approved):

tvararu:
@thomasleese temporarily disabled auto-merge on this so that we can see if the current deploy from `main` succeeds without having it cancelled

PR #2323, opened 2024-11-05
Reference mesh and nhs_api credentials

tvararu:
This makes managing environment variables easier.


thomasleese (approved):

tvararu:
Linter fails on yml because of the weird erb indentation; will add an exception

PR #2324, opened 2024-11-05
Reference cis2 settings from credentials

tvararu:
This just about completes the migration from using `MAVIS__` environment variables for secrets to using the Rails credentials.


thomasleese (approved):

tvararu:
Have added `SENTRY_DSN` to this as well.

PR #2326, opened 2024-11-05
Reset unused vaccination record fields

thomasleese:
This handles the situation where going back to edit previously set values on a vaccination record and this method ensures the record is consistent.

To support the tests I've had to set a `create` strategy on the associations to ensure that building a vaccination record has a valid state, without this various `has_one` through associations end up being `nil` which leads to validation errors.


tvararu (approved):

tvararu:
@thomasleese I've paused your auto-merge once again briefly 🤞

PR #2329, opened 2024-11-05
Extract shared deploy template

tvararu:
This DRYs out the test/pentest deploy scripts.

PR #2338, opened 2024-11-06
Revert "Extract shared deploy template"

tvararu:
Reverts nhsuk/manage-vaccinations-in-schools#2329

Main isn't auto-deploying anymore and we shouldn't look into why now because time is limited.


tvararu:
```
[Invalid workflow file: .github/workflows/deploy-test.yml#L21](https://github.com/nhsuk/manage-vaccinations-in-schools/actions/runs/11701302195/workflow)
invalid value workflow reference: workflows must be defined at the top level of the .github/workflows/ directory
```

Probably wrong workflow file location...

misaka (approved):

PR #2348, opened 2024-11-06
Add ability to withdraw and invalidate consents

thomasleese:
This adds functionality allowing nurses to withdraw consent (marking consent that's already given and change to refused) or to invalidate consent (consent that's either refused or given and mark it as invalid so it's ignored).

## Screenshots

Screenshot 2024-11-06 at 17 48 54
Screenshot 2024-11-06 at 17 48 51
Screenshot 2024-11-06 at 17 47 55
Screenshot 2024-11-06 at 17 23 36
Screenshot 2024-11-06 at 16 53 42


config/locales/en.yml
```diff
@@ -343,7 +343,7 @@ en:
               blank: Choose who you are trying to get consent from
             reason_for_refusal:
               inclusion: Choose a reason
-            reason_for_refusal_notes:
+            notes:
               blank: Enter details for refusing
```
benilovj, line 347: I think the `blank` error messages need to be more general now?

thomasleese: Good point, I've amended the commit to go with "Enter notes" which I think matches our other error messages.

spec/features/invalidate_consent_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    then_i_see_the_consent
+
+    when_i_click_on_the_consent
+    then_i_see_the_consent
+    and_i_click_invalidate_consent
+
+    when_i_fill_in_the_notes
+    and_i_click_invalidate_consent
+    then_i_see_the_consent_has_been_withdrawn
+    and_i_cant_withdraw
```
benilovj, line 21: worth asserting here that patient session status has changed and that it's not possible to vaccinate anymore

app/components/app_activity_log_component.rb
```diff
@@ -39,11 +39,40 @@ def all_events
   end
 
   def consent_events
```
benilovj, line 41: would be good add spec coverage for these events

thomasleese:
@benilovj Thanks, I think all those points should be resolved now.

benilovj (approved):

PR #2352, opened 2024-11-06
Don't send email if consent is not provided

thomasleese:
Currently we send an email and a text saying consent was given if the user has selected the consent not provided option which is incorrect and could lead to confusion.


benilovj:
I think the code change here is good to bring in, as it's safer than a catch-all `else` statement.
However it might be best to remove the "No response" option altogether - I don't think it's in the prototype anymore, I believe it's a holdover from the pilot.

thomasleese:
> However it might be best to remove the "No response" option altogether - I don't think it's in the prototype anymore, I believe it's a holdover from the pilot.

Oh that makes things simpler! I'll do that in a separate PR.

benilovj (approved):

benilovj:
d'oh sorry, i thought you meant you wanted to merge this one in...

thomasleese:
> d'oh sorry, i thought you meant you wanted to merge this one in...

No worries!

PR #2356, opened 2024-11-07
Update content when vaccination is not administered

thomasleese:
In the activity log, currently vaccinations are recorded as having been given even if they weren't administered which is incorrect.

### DIFF
--- a/app/components/app_activity_log_component.rb
+++ b/app/components/app_activity_log_component.rb
@@ -110,12 +110,19 @@ class AppActivityLogComponent < ViewComponent::Base
   end
 
   def vaccination_events
-    vaccination_records.recorded.map do
+    vaccination_records.recorded.map do |vaccination_record|
+      title =
+        if vaccination_record.administered?
+          "Vaccinated with #{helpers.vaccine_heading(vaccination_record.vaccine)}"
+        else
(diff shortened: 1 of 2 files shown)


tvararu (approved):

app/components/app_activity_log_component.rb
```diff
@@ -110,12 +110,21 @@ def triage_events
...
   def vaccination_events
-    vaccination_records.recorded.map do
+    vaccination_records.recorded.map do |vaccination_record|
+      vaccine = vaccination_record.vaccine
+
+      title =
+        if vaccination_record.administered?
+          "Vaccinated with #{helpers.vaccine_heading(vaccine)}"
+        else
+          "Not vaccinated with #{helpers.vaccine_heading(vaccine)}"
```
benilovj, line 120: Not sure about listing the vaccine here? It may require rework for flu, where you have two possible products (nasal and injection). Maybe we should mention the programme instead? I'll check with the design team.

thomasleese: Yeah, I did wonder about that myself, I think not including the vaccine would be reasonable.

benilovj (approved):
Let's roll forward with this change as is, and then tweak it once we have different copy.

PR #2364, opened 2024-11-08
Create draft vaccination records in browser session

thomasleese:
This updates how vaccination recording and editing is build to use the new `DraftVaccinationRecord` class instead of storing draft records in the database. This implements the behaviour in the prototype where users can make numerous changes before committing the changes.

This will also result in there being no more draft records in the database so a follow up change will be made to remove them.

See individual commits for more details.


tvararu (approved):

benilovj:
I hit an exception when i was recording a new vaccination and hit Edit on the vaccine row of Check and confirm:

image

thomasleese:
> I hit an exception when i was recording a new vaccination and hit Edit on the vaccine row of Check and confirm:

Ah looks like we're missing test coverage for that, I'll add one now.

PR #2369, opened 2024-11-08
Create draft gillick assessments in browser session

thomasleese:


thomasleese:
I don't think we'll need this actually as the design of this page has changed quite a bit and we can probably do it without needing a wizard.

PR #2385, opened 2024-11-10
Add record attendance feature

tvararu:
This adds a tag and a link to update a patient's attendance in a particular session on a particular date. The functionality only appears when the feature flag is enabled and the session is in progress. With the flag on, users can't record vaccinations before they register attendance.

A `SessionAttendance` can be in 3 possible states:

- Non existent, which is the "Not registered yet" state
- `attending = false`, which is the `Absent from today's session" state
- `attending = true`, which is the "Attending today's session" state

The `update` action in the controller manages moving between these states, which means that if the user chooses the "They have not been registered yet," it actually destroys the `SessionAttendance` record.

This is slightly unconventional, but makes sense in this context. If we keep them around, it would potentially create an expectation of having to create `SessionAttendance` records for each N patient times each M date, up-front. It also allows us to make the `attending` field nullable.

### Screenshots

Screenshot 2024-11-10 at 16 28 03
Screenshot 2024-11-10 at 16 28 09
Screenshot 2024-11-10 at 16 28 14
Screenshot 2024-11-10 at 16 28 21


thomasleese (commented):
Looks great, just got a few comments around the foreign keys on the model.

app/models/patient_session.rb
```diff
@@ -92,7 +93,8 @@ def able_to_vaccinate?
   end
 
   def safe_to_destroy?
-    vaccination_records.empty? && gillick_assessment.nil?
+    vaccination_records.empty? && gillick_assessment.nil? &&
+      session_attendances.none?(&:attending?)
```
thomasleese, line 97: I think this is the right logic but I think we'll need to add something to `destroy_if_safe!` or the relationship to make sure that if we do try and destroy this patient session and there are only session attendances where `attending = false` they get deleted as well, I think currently we'll get a foreign key constraint error.

tvararu: We can probably use `dependent:` again 🤔 Good catch

thomasleese (approved):

PR #2387, opened 2024-11-11
Treat already had vaccination records as administered

thomasleese:
This ensures that patients who have already had the vaccine are treated as administered, which means we don't move them to clinic when sessions are closed, for example.

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -246,7 +246,7 @@ class Patient < ApplicationRecord
     # TODO: This logic doesn't work for vaccinations that require multiple doses.
 
     vaccination_records.any? do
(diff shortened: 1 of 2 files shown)


app/models/patient.rb
```diff
@@ -246,7 +246,7 @@ def vaccinated?(programme)
     # TODO: This logic doesn't work for vaccinations that require multiple doses.
 
     vaccination_records.any? do
-      _1.administered? && _1.programme_id == programme.id
```
tvararu, line 249: I do like how straightforward this change was

tvararu (approved):

app/models/patient.rb
thomasleese: Me too!

PR #2417, opened 2024-11-12
Fix TIME_OF_VACCINATION validation

thomasleese:
This fixes an error in the validation where it was expecting a value where it shouldn't.


benilovj:
@thomasleese I added a spec

benilovj (approved):

PR #2432, opened 2024-11-13
Tweak Careplus exporter

tvararu:
We've confirmed that the Suspension end date field isn't used. We also found out some more information about the Dose and Site fields that help us get them closer to correct, but they still need some more tweaking.

### DIFF
--- a/app/lib/reports/careplus_exporter.rb
+++ b/app/lib/reports/careplus_exporter.rb
@@ -125,7 +125,7 @@ class Reports::CareplusExporter
       "School Nurse", # Staff Code
       "Y", # Attended; Did not attends do not get recorded on GP systems
       "", # Reason Not Attended; Always blank
-      "", # Suspension End Date; Not sure what this is, leaving blank
+      "", # Suspension End Date; Doesn't need to be used
       *vaccine_fields(vaccination_records, 0),
       *vaccine_fields(vaccination_records, 1),
       *vaccine_fields(vaccination_records, 2),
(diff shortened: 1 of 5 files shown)


app/lib/reports/careplus_exporter.rb
```diff
@@ -140,11 +140,31 @@ def vaccine_fields(vaccination_records, index)
 
     [
       record.vaccine.snomed_product_code, # Vaccine X
-      "", # Dose X field; Not sure, documentation says this is derived later?
+      "1P", # Dose X field; Apparently 1P or 2P for HPV, but not sure why/when it changes
```
benilovj, line 143:
```suggestion
      "#{record.dose_sequence}P", # Dose X field;
```

app/lib/reports/careplus_exporter.rb
```diff
@@ -140,11 +140,31 @@ def vaccine_fields(vaccination_records, index)
...
     ]
   end
+
+  def coded_site(site)
+    {
+      # These are confirmed
+      left_arm: "LA",
+      right_arm: "RA",
+      # These are made up/guessed
+      left_arm_upper_position: "LUP",
```
benilovj, line 157: I think we really should check this first

tvararu: Should we merge and iterate later?

tvararu: @benilovj I've temporarily disabled the Careplus format via a setting.

thomasleese (approved):

PR #2440, opened 2024-11-14
Add additional information to offline export

thomasleese:
When exporting the information for offline use, we want to include additional information related to triages, gillick assessments, consents, so that the nursing team have as much information available to them as possible when performing vaccinations offline.


tvararu (approved):
Nice, ExportFormatters is a good refactor 👍

PR #2441, opened 2024-11-14
Add support for parsing ISO 8601 dates

thomasleese:
This extends our date parsing to handle another common date format that we might sometimes see in the files provided by our users.


tvararu (approved):
TIL about both `lazy` and `filter_map`

PR #2444, opened 2024-11-14
Update how we generate `kid` for JWKS and update keys in staging and prod.

misaka:
We had this hard-coded as `mavis-int-1` for the INT env, which follows the [guidelines laid out by the NHS for what `kid` to use](https://digital.nhs.uk/developer/guides-and-documentation/security-and-authorisation/application-restricted-restful-apis-signed-jwt-authentication#step-2-generate-a-key-pair).

To support different keys in different envs we need a way to link the kid to the private keys we use, either by pre-generating them and storing them as a jwk, or by using the ruby-jwt gem to do derive it from our private key. The latter approach has been chosen here.

We also update key for test, staging and production environments. The production one has been upgraded to 4096 bits.

The new staging public key has been added to our `jwks` endpoint in the Digital Onboarding, and tested locally.


docs/nhs_api.md
```diff
@@ -5,18 +5,114 @@ Mavis connects to [PDS](pds.md) using the NHS API platform.
...
+[1] pry(main)> pkey = OpenSSL::PKey::RSA.generate(4096)
+=> #<OpenSSL::PKey::RSA:0x0000000134fe3578 oid=rsaEncryption>
+[2] pry(main)> puts pkey.to_pem
+-----BEGIN RSA PRIVATE KEY-----
+...
+```
+
+The pem goes straight into the credentials file under `nhs_api.jwt_private_key`.
+
+### Adding a new key to the NHS API Managemnet JWKS
```
tvararu, line 35:
```suggestion
### Adding a new key to the NHS API Management JWKS
```

tvararu (approved):

PR #2448, opened 2024-11-15
Refactor how organisations are onboarded

thomasleese:
This introduces a new Rake task `onboard` and a YAML configuration file that needs to contain all the information required to onboard an organisation (teams, schools, clinics, programmes, etc) and only if it is all valid does it create the database records.

This helps to avoid situations where the organisations can be partially created, which in the case of missing clinics can be clinically unsafe.

I'm raising this to get a review on the approach taken, but I think there are a number of follow tasks required:

- Document this onboarding process
- Set up our test environments using this process rather than seeds
- Remove Rake tasks which should no longer be used

## Example invalid output

```
❯ bundle exec rails onboard[spec/fixtures/files/onboarding/invalid.yaml]
Programmes can't be blank
Clinics can't be blank
Organisation name can't be blank
Organisation ODS code can't be blank
Organisation phone can't be blank
Organisation phone is invalid
Team name can't be blank
Team email can't be blank
School 0 team can't be blank
School 1 location can't be blank
School 1 team can't be blank
```

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -56,16 +56,20 @@ class Location < ApplicationRecord
 
   validates :name, presence: true
   validates :url, url: true, allow_nil: true
+  validates :urn, uniqueness: true, allow_nil: true
 
-  validates :ods_code, presence: true, if: :clinic?
-
-  with_options if: :generic_clinic? do
+  with_options if: :clinic? do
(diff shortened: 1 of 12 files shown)


tvararu (approved):
I like this approach, let's give it a go 👍

PR #2461, opened 2024-11-16
Easier local testing for clinics

benilovj:
Improve the factories and seeds so that local testing of clinic interactions can be done straight away.

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -35,7 +35,7 @@ class Organisation < ApplicationRecord
   has_many :sessions
   has_many :teams
 
-  has_many :clinics, through: :teams
+  has_many :community_clinics, through: :teams
   has_many :locations, through: :teams
   has_many :patient_sessions, through: :sessions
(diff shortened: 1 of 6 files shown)


app/models/organisation.rb
```diff
@@ -44,6 +44,11 @@ class Organisation < ApplicationRecord
 
   has_and_belongs_to_many :users
 
+  has_many :community_clinic_locations,
```
thomasleese, line 47: We have a `has_many :clinics, through: :teams` above which will return these already plus the generic clinic. However I'm not sure it's actually used anywhere, perhaps we should change that to only return community clinics (maybe renaming it too?) and use that instead, through the team model.

benilovj: Makes sense. I've updated the implementation (see https://github.com/nhsuk/manage-vaccinations-in-schools/pull/2461/commits/ce2dbde89f3867b2a63a9143dd98318d87a56d90) - hopefully this is what you had in mind?

thomasleese (approved):

PR #2470, opened 2024-11-17
Invalidate self-consent when closing a session

thomasleese:
This ensures that the consent isn't available to be used in the clinic that the patient is moved to as self-consent can only be considered valid at a specific location.

### DIFF
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -245,6 +245,15 @@ class Session < ApplicationRecord
         on_duplicate_key_ignore: true
       )
 
+      Consent
+        .via_self_consent
+        .where(
+          patient: unvaccinated_patients,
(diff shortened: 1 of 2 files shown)


app/models/session.rb
```diff
@@ -245,6 +245,11 @@ def close!
         on_duplicate_key_ignore: true
       )
 
+      Consent
```
benilovj, line 248: Is it worth being extra careful and using `policy_scope(Consent)` here?

benilovj (approved):
LGTM, with one non-blocking question.

app/models/session.rb
thomasleese: Hmm yes, that's a good point.

PR #2475, opened 2024-11-17
Attach NHS number to consent form when matching 

tvararu:
If we find a PDS patient with an NHS number that doesn't exist in our database, and no patients match with the consent form, we should update the consent form with the PDS NHS number. This will allow nurses to create a patient record from this consent form at a later stage.

This PR also adds the `nhs_number` optional field to a consent form for this purpose.

This is a prerequisite to https://github.com/nhsuk/manage-vaccinations-in-schools/pull/2454.


thomasleese (approved):
It might be worth adding the normalisation we've got on patients: https://github.com/nhsuk/manage-vaccinations-in-schools/blob/e80605e864fb95d707309d92c0865eacb73a213c/app/models/patient.rb#L154

Perhaps in a concern for both this and patients?

tvararu:
@thomasleese good idea, I've added the call but holding back on making it a concern since it's only used in 2 places

PR #2482, opened 2024-11-18
Store browser sessions in database

thomasleese:
This changes the session store to use an Active Record one, allowing us to store more data in the session than the cookie store allows us to.

Although we're not seeing problems at the moment, I think it's worth doing now as we're starting to create draft vaccination records and consents in the session, there's a risk the cookie will get too big.


misaka:
Somewhere, maybe the SLSP, we documented that we store data on the client browser in the cookie, and had to explain that it's encrypted, etc. Gotta love over-documentation.

tvararu (approved):

benilovj (approved):

PR #2508, opened 2024-11-19
Update production database config

tvararu:
Make all the configs consistent.

Remove the username/password as that's inferred from the DB_SECRET env var in AWS.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,7 +1,7 @@
 default: &default
   adapter: postgresql
   encoding: unicode
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS", 5).to_i + ENV.fetch("GOOD_JOB_MAX_THREADS", 4).to_i %>
 
 development:
   <<: *default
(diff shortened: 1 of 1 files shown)


misaka (approved):

config/database.yml
```diff
@@ -1,7 +1,7 @@
 default: &default
   adapter: postgresql
   encoding: unicode
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i + ENV.fetch("GOOD_JOB_MAX_THREADS", 4).to_i %>
```
benilovj, line 4: It might not really matter, just made me do a double-take that there are two different ways of calling the fetch... or do they need to be different?
```suggestion
  pool: <%= ENV.fetch("RAILS_MAX_THREADS", 5).to_i + ENV.fetch("GOOD_JOB_MAX_THREADS", 4).to_i %>
```

misaka: You're right, the `5` should be the second parameter. I've noticed this before in here, but never bothered to fix. Now's a good time to do that.

PR #2518, opened 2024-11-20
Remove LocationNameConcern

thomasleese:
This was only being used in one model, and I think having this in a separate concern makes the code slightly harder to follow whereas now it's easier to see all the validations in one place when looking at the model.


benilovj (approved):
Yes! The location of this validation completely stumped me on the weekend for a little while...

PR #2522, opened 2024-11-20
Vaccines page layout

paulrobertlloyd:
Brings pages a little closer to design used in prototype:

- Make pages full width within this controller
- Use secondary button style for ‘Add a new batch’ (adding ‘new’ to this text)
- Use normal paragraph style for manufacture

I haven’t update the card header to the design used in the prototype, as the prototype doesn’t currently provide for a link to vaccine details, and I don’t want to change the scope of what’s shown/not shown here.

| Before | After |
|-|-|
| Screenshot before | Screenshot after |

### DIFF
--- a/app/views/vaccines/index.html.erb
+++ b/app/views/vaccines/index.html.erb
@@ -7,9 +7,7 @@
       <%= vaccine.manufacturer %>
     </p>
 
-    <p>
-      <%= link_to "Add a batch", new_vaccine_batch_path(vaccine) %>
-    </p>
+    <%= govuk_button_link_to "Add a new batch", new_vaccine_batch_path(vaccine), class: "app-button--secondary" %>
 
     <% if (batches = @batches_by_vaccine_id[vaccine.id]).present? %>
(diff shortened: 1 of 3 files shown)


app/views/vaccines/index.html.erb
```diff
@@ -3,13 +3,11 @@
 <% @vaccines.each do |vaccine| %>
   <%= render AppCardComponent.new do |c| %>
     <% c.with_heading { link_to(vaccine_heading(vaccine), vaccine_path(vaccine)) } %>
-    <p class="nhsuk-body-s nhsuk-secondary-text-color">
+    <p class="nhsuk-body">
```
tvararu, line 6: I don't think `nhsuk-body` does anything in our app because we have global styles for paragraphs 🤔

tvararu (approved):

app/views/vaccines/index.html.erb
paulrobertlloyd: Fair enough, will remove.

PR #2538, opened 2024-11-21
Conditionally show links in footer

paulrobertlloyd:
- Service guidance should only be shown to users of ‘Manage vaccinations in schools’
- The privacy policy should only be shown to users of ‘Give or refuse consent for vaccinations’ (if a privacy policy URL has been set for the organisation)

This PR makes the existence of values for `@service_guide_url` and `@privacy_policy_url` determine whether links to these resources are shown in the footer, with controllers setting either the appropriate values for these resources, or `false`.

This also fixes an issue with an empty `li` being shown if no `@privacy_policy_url` is set.

Also renames the remaining `guide.show.title` locale string, moving it alongside the other `service` locale key to `service.guide.title`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,6 +10,7 @@ class ApplicationController < ActionController::Base
   before_action :set_disable_cache_headers
   before_action :set_header_path
   before_action :set_service_name
+  before_action :set_service_guide_url
   before_action :set_secondary_navigation
   before_action :set_privacy_policy_url
   before_action :authenticate_basic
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -59,8 +60,11 @@ def user_not_authorized
...
   end
 
+  def set_service_guide_url
+    @service_guide_url = "https://guide.manage-vaccinations-in-schools.nhs.uk"
+  end
+
   def set_privacy_policy_url
-    @privacy_policy_url =
-      current_user&.selected_organisation&.privacy_policy_url
+    @privacy_policy_url = false
```
thomasleese, line 68:
```suggestion
    @privacy_policy_url = nil
```

Not major, but I think this would be more typical than setting it to `false`.

paulrobertlloyd: Fixed (and for `@service_guide_url` in the other controller)

thomasleese (approved):

tvararu (approved):

PR #2549, opened 2024-11-23
Override load balancer names

misaka:
By default copilot (or AWS) creates our load balancers with a name that has a double-dash. This is causing issues with NHS's DNS provider, their systems isn't able to setup a DNS record with a double-dash in it, supposedly citing RFC5891, which doesn't actually apply to these names.

This change sets the name of the load balancers to a name without a double-dash, and also with the environment name in it which fixes a small niggle we had when viewing these resources in AWS where it was difficult to know which environment a load balancer belonged to.

This has been tested by setting up and deploying out the `lbtest` environment (replacing `accessibility`). It appears to have the desired result of setting the name of the load balancer.

image

I've performed 3 deploys now and copilot doesn't seem to have a problem with this naming.

This change won't take on existing environments, we'll have to recreate them for this new ALB name to take effect. There is the option to create a new load balancer manually, using this naming convention, and then replacing the old ALB with it. But that seems a bit fiddly and I would prefer to just re-create the environment where possible.


copilot/environments/overrides/cfn.patches.yml
```diff
@@ -40,3 +40,6 @@
   path: /Resources/EnvironmentHTTPSSecurityGroupIngressFromPublicALB
 - op: remove
   path: /Resources/EnvironmentSecurityGroupIngressFromSelf
+- op: add
+  path: /Resources/PublicLoadBalancer/Properties/Name
+  value: !Sub ${EnvironmentName}-${AppName}-pub-lb
```
tvararu, line 45: Okay, this is much easier than expected!

tvararu (approved):

copilot/environments/overrides/cfn.patches.yml
misaka: Ya, it's surprisingly ... ok.

misaka:
I propose we merge this tomorrow at the earliest. Get some rest and see if we still think this is the right thing tomorrow.

misaka:
Some additional details.

The CloudFormation resource specifier generated by copilot for the load balancers look like this:

```
    PublicLoadBalancer:
        Metadata:
            'aws:copilot:description': 'An Application Load Balancer to distribute public traffic to your services'
        Condition: CreateALB
        Type: AWS::ElasticLoadBalancingV2::LoadBalancer
        Properties:
            LoadBalancerAttributes:
                - Key: 'access_logs.s3.enabled'
                  Value: false
            Scheme: internet-facing
            SecurityGroups:
                - !GetAtt PublicHTTPLoadBalancerSecurityGroup.GroupId
                - !If [ExportHTTPSListener, !GetAtt PublicHTTPSLoadBalancerSecurityGroup.GroupId, !Ref "AWS::NoValue"]
            Subnets: [!Ref PublicSubnet1, !Ref PublicSubnet2]
            Type: application
```

This addition adds one line to it:

```diff
--- before.cf	2024-11-23 11:23:23
+++ after.cf	2024-11-23 11:23:32
@@ -13,3 +13,4 @@
                 - !If [ExportHTTPSListener, !GetAtt PublicHTTPSLoadBalancerSecurityGroup.GroupId, !Ref "AWS::NoValue"]
             Subnets: [!Ref PublicSubnet1, !Ref PublicSubnet2]
             Type: application
+            Name: !Sub ${EnvironmentName}-${AppName}-pub-lb
```

From this, it would follow that it's not copilot that's generating the name for the load balancer, but AWS or maybe CloudFormation. The aws cli states that the `name` must be provided, but CloudFormation's documentation says:

> Name
>
> The name of the load balancer. This name must be unique per region per account, can have a maximum of 32 characters, must contain only alphanumeric characters or hyphens, must not begin or end with a hyphen, and must not begin with "internal-".
>
> If you don't specify a name, AWS CloudFormation generates a unique physical ID for the load balancer. If you specify a name, you cannot perform updates that require replacement of this resource, but you can perform other updates. To replace the resource, specify a new name.

So I believe the name generated `mavis--*` must come from CloudFormation.

misaka:
The possible issues I can see with this change are:

1. What if copilot needs to refer to the load balancer? Will it be able to find it? I don't think this is an issue as the default is to let CloudFormation generate a random name. copilot does add tags to the load balancer, so if necessary, e.g. on deletion, it may just use that.
2. Is creating a LB without a name different from giving it a name. The CloudFormation docs say that if you supply a name, you "cannot perform updates that require replacement of this resource". I'm not sure this is an issue but we may need to investigate this.

Also, this begs the question, why has copilot opted not to specify the name of the load balancer as we have here, it seems it would be much more convenient by informing any users of which environment it belongs to. I explored the potential issues with a robot friend recently and basically came up with this kind of scenario where it could be a problem:

1. We create an environment with the lb and with the given name.
2. We need to recreate the environment, and start by deleting the old one.
3. We then begin to create the new environment, but the deletion of the resources in the old one haven't completed, and 💥 we get a clash when we try to create a load balancer with the pre-defined name.

Essentially issues may come from the fact that LB names must be unique to the account and the region they are in.

I think we're unlikely to run into this issue, and if we do I feel like we'll be able to work around it.

PR #2553, opened 2024-11-25
Add www. to hostnames

misaka:
Previously we had the root domain, 'manage-vaccinations-in-schools.nhs.uk', CNAMEd to our app so that users can hit it and be redirected to the 'www.'. This is not RFC compliant, and we've been told by NHS's DNS team that we can't do that. Apparently there are ways to do this, but it can depend on the provider and in our case, they simply setup A records to point to the IPs of our LBs at the time, which can go (and have gone) stale.

Until we find a way to properly CNAME the root domain we need to change our settings to use the 'www' hostname.

### DIFF
--- a/copilot/webapp/manifest.yml
+++ b/copilot/webapp/manifest.yml
@@ -106,5 +106,5 @@ environments:
     variables:
       RAILS_ENV: production
       SENTRY_ENVIRONMENT: production
-      MAVIS__HOST: "manage-vaccinations-in-schools.nhs.uk"
-      MAVIS__GIVE_OR_REFUSE_CONSENT_HOST: "give-or-refuse-consent-for-vaccinations.nhs.uk"
+      MAVIS__HOST: "www.manage-vaccinations-in-schools.nhs.uk"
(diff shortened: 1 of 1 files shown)


misaka:
CloudFlare's implementation of root-level CNAMEs is interesting. Apparently they resolve the CNAME chain behind the scenes to effectively replicate the CNAME behaviour that name servers would perform, without breaking the spec.

> To accomplish this, we extended our authoritative DNS infrastructure to, in certain cases, act as a kind of DNS resolver. What happens is that, if there's a CNAME at the root, rather than returning that record directly we recurse through the CNAME chain ourselves until we find an A Record. At that point, we return the IP address associated with the A Record. This, effectively, "flattens" the CNAME chain.

https://blog.cloudflare.com/introducing-cname-flattening-rfc-compliant-cnames-at-a-domains-root/

misaka:
Also on the topic of "how do other DNS providers manage to do this?", this post mentions that some will just create a static A record:

https://serverfault.com/questions/1092725/domain-registrar-allows-cname-for-root-domain-is-this-reliable

Obviously this can be unreliable as the IP pointed-to may change, and indeed this explains what we saw with our old root domains that we requested be setup as CNAMEs, but ended up pointing to different IPs from the AWS ALBs we had provisioned in that env.

PR #2557, opened 2024-11-25
Use real numbers in smoke test setup

misaka:
Production needs real phone numbers for validations


thomasleese (approved):

thomasleese:
I'm going to merge this in as it's what we deployed in production.

PR #2574, opened 2024-11-27
Refactor locations factories

thomasleese:
This changes the factories for locations to use separate factories for the three types of locations, schools, community clinics and generic clinics. This makes it impossible to use the school traits in clinics and vice-versa, by introducing a clear separation between the types of locations.

I also think this helps to make the code that uses these factories clearer and easier to read.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -20,8 +20,8 @@ end
 
 def import_schools
   if Settings.fast_reset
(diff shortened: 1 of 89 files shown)


tvararu (approved):

tvararu:
I'm getting this into main so I can rebase `release-1c`

PR #2597, opened 2024-12-02
Freeze header and patient name cells

thomasleese:
When exporting the offline session spreadsheet this ensures that the header row, and the first two columns (the patient first name and last name) are frozen, meaning that they are always visible when scrolling and navigating around the spreadsheet. This should make it easier for nurses to use.


tvararu (approved):
Nice

PR #2598, opened 2024-12-02
Update production basic auth credentials

tvararu:
Change from the default set of credentials that are used in the other environments.


misaka (approved):
Checked locally, looks good. 🚢

PR #2640, opened 2024-12-04
Refactor vaccination flow to match prototype

thomasleese:
This changes the vaccination form shown at the bottom of the patient page to ask only whether the patient is ready to be vaccinated and then ask the user to specify the delivery location after choosing an option.

This makes the necessary changes in preparation for adding the new pre-screening questions.

## Screenshot

Screenshot 2024-12-04 at 14 20 15

### DIFF
--- a/app/components/app_patient_page_component.html.erb
+++ b/app/components/app_patient_page_component.html.erb
@@ -121,6 +121,6 @@
   <% end %>
 <% end %>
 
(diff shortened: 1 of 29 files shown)


paulrobertlloyd:
Is the current behaviour here to show the site option in a conditional if selecting yes? If so, we should keep this. The prototype is a regression I’ve been meaning to revert!

thomasleese:
Replaced by #2641

PR #2663, opened 2024-12-06
Replace annotate with annotaterb

tvararu:
`annotate` is not compatible with Rails 8.

`annotaterb` is a near drop-in replacement.


lib/tasks/annotate_rb.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# This rake task was added by annotate_rb gem.
+
+# Can set `ANNOTATERB_SKIP_ON_DB_TASKS` to be anything to skip this
+if Rails.env.development? && ENV["ANNOTATERB_SKIP_ON_DB_TASKS"].nil?
+  require "annotate_rb"
```
thomasleese, line 7: Do we need this if we're including it in the Gemfile without `require: false`?

thomasleese (approved):

PR #2669, opened 2024-12-06
Refactor Dockerfile

tvararu:
This aligns it closer to the one from a fresh `rails new` app made with Rails 8.

Also set `NODE_ENV=production` which stops us from installing `devDependencies` in the Docker image, saving some time downloading them.


thomasleese (approved):
Nice!

PR #2671, opened 2024-12-09
Look up patients when NHS number doesn't exist

thomasleese:
This expands upon https://github.com/nhsuk/manage-vaccinations-in-schools/commit/e0aaf91f3e79bc121e52558a235a90c54a049652 to also look up the patient by demographics if the NHS number doesn't exist in PDS. This could help to handle the case where the nurse has entered an NHS number incorrectly.

https://good-machine.sentry.io/issues/6091497694/

### DIFF
--- a/app/jobs/patient_update_from_pds_job.rb
+++ b/app/jobs/patient_update_from_pds_job.rb
@@ -5,8 +5,6 @@ class PatientUpdateFromPDSJob < ApplicationJob
 
   queue_as :pds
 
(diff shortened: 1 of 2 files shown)


thomasleese:
Without this change, https://trello.com/c/OrfiwKDt/1641-patients-with-bad-wrong-nhs-numbers-should-be-re-traced-using-demographic-details hasn't actually been fixed for this release, but happy to delay merging this in if we need more time to test it.

tvararu (approved):

PR #2682, opened 2024-12-09
Update Splunk index in production

tvararu:
We accidentally ended up using the nonprod one for production, so our Splunk logs were going to the wrong place.


misaka (approved):

misaka:
🚀

PR #2687, opened 2024-12-09
Convert aws-account-setup.sh to ruby

misaka:
Ruby code is easier to maintain and makes it easier for us to do more complex operations, like fixing some of the policies instead of spitting out instructions for the user to follow.

Also adds functionality to remove any inbound rules from the app's VPC default security groups to ensure we can't accidentally apply it to resources and inadvertently give access. Since copilot manages our security groups for us we can assume we'll always have one setup instead of relying on the default.

Fixes `NCC-E017705-HUW` from the pentest.

Sample run:

image


tvararu (approved):
#YOLO

PR #2696, opened 2024-12-10
Upgrade to Rails 8.0

thomasleese:
This upgrades to the latest version of Rails, 8.0.

The biggest change for us would be the introduction of [the Thurster proxy server](https://github.com/basecamp/thruster) that we'll use when deploying the app to production, which means our container will be listening to requests on port 80 rather than port 3000.

The other changes are relatively minor and shouldn't have any major impacts on the service, see individual commits for more details.


bin/setup
```diff
@@ -29,6 +28,9 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Removing old logs and tempfiles =="
   system! "bin/rails log:clear tmp:clear"
 
-  puts "\n== Restarting application server =="
-  system! "bin/rails restart"
+  unless ARGV.include?("--skip-server")
+    puts "\n== Starting development server =="
+    STDOUT.flush # flush the output before exec(2) so that it displays
+    exec "bin/dev"
+  end
```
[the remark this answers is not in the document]
tvararu: I pre-emptively aliased `bin/setup` to `bs` for this:

```bash
abbr --add bs "bin/setup --skip-server"
```

bin/dev
```diff
@@ -1,8 +1,2 @@
-#!/usr/bin/env sh
-
-if ! gem list foreman -i --silent; then
-  echo "Installing foreman..."
-  gem install foreman
-fi
-
-exec foreman start -f Procfile.dev "$@"
+#!/usr/bin/env ruby
+exec "./bin/rails", "server", *ARGV
```
tvararu, line 2: Come to think of it, this should probably still attempt to use `Procfile.dev` via `foreman`.

config/environments/production.rb
```diff
@@ -60,14 +47,18 @@
...
   config.log_tags = { request_id: :request_id }
 
-  # "Info" include generic and useful information about system operation, but avoids logging too much
-  # information to avoid inadvertent exposure of personally identifiable information (PII). If you
-  # want to log everything, set the level to "debug".
+  # Change to "debug" to log everything (including potentially personally-identifiable information!)
   config.log_level = ENV.fetch("RAILS_LOG_LEVEL", "info")
 
+  # Prevent health checks from clogging up the logs.
+  config.silence_healthcheck_path = "/up"
```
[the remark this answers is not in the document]
thomasleese: Fair enough, I was trying to keep these files as close to standard Rails as possible, I might put it in `environments/development.rb` instead of `application.rb`.

bin/setup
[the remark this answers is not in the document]
thomasleese: If you're happy to make the change to your local setup, I would prefer to stick to the Rails defaults here, but don't feel too strongly.

bin/thrust
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env ruby
+require "rubygems"
+require "bundler/setup"
+
+load Gem.bin_path("thruster", "thrust")
```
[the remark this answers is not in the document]
thomasleese: I've added a commit that configures the app to run via Thruster.

bin/dev
thomasleese: Yep, agreed.

copilot/environments/overrides/cfn.patches.yml
```diff
@@ -29,13 +29,13 @@
...
-  value: HTTP ingress from the public ALB on port 4000
+  value: HTTP ingress from the public ALB on port 80
 - op: add
   path: /Resources/EnvironmentHTTPSecurityGroupIngressFromPublicALB/Properties/FromPort
-  value: 4000
+  value: 80
 - op: add
   path: /Resources/EnvironmentHTTPSecurityGroupIngressFromPublicALB/Properties/ToPort
-  value: 4000
+  value: 80
```
misaka, line 38: Do we know what we'll need to do to make these changes take? We'll need to be sure the load balancers forwarders are re-written, that may require an env deploy if not more. Worth testing.

thomasleese: Not sure, I wasn't very confident making the changes to this file, one option could be to extract the change to use Thruster to a separate PR so we can at least switch to Rails 8.0 first.

tvararu: Tbh, I think we don't have to use port 80. Most operating systems require root to attach something to port 80 -- could make local dev harder for no reason, say if someone was using Docker to run the app. Let's stick to 4000 which works fine.

tvararu: We could switch to 80 in a separate PR if we think it's worth it.

misaka: Maybe let's test a deploy from this branch to one of the envs?

thomasleese: > Tbh, I think we don't have to use port 80. 

Yeah, this is another reasonable option. I switched to 80 to match the Rails defaults, but we could stick with 4000.

tvararu: Yeah, I had trouble in a separate project getting Fly to work on 80 so switched to 3000 which is the previous default. It's not a big difference.

The reason we use 4000 locally is to allow the prototype to run on 3000.

thomasleese: I've configured Thruster to run on port 4000.

config/environments/development.rb
```diff
@@ -78,23 +85,18 @@
   # Raise error when a before_action's only/except options reference missing actions.
   config.action_controller.raise_on_missing_callback_actions = true
 
+  # Apply autocorrection by RuboCop to files generated by `bin/rails generate`.
+  # config.generators.apply_rubocop_autocorrect_after_generate!
```
tvararu, line 89: I like this new option, we could enable it now or later

tvararu (approved):
:shipit:

config/environments/development.rb
thomasleese: Yeah, I wondered about that too. Decided to leave it commented out since we're going to need to run Prettier on any files too so it doesn't safe us any steps.

PR #2707, opened 2024-12-10
Don't validate dose_sequence if not administered

tvararu:
A vaccine that isn't administered doesn't have a batch, which means it doesn't have a vaccine. Without the vaccine, `maximum_dose_sequence` defaults to 1. If the `dose_sequence` is > 1, this breaks the validation.

We can also remove the `vaccine& || 1` default with this change.

### DIFF
--- a/app/models/vaccination_record.rb
+++ b/app/models/vaccination_record.rb
@@ -144,12 +144,13 @@ class VaccinationRecord < ApplicationRecord
               if: :requires_location_name?
             }
 
+  validates :dose_sequence, presence: true
   validates :dose_sequence,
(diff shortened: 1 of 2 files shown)


thomasleese (approved):
I’m wondering if it makes sense to store a dose sequence on not administered vaccinations, but this fix sounds good for now.

benilovj:
> I’m wondering if it makes sense to store a dose sequence on not administered vaccinations, but this fix sounds good for now.

We had the same thought, but dose sequence is currently a non-nullable field in the DB and removing that restriction would be a bigger incision.

PR #2733, opened 2024-12-12
Include Sentry SDK in staging envs

tvararu:
This enables a feature called Session Replay, which in theory should give us a blow-by-blow list of user interactions that lead to an error. This could be useful for debugging.

This doesn't affect development/test/production, only the staging environments.


tvararu:
Closing this as while it works, it's only valid for JavaScript/frontend errors, and since we use progressive enhancement, those are rarely ever show stopping. It doesn't do anything for 500s.

PR #2745, opened 2024-12-13
Choose an organisation on login without CIS2

tvararu:
We currently default to `current_user.organisations.first` if users aren't signed in using CIS2 when deciding their `selected_organisation`.

This makes testing certain features harder, as there is no way for users to sign into multiple organisations. Amongst other things, we could use this capability in the training environment to set up 2 different organisations; one that is a blank slate, one that has data in it, to train on different scenarios.

This adds a page that simulates the "Select a role" page that users see when they login via CIS2. However, it asks the users to choose an organisation, not a role. The user is only asked on login, or during their current session if they haven't chosen one already. In environments where CIS2 is enabled, this functionality should not affect anything.

### Choosing an organisation on login

Screenshot 2024-12-13 at 17 52 15

### DIFF
--- a/app/controllers/users/organisations_controller.rb
+++ b/app/controllers/users/organisations_controller.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+class Users::OrganisationsController < ApplicationController
+  skip_before_action :set_selected_organisation
+  skip_after_action :verify_policy_scoped
+
+  before_action :redirect_to_dashboard_if_cis2_is_enabled
+
+  layout "two_thirds"
+
+  def new
+    @organisations = current_user.organisations
+  end
+
+  def create
+    organisation = current_user.organisations.find(params[:organisation_id])
+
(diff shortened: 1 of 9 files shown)


app/controllers/users/organisations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    organisation = current_user.organisations.find(params[:organisation_id])
+
+    if organisation.present?
+      session["cis2_info"] = {
+        "selected_org" => {
+          "name" => organisation.name,
+          "code" => organisation.ods_code
+        },
+        "selected_role" => {
+          "code" => valid_cis2_roles.first
```
thomasleese, line 25: I wonder if we should include the `workgroups` here? Just so it matches what we'll get from CIS2 and helps with local testing. In the future we could also add the option to sign in as an admin maybe as well?

thomasleese (approved):

PR #2747, opened 2024-12-13
Refactor GOV.UK Notify personalisation

thomasleese:
This refactors the personalisation generation for GOV.UK Notify to split the `GovukNotifyPersonalisation` class in to two separate classes (`GovukNotifyParameters` and `GovukNotifyPersonalisation`). This allows the `ApplicationMailer` and `TextDeliveryJob` to use the `GovukNotifyParameters` class to extract a suitable `organisation`, `parent` or `consent_form` from the input parameters to the job.


misaka (approved):

thomasleese:
This will be refactored in a different way to support tracking failed emails.

PR #2755, opened 2024-12-16
Add hertfordshire onboarding ymls

tvararu:
These will be used to set up the training environment.

Also make some updates to the onboarding script to allow running two different orgs in parallel with the same `nurse.x` user.

### DIFF
--- a/app/models/onboarding.rb
+++ b/app/models/onboarding.rb
@@ -34,7 +34,9 @@ class Onboarding
       config
         .fetch(:users, [])
         .map do |user_config|
-          User.new(**user_config, organisations: [organisation])
+          User
+            .find_or_initialize_by(email: user_config[:email])
+            .tap { |user| user.assign_attributes(user_config) }
         end
 
     @schools =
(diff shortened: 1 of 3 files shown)


app/models/onboarding.rb
```diff
@@ -7,7 +7,6 @@ class Onboarding
   validates :programmes, presence: true
   validates :teams, presence: true
   validates :schools, presence: true
-  validates :clinics, presence: true
```
thomasleese, line 10: I don't think we should make this change, one of the reasons the onboarding process was added to make it impossible to set up an organisation without clinics.

Without clinics, it's impossible to vaccinate patients in the "Community clinic" session as the nurse will be asked for which location and [there won't be any options to choose from](https://trello.com/c/XK3ye26w/1418-clinic-vaccinations-cannot-be-recorded-if-the-sais-org-is-missing-clinic-locations).

To prepare for the training tomorrow can we create a fake clinic for now?

tvararu: Yeah, I'll update with a fake clinic then 👍

PR #2775, opened 2024-12-18
Show health answers from latest consents

thomasleese:
This avoids showing health answers for consents that have been invalidated or overridden by a newer consent response from the same person.


tvararu (approved):

thomasleese:
It's possible the approach here might not be right, further discussion required before we can merge this.

misaka (approved):

PR #2780, opened 2024-12-19
Allow specifying year group when importing

thomasleese:
This adds the ability to upload a class list or a cohort with the year group for each patient optionally specified in a `CHILD_YEAR_GROUP` column. When specified, the year group will be taken from this value rather than calculated from the date of birth.

Internally, we don't store the year group because this changes each year, instead we store the `birth_academic_year` (similar to the approach taken for cohorts) which stores the year of birth in relation to the current academic year.

Patients are then assigned a cohort according to their `birth_academic_year` value. We can't entirely rely on cohorts, as we need to track this value even if a patient were to move between organisations. Due to this refactor, we could eventually remove the `Cohort` model and instead group patients by their `birth_academic_year` and have a direct relationship between a patient and an organisation.

## Screenshots

Screenshot 2024-12-19 at 11 27 25

Screenshot 2024-12-19 at 11 30 05

Screenshot 2024-12-19 at 11 30 39


tvararu (approved):

misaka (approved):
🚢

PR #2787, opened 2024-12-20
Ensure parents provide contact method with phone number

thomasleese:
If a parent adds a phone number after the consent confirmation page we also need to ensure that they also provide a contact method otherwise they're unable to submit the consent form.

https://good-machine.sentry.io/issues/6049268593/


app/controllers/dev_controller.rb
```diff
@@ -66,26 +66,43 @@ def reset_organisation
...
+    request.session[:consent_form_id] = consent_form.id
+    redirect_to confirm_parent_interface_consent_form_path(consent_form)
   end
 
   private
 
   def ensure_dev_env_or_dev_tools_enabled
-    unless Rails.env.development? || Rails.env.test? ||
-             Flipper.enabled?(:dev_tools)
+    unless Rails.env.local? || Flipper.enabled?(:dev_tools)
```
tvararu, line 105: :shipit:

thomasleese: I knew you'd like that!

misaka (approved):

PR #2797, opened 2024-12-25
Class list uploads: make it easier to tell which school the upload is for, and to navigate back to the school

benilovj:
## Upload list

|| Before | After |
|--------|--------|--------|
|Upload list| image | image |
|Upload page| image | image |

### DIFF
--- a/app/components/app_imports_table_component.html.erb
+++ b/app/components/app_imports_table_component.html.erb
@@ -31,6 +31,12 @@
           <% row.with_cell do %>
             <span class="nhsuk-table-responsive__heading">Import type</span>
             <%= record_type(import) %>
+            <% if import.is_a?(ClassImport) %>
+              <br />
+              <span class="nhsuk-u-secondary-text-color nhsuk-u-font-size-16">
+                <%= import.session.location.name %>
+              </span>
+            <% end %>
           <% end %>
           <% row.with_cell do %>
             <span class="nhsuk-table-responsive__heading">Status</span>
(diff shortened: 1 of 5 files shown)


app/components/app_imports_table_component.html.erb
```diff
@@ -31,6 +31,13 @@
           <% row.with_cell do %>
             <span class="nhsuk-table-responsive__heading">Import type</span>
             <%= record_type(import) %>
+            <% if import.is_a?(ClassImport) %>
+              <br />
+              <span class="nhsuk-u-secondary-text-color nhsuk-u-font-size-16">
+                <%= govuk_link_to import.session.location.name,
```
paulrobertlloyd, line 37: Looks good. Only question I have is does the school session need to be linked to from the table view?

tvararu (approved):

thomasleese (approved):

app/components/app_imports_table_component.html.erb
benilovj: Yeah, fair.

I think I originally added it because:

* I remember from observing the first couple of days of usage, staff were often uploading the class lists and trying to get back to the school page, often going via sessions > un/scheduled > and trying to find the school in a school list, which looked a bit frustrating.
* I think there's benefit in every mention of a school linking to the school page, similar to how (almost) all mentions of a patient are links to either the patients's global page or session page.

Having said that, I'm not convinced that the first two days are necessarily representative of regular usage, and once staff discover the link to the school from the individual upload page, they'll (hopefully) begin using that...

I've backed the link out for now.

PR #2798, opened 2024-12-27
Update erroneously added primary school URN to the correct secondary school URN

benilovj:


thomasleese (approved):

tvararu (approved):

thomasleese:
This change has already been applied in production so this is safe to merge in with no effect on the functionality.

PR #2799, opened 2024-12-28
Upgrade to Ruby 3.4.1

thomasleese:
This is the latest version of Ruby that was released on the 25th December 2024:

https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-1-released/


tvararu (approved):

tvararu:
https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/

tvararu:
We can also remove `# frozen_string_literal: true` from all files now.

thomasleese:
> We can also remove `# frozen_string_literal: true` from all files now.

I looked in to this but I'm not sure we can yet. With the comment we get the future behaviour where strings can't be mutated, however if we remove the comment, the Ruby 3.4 behaviour is to allow mutation but warn about deprecation in the future.

If we remove the comment I'm wondering if we'd lose the performance improvements that come with immutable strings.

misaka:
> > We can also remove `# frozen_string_literal: true` from all files now.
> 
> I looked in to this but I'm not sure we can yet. With the comment we get the future behaviour where strings can't be mutated, however if we remove the comment, the Ruby 3.4 behaviour is to allow mutation but warn about deprecation in the future.
> 
> If we remove the comment I'm wondering if we'd lose the performance improvements that come with immutable strings.

Ya, I read this too.

> Now files without the frozen_string_literal comment will treat string literals as “chilled”, which means they can still be mutated, but it will emit a warning on mutation.

https://nithinbekal.com/posts/ruby-3-4/

Almost there.

misaka (approved):

PR #2804, opened 2024-12-30
Fix performing professionals dropdown in offline spreadsheets

misaka:
In Excel spreadsheets when the list of available options for data validation is too large this can cause issues.

    https://stackoverflow.com/questions/16017403/maximum-drop-down-list-formula-length-in-excel

To avoid this we can move the list of options to another sheet, table or named range.

    https://contextures.com/xldataval11.html

This change moves the list of emails of the performing professionals to it's own sheet and adds a formula to use them as the source data for the data validation for the PERFORMING_PROFESSIONAL_EMAIL field of the offline spreadsheet.

[Fixes MAVIS-1800](https://trello.com/c/FZvpVCEk/1800-offline-spreadsheet-opens-with-errors-when-many-vaccinators-present-in-the-organisation?filter=offline)

### DIFF
--- a/app/lib/reports/offline_session_exporter.rb
+++ b/app/lib/reports/offline_session_exporter.rb
@@ -10,8 +10,12 @@ class Reports::OfflineSessionExporter
   end
 
   def call
+    # stree-ignore
     Axlsx::Package
-      .new { |package| add_vaccinations_sheet(package) }
+      .new { |package|
+        add_vaccinations_sheet(package)
+        add_performing_professionals_sheet(package)
+      }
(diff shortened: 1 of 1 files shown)


benilovj (approved):

tvararu (approved):

benilovj:
@misaka looks like one of the tests that's failing on this branch also fails on `main`; I have a fix here: https://github.com/nhsuk/manage-vaccinations-in-schools/pull/2805

benilovj (approved):

PR #2807, opened 2025-01-04
Add .cursorrules

tvararu:
[`.cursorrules`](https://docs.cursor.com/context/rules-for-ai#cursorrules) is a dotfile that contains project-specific instructions that automatically get passed into Cursor editor's context window when querying LLMs.

It can also be manually specified when using other editors/extensions.

At some point a more standard (e.g. `.llmrules`) dotfile will probably emerge to replace this.


thomasleese (approved):
I don't use this personally, but no major problem with having it in the repo.

PR #2815, opened 2025-01-07
Set up `next` branch deployment workflow

thomasleese:
This makes a number of changes to the GitHub Actions and Dependabot configuration to support the workflow we're going to try going forward.

- Dependabot will raise PRs against the `next` branch, which will the branch used when merging in features for the next release, and therefore need to be tested.
- `next` will automatically deploy to the `qa` and `test` environments so the latest changes are always available to the testers.
- When we're ready to make a release, `next` will be fast-forward merged in to `main`, which triggers an automatic deploy to `preview`, ready to be manually pushed to `training` and `production` when going live.

Worth noting that the final step is currently not possible with our branch protection rules so we'll need to find a way around this other than temporarily turning off/on the rules long term, but for now I think it's okay to leave it as we test this approach.


tvararu (approved):

thomasleese:
We're going with a slightly different approach where we keep `main` our default branch, ahead of a `release` branch which contains the latest release, so instead of merging changes in to `next`, we'll merge straight in to `main`.

PR #2789, opened 2024-12-20
Fix bug where nurses can't sort with blank postcodes

thomasleese:
This fixes a bug where if a user tries to sort on a table by postcode and some of the rows don't have a postcode, the sorting fails.

https://good-machine.sentry.io/issues/6164317867/

PR #2817, opened 2025-01-07
Fix bug where nurses can't sort with blank postcodes

misaka:
This fixes a bug where if a user tries to sort on a table by postcode and some of the rows don't have a postcode, the sorting fails.


misaka:
#2789 was accidentally merged to `main` when it should have gone to `next`.

thomasleese (approved):

PR #2822, opened 2025-01-07
Show environment banner on staging environments

paulrobertlloyd:
Styles and markup for an `AppHostingEnvironmentComponent` that renders a banner above the header on all staging environments, giving an environment’s name and assigned colour, default to staging/grey if this cannot be determined.

- Largely inspired by what was used on Apply, meaning parts of this are likely incorrect; do we use `HOSTING_ENVIRONMENT_NAME` variable? If not, where can I get the named environment?
- Do we also have an Heroku review environment that needs to be accounted for?
- Does it need any tests?

Also deleted the custom styling for white tags as we don’t use that anymore. 

## Banners

Development banner.
Staging banner.
Test banner.
QA banner.
Preview banner.
Training banner.

### DIFF
--- a/app/components/app_hosting_environment_component.rb
+++ b/app/components/app_hosting_environment_component.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+class AppHostingEnvironmentComponent < ViewComponent::Base
+  erb_template <<-ERB
+    <div class="app-environment nhsuk-tag--<%= colour %>">
+      <div class="nhsuk-width-container">
+        <strong class="nhsuk-tag nhsuk-tag--<%= colour %>">
+          <%= title %>
+        </strong>
+        <span><%= t("hosting_environment", name:) %></span>
+      </div>
+    </div>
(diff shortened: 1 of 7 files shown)


paulrobertlloyd:
Thanks for the spruce-up @thomasleese!

app/components/app_hosting_environment_component.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    environment
+  end
+
+  def colour
+    ENVIRONMENT_COLOR[environment.to_sym]
+  end
+
+  def environment
+    ENV.fetch("SENTRY_ENVIRONMENT", "development")
```
thomasleese, line 42: Since this is the only place we're using this environment variable (other than Sentry itself) I thought it was okay to use it here, but if we start needing this in more places I suggest we have a `HOSTING_ENVIRONMENT` variable and configure Sentry to use that instead.

tvararu (approved):

PR #2824, opened 2025-01-08
Always send text messages for consent requests/reminders

thomasleese:
This updates the `ConsentNotification` model to always send a text message to a parent when sending consent requests and reminders, not taking in to account whether the parent has opted-in to receive updates on the vaccination.

Instead, opting-in relates to other updates like consent confirmation, rejected, or vaccination given/not given updates.

To make this change I've had to update the `TextDeliveryJob` to not take this value in to account and instead everywhere that the job is enqueued considers whether to send the text or not based on the context and whether the parent has opted-in.


misaka:
Fixes MAVIS-1829

tvararu (approved):

misaka (approved):

PR #2829, opened 2025-01-10
Ensure `parent_full_name` personalisation is present

thomasleese:
This variable is used when sending some emails (specifically the consent triage confirmation email), and isn't not guaranteed that the parent will have a name attached to their record. In that case, we still need to send a value to GOV.UK Notify otherwise we get back a `BadRequestError` about a missing personalisation variable.


tvararu (approved):

thomasleese:
Merging this in as the fix is live in production, we've removed the `parent_full_name` variable from all the templates, so this has effect on the 1.3.0 release.

PR #2841, opened 2025-01-14
Use GOV.UK Notify directly to send emails

thomasleese:
This adds a new job, similar to the `SMSDeliveryJob` that allows us to send emails using GOV.UK Notify directly rather than going via `ActionMailer`. Using the API directly allows us to record the delivery message ID, which we can then use to get the status of the receipt later on and present this information to users.

By switching to this job we get automatic retry behaviour on GOV.UK Notify server errors, and tracking of message IDs.


tvararu (approved):
Nice!

PR #2846, opened 2025-01-14
Add ability to mark unmatched consent forms as invalid

thomasleese:
This adds some new functionality allowing nurses to mark unmatched consent forms as invalid, meaning that they will disappear from the list of unmatched consent responses reducing work for the SAIS team who need to look through the list.

We know that sometimes consent responses are provided for patients who aren't in the cohort, so the only action to take with these is to mark them as invalid. Although we will stop showing them on the page, we will continue to store the information internally so we can surface them later.

## Screenshots

Screenshot 2025-01-14 at 15 36 49
Screenshot 2025-01-14 at 16 07 04
Screenshot 2025-01-14 at 16 07 10

### DIFF
--- a/app/controllers/consent_forms_controller.rb
+++ b/app/controllers/consent_forms_controller.rb
@@ -45,6 +45,24 @@ class ConsentFormsController < ApplicationController
     redirect_to action: :index
   end
 
+  def edit_invalidate
+    render :invalidate, layout: "two_thirds"
+  end
+
+  def update_invalidate
+    @consent_form.assign_attributes(invalidate_params)
+
+    if @consent_form.save
+      redirect_to consent_forms_path,
+                  flash: {
+                    success:
+                      "Consent response from #{@consent_form.parent_full_name} marked as invalid"
+                  }
+    else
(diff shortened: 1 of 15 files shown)


tvararu (approved):

paulrobertlloyd:
- Can we have a red button (given this action is currently, for all intents and purposes, destructive)
- Need to confirm the content shown on the confirm page, perhaps:

> This unmatched response will be hidden. This operation cannot be undone.

thomasleese:
> * Can we have a red button (given this action is currently, for all intents and purposes, destructive)

I assume this button should be the red when invalidating consents that have already been automatically matched as well?

paulrobertlloyd:
> I assume this button should be the red when invalidating consents that have already been automatically matched as well?

Yeah, I think so. Missed that. If you can’t undo something, and we are asking for confirmation, we should likely use the warning button.

paulrobertlloyd:
Emma’s suggested content:

> The unmatched response will be hidden. This operation cannot be undone.


(Replace first ’This’ with ‘The’)

PR #2850, opened 2025-01-15
Remove mailers

thomasleese:
This removes various references to `ActionMailer` and the mailer classes since they are no longer being used as we're sending emails using the GOV.UK Notify API directly now (#2841).

We can't remove the `ApplicationMailer` and stop requiring `action_mailer/railtie` until https://github.com/heartcombo/devise/pull/5731 has been released.


tvararu (approved):
:scissors:

PR #2871, opened 2025-01-21
Add SystmOne exporter

misaka:
The exporter has been tested and confirmed working by Hertfordshire. It's hidden behind a feature flag for the time being, and only available on HPV programmes.

image

Worth noting the current export format is fairly bespoke to Hertfordshire's needs, and there are a few areas we'll need to review for new teams or even new vaccines:

-  **vaccination field** - This maps from the vaccine brand and dose sequence to a code. We only have mappings for "Gardasil 9" at the moment, and we don't even know if maybe these codes change depending on provider.
- **practice code** - The URN or ODS code of where the vaccination was performed. Hertfordshire has only defined how to do this for vaccinations in schools (URN) or in GP practices (ODS code), we've just assumed that for clinics this is also the ODS code, but this may be wrong.
- **site** - The delivery site of the vaccination. These value may be bespoke to Hertfordshire, we don't know if other SAIS teams will need these to be customised.

### DIFF
--- a/app/lib/reports/systm_one_exporter.rb
+++ b/app/lib/reports/systm_one_exporter.rb
@@ -125,10 +125,10 @@ class Reports::SystmOneExporter
       patient.given_name, # Forename
       gender_code(patient.gender_code), # Gender
       patient.date_of_birth.to_fs(:uk_short),
-      patient.address_line_2, # House name
-      patient.address_line_1, # House number and road
-      patient.address_town, # Town
-      patient.address_postcode, # Postcode
+      patient.restricted? ? "" : patient.address_line_2, # House name
+      patient.restricted? ? "" : patient.address_line_1, # House number and road
+      patient.restricted? ? "" : patient.address_town, # Town
+      patient.restricted? ? "" : patient.address_postcode, # Postcode
       vaccination(vaccination_record), # Vaccination
       "", # Part
       vaccination_record.performed_at.to_date.to_fs(:uk_short), # Admin date
--- a/app/lib/reports/systm_one_exporter.rb
+++ b/app/lib/reports/systm_one_exporter.rb
@@ -9,6 +9,15 @@ class Reports::SystmOneExporter
     }
   }.freeze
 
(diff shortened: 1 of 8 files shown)


thomasleese (approved):

tvararu (approved):

app/lib/reports/systm_one_exporter.rb
```diff
@@ -0,0 +1,187 @@
...
+      patient.nhs_number, # NHS number
+      patient.family_name, # Surname
+      "", # Middle name (not stored)
+      patient.given_name, # Forename
+      gender_code(patient.gender_code), # Gender
+      patient.date_of_birth.to_fs(:uk_short),
+      patient.address_line_2, # House name
+      patient.address_line_1, # House number and road
+      patient.address_town, # Town
+      patient.address_postcode, # Postcode
```
benilovj, line 131:
```suggestion
      patient.date_of_birth.to_fs(:uk_short),
      patient.restricted? ? "" : patient.address_line_2, # House name
      patient.restricted? ? "" : patient.address_line_1, # House number and road
      patient.restricted? ? "" : patient.address_town, # Town
      patient.restricted? ? "" : patient.address_postcode, # Postcode
```

benilovj (changes requested):
I think we need to mirror the restricted patient functionality we have for the other reports here as well.

app/lib/reports/systm_one_exporter.rb
misaka: Good catch! Do we need to double-check that those are the only fields that need to be removed?

misaka: Changes done, btw.

benilovj: > Do we need to double-check that those are the only fields that need to be removed?

The rule is that contact details for restricted patients shouldn't be used, which I think is now satisfied.

benilovj (approved):

PR #2903, opened 2025-01-28
Improve session slug string sequence generation

thomasleese:
This improves the session factory to generate session slugs with alphanumeric characters to make it clear that these are slugs and not integer IDs, avoiding any potential confusiong and making them more realistic.


tvararu (approved):
✨

PR #2921, opened 2025-02-01
Handle school move log entries when merging patients 

thomasleese:
This reverts the change that was made in https://github.com/nhsuk/manage-vaccinations-in-schools/commit/6d8880d3c5b4810b7851614fdf30ce365a163f6e to remove these foreign keys to fix a bug where patients couldn't be merged due to the school move log entries existing.

Instead, we can handle the school move log entries by modifying the `patient_id` value, ensuring that the foreign keys continue to be valid. Removing the foreign keys means we can no longer rely on this model to generate activity logs which is one of the reason it was added.

It's true that it also prevents us from being able to delete any schools or users, but we already have this problem or more crucial models (like vaccination records, triage, sessions, etc) so when if we come to a situation where we need to handle that we should do it in a consistent way (likely by soft-deleting the foreign keyed model).

### DIFF
--- a/app/lib/patient_merger.rb
+++ b/app/lib/patient_merger.rb
@@ -31,6 +31,10 @@ class PatientMerger
         end
       end
 
+      patient_to_destroy.school_move_log_entries.update_all(
+        patient_id: patient_to_keep.id
+      )
+
       patient_to_destroy.session_notifications.update_all(
         patient_id: patient_to_keep.id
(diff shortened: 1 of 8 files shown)


misaka (approved):
Ya, I did consider this and see your point. I didn't do it largely because I wasn't sure I'd be able to do it in time. 👍

PR #2922, opened 2025-02-02
Increase wait time between PDS jobs

thomasleese:
We're seeing a problem in production where the queues for PDS jobs gets full with jobs that are not succeeding, likely related to the rate limit.

Some potentially related issues:

- https://github.com/bensheldon/good_job/issues/838
- https://github.com/bensheldon/good_job/issues/1278
- https://github.com/bensheldon/good_job/issues/1353

This wait time was added to artifically slow down the queues and prevent them from needing to be retried, so increasing this wait time should reduce the changes of the jobs ending up in this stuck state.

This is a temporary measure until we can either move away from Good Job or resolve the underlyling issue which is still not 100% clear.

### DIFF
--- a/app/jobs/bulk_update_patients_from_pds_job.rb
+++ b/app/jobs/bulk_update_patients_from_pds_job.rb
@@ -25,7 +25,7 @@ class BulkUpdatePatientsFromPDSJob < ApplicationJob
 
           PatientUpdateFromPDSJob.set(
             priority: 50,
-            wait: 0.21 * index
+            wait: 0.5 * index
           ).perform_later(patient)
         end
     end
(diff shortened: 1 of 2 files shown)


app/jobs/bulk_update_patients_from_pds_job.rb
```diff
@@ -25,7 +25,7 @@ def perform
 
           PatientUpdateFromPDSJob.set(
             priority: 50,
-            wait: 0.21 * index
```
misaka, line 28: I'm starting to wonder if maybe we want this to be configurable outside of the source code. Not sure if we have a good way to do it, though, am I right in thinking that Settings don't get applied unless we deploy a new image or force a redeploy?

thomasleese: Yeah that's probably not a bad idea. I think if we had it as an environment variable we'd only need to restart the container I think to get the new change in.

tvararu (approved):

PR #2940, opened 2025-02-06
Use test notify templates for consent requests

mikejamesthompson:
**Do not merge**

This is a temporary change to use test Notify templates for consent requests that include a patient's first name to facilitate testing of the templates before publication of the change.

Contributes towards resolving MAVIS-1783.


thomasleese:
The testers have confirmed this is working as expected.

PR #2961, opened 2025-02-12
Add Td/IPV and MenACWY programmes

thomasleese:
This adds the two new programme types needed for doubles and updates the various parts of the code that rely on the programme type to work for Td/IPV and MenACWY.

In a future commit we'll need to add the vaccines for these two programmes, and then the health questions.


tvararu (approved):
Nice, I'll rebase https://github.com/nhsuk/manage-vaccinations-in-schools/pull/2959

PR #2965, opened 2025-02-12
Filter vaccination records by programme

thomasleese:
This follows up on #2939 to ensure that vaccination records are filtered by the current programme. At the moment this is the first programme administered in a session but in the future this will come from a selection by the user.

At the same time, this also fixes a bug where vaccination records administered outside of the current session weren't shown on a patient's record. Now we show all vaccination records for a particular patient.

[Trello Card](https://trello.com/c/cfPtR6Am/2000-if-a-child-is-added-to-a-clinic-and-vaccinated-in-school-you-cant-see-in-the-clinic-area-that-they-are-already-vaccinated)

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -44,7 +44,6 @@ class Organisation < ApplicationRecord
   has_many :programmes, through: :organisation_programmes
   has_many :schools, through: :teams
   has_many :vaccines, through: :programmes
-  has_many :vaccination_records, through: :patient_sessions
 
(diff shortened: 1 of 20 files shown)


app/models/organisation.rb
```diff
@@ -37,14 +37,14 @@ class Organisation < ApplicationRecord
   has_many :patients
   has_many :sessions
   has_many :teams
+  has_many :vaccination_records, -> { kept }
```
misaka, line 40: ?

PR #2967, opened 2025-02-13
Search globally for patients when merging

thomasleese:
This prevents a bug where a user cannot be merged with a patient that has been moved out of the cohort as they can't be found by the policy scope.

To support this, we've slightly updated the logic around the patient merging to move the patient back in to the cohort if they are not currently part of any cohort as this is likely to be the expected behaviour.

There is an outstanding question about what happens if we're merging two patients from two different organisation's cohorts, it's unclear which organisation's cohort the merged patient should belong to.

[Trello Card](https://trello.com/c/LIFt9GVO/1821-for-some-patients-when-you-try-to-add-an-nhs-number-to-a-duplicate-record-so-that-you-can-merge-them-you-get-an-error-saying-nhs)

### DIFF
--- a/app/controllers/patients/edit_controller.rb
+++ b/app/controllers/patients/edit_controller.rb
@@ -12,14 +12,7 @@ class Patients::EditController < ApplicationController
 
     redirect_to edit_patient_path(@patient) and return unless @patient.changed?
 
-    @existing_patient =
-      if nhs_number.present?
-        policy_scope(Patient).includes(parent_relationships: :parent).find_by(
-          nhs_number:
-        )
-      end
-
-    render :nhs_number_merge and return if @existing_patient
+    render :nhs_number_merge and return if existing_patient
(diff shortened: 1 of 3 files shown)


app/controllers/patients/edit_controller.rb
```diff
@@ -14,9 +14,7 @@ def update_nhs_number
 
     @existing_patient =
       if nhs_number.present?
-        policy_scope(Patient).includes(parent_relationships: :parent).find_by(
-          nhs_number:
-        )
+        Patient.includes(parent_relationships: :parent).find_by(nhs_number:)
```
misaka, line 17: Would it be worth limiting the scope of patients here to be either those connected to the user's organisation OR with no organisation, versus what we have here which could move patients from other organisations into the user's?

misaka (approved):

PR #2970, opened 2025-02-13
Terraform migration

TheOneFromNorway:
This commit includes a working setup for deployments via terraform. Furthermore, it adds helper scripts for the copilot -> terraform migration.


bogsi17 (approved):

terraform/.gitignore
```diff
@@ -0,0 +1,6 @@
+.idea/
+.terraform/
+.terraform.lock.hcl
+venv/
+**/*.tfstate
+**/*.tfstate.backup
```
thomasleese, line 6: I reckon we can put these in the `.gitignore` in the root of the project, I think it keeps things simple if there's only one `.gitignore`.

terraform/app/modules/firewall/main.tf
```diff
@@ -0,0 +1,123 @@
...
+resource "aws_networkfirewall_firewall" "this" {
+  name                = "firewall-${var.environment_string}"
+  firewall_policy_arn = aws_networkfirewall_firewall_policy.this.arn
+  vpc_id              = var.vpc_id
+  subnet_mapping {
+    subnet_id = aws_subnet.this.id
+  }
+}
+
+############## LOGGING ################
```
thomasleese, line 95: Rather than these comments could we structure this file so the logging and firewall rules go in their own file?

terraform/app/variables.tf
```diff
@@ -0,0 +1,201 @@
...
+
+variable "container_name" {
+  type        = string
+  default     = "mavis"
+  description = "Name of essential container in the task definition."
+}
+
+variable "docker_image" {
+  type        = string
+  default     = "<CHANGE_ME>"
```
thomasleese, line 106: Could this default to `null` rather than this specific string?

terraform/app/variables.tf
thomasleese, line 1: I'm wondering if we need to set `nullable = false` to a number of these variables?

https://developer.hashicorp.com/terraform/language/values/variables#disallowing-null-input-values

terraform/app/variables.tf
```diff
@@ -0,0 +1,201 @@
+variable "region" {
+  type        = string
+  default     = "eu-west-2"
+  description = "AWS region"
+}
+
+variable "environment_string" {
```
thomasleese, line 7:
```suggestion
variable "environment" {
```

Do we need `string` in the variable name, as we're not doing this elsewhere it looks a bit odd. If `environment` isn't clear enough we could go with `app_environment` or something like that?

terraform/app/variables.tf
```diff
@@ -0,0 +1,201 @@
...
+variable "image_tag" {
+  type        = string
+  description = "The docker image tag for the essential container in the task definition."
+}
+
+locals {
+  container_name = "${var.container_name}-${var.environment_string}"
+  docker_image   = var.docker_image == "<CHANGE_ME>" ? "mavis-${var.environment_string}" : var.docker_image
+  is_production  = var.environment_string == "production"
+  dev_task_envs = [
```
thomasleese, line 119: Do we need to be able to deploy a `development` version of the app? I'm wondering if it would be safer to have some validation that makes sure the `environment_string` _isn't_ set to `development`?

terraform/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+The POC demonstrates a multi-backend configuration. Just run
+
+```bash
+terraform init -backend-config=env/<config-file>
+```
+
+in the `app` directory to select which environment/state-file you want.
+
+### Pre-commit
```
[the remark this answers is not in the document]
bogsi17: No, I will remove it

.gitignore
```diff
@@ -4,6 +4,7 @@
 .env
 .env.local
 .envrc
+venv
```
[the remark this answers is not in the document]
bogsi17: removing it

terraform/app/modules/firewall/main.tf
TheOneFromNorway: I think for a module it is nice if the file is not too long that everything is in one location. I would in general not start splitting terraform configuration into multiple files unless the length became an issue.

terraform/.gitignore
bogsi17: Isn't it simpler to have a separate `.gitignore` here, specifically for the files in this folder?

thomasleese: I think it can then be surprising, for example if a file is ignored and you're not sure why, you have to check all the `.gitignore` files to figure out why.

PR #2979, opened 2025-02-17
Remove line about header row not counting

thomasleese:
We've re-numbered the errors so it's clearer now which row they relate to, so we no longer need this content.


mikejamesthompson (approved):
👍

PR #2996, opened 2025-02-18
Add ability to assign programmes to sessions 

thomasleese:
This makes it possible to edit the programmes that a session administers after the session dates have been scheduled, as long as all the dates are in the past and the session is still open.

## Screenshots

Screenshot 2025-02-18 at 13 44 47
Screenshot 2025-02-18 at 13 44 41
Screenshot 2025-02-18 at 13 44 26
Screenshot 2025-02-18 at 13 31 26
Screenshot 2025-02-18 at 13 31 20

### DIFF
--- a/app/controllers/sessions/edit_controller.rb
+++ b/app/controllers/sessions/edit_controller.rb
@@ -3,6 +3,18 @@
 class Sessions::EditController < ApplicationController
   before_action :set_session
 
+  def edit_programmes
+    render :programmes
+  end
+
+  def update_programmes
+    if @session.can_change_programmes? && @session.update(programmes_params)
+      redirect_to edit_session_path(@session)
+    else
+      render :programmes, status: :unprocessable_entity
+    end
+  end
+
   def edit_send_consent_requests_at
     render :send_consent_requests_at
   end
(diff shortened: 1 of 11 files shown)


app/controllers/sessions/edit_controller.rb
```diff
@@ -3,6 +3,18 @@
 class Sessions::EditController < ApplicationController
   before_action :set_session
 
+  def edit_programmes
+    render :programmes
+  end
+
+  def update_programmes
+    if @session.update(programmes_params)
```
misaka, line 11: Should we be checking `@session.can_change_programmes?` here too?

app/models/session.rb
```diff
@@ -122,8 +122,7 @@ class Session < ApplicationRecord
                 location.generic_clinic?
             end
 
-  validates :programmes, presence: true
-  validate :programmes_part_of_organisation
```
misaka, line 126: Do we not still want to validate that the programmes are a part of the organisation?

thomasleese: The organisation programmes are current, but I can imagine a scenario where an organisation used to administer a programme but doesn't anymore, and we wouldn't want all the historical sessions to be invalid.

app/controllers/sessions/edit_controller.rb
thomasleese: Yeah good point, will amend.

thomasleese: This has been fixed now.

misaka (approved):

PR #2997, opened 2025-02-18
MAV-384: Modify application configuration

TheOneFromNorway:
* The credentials for autogenerated secrets by aurora RDS is different
** To not be reliant on copilot in the future we should be able to handle terraform native setups

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -15,6 +15,7 @@ require "action_text/engine"
 require "action_view/railtie"
 require "action_cable/engine"
 require "rails/test_unit/railtie"
+require "cgi"
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -42,6 +43,18 @@ class Application < Rails::Application
       ENV[
         "DATABASE_URL"
       ] = "postgres://#{username}:#{password}@#{host}:#{port}/#{dbname}"
+      # for environment which uses RDS aurora managed credentials only the the username
+      # and password is automaticallyset. The environment variable is then DB_CREDENTIALS
```
misaka, line 47: Would it make sense to have these comments after the `elsif`? I can see they are related to that block because it's obvious in this diff, but if I were looking at this outside the diff I'm not so sure it would be as clear.

config/application.rb
```diff
@@ -42,6 +43,18 @@ class Application < Rails::Application
...
       ] = "postgres://#{username}:#{password}@#{host}:#{port}/#{dbname}"
+    elsif ENV["DB_CREDENTIALS"].present?
+      # for environment which uses RDS aurora managed credentials only the the username
+      # and password is automaticallyset. The environment variable is then DB_CREDENTIALS
+      db_config = JSON.parse(ENV["DB_CREDENTIALS"])
+      username = CGI.escape(db_config["username"])
+      password = CGI.escape(db_config["password"])
+      host = CGI.escape(ENV["DB_HOST"])
+      dbname = CGI.escape(db_config["DB_NAME"])
+      port = 5432
```
thomasleese, line 54: Would this ever come from an environment variable? Perhaps something like `ENV.fetch("DB_PORT", 5432)` might be good?

config/application.rb
```diff
@@ -42,6 +43,18 @@ class Application < Rails::Application
       ENV[
         "DATABASE_URL"
       ] = "postgres://#{username}:#{password}@#{host}:#{port}/#{dbname}"
+    elsif ENV["DB_CREDENTIALS"].present?
+      # for environment which uses RDS aurora managed credentials only the the username
+      # and password is automaticallyset. The environment variable is then DB_CREDENTIALS
```
thomasleese, line 48:
```suggestion
      # and password is automatically set. The environment variable is then DB_CREDENTIALS
```

thomasleese (commented):
I think we can probably squash together the commit messages here, and perhaps rename the title of the commit to briefly explain what configuration is being modified?

thomasleese (approved):

PR #3001, opened 2025-02-19
Use pre-commit to run linters

bogsi17:
- Ensure we are formatting properly and consistently
  - Users can in principle choose to opt out of automatic hooks by not
    installing the tool in the repo
- Simplicity of including linting into CI
  - For stability we run them initially in parallel to existing linting
    scripts
- Scalability and further development is simple as we can use existing
  solutions instead of implementing our own linting setup


README.md
```diff
@@ -329,6 +349,13 @@ Then to install the required tools (or update, following a change to
 
 ```shell
 mise install
+mise exec python -- pip install pre-commit
```
thomasleese, line 352: Is this step necessary? In the section above it mentions installing `pre-commit` via `brew` or via `apt`.

README.md
```diff
@@ -329,6 +349,13 @@ Then to install the required tools (or update, following a change to
...
+```
+
+Next ensure you have the pre-commit hook installed by running
```
thomasleese, line 355: I think we should add something here about this step being optional, I don't think we should imply that it's required to install the hooks to set up the project.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,43 @@
...
+        name: Rubocop
+        entry: bin/bundle exec rubocop --autocorrect-all
+        language: system
+        files: ^app/
+        types: [ruby]
+      - id: rufo
+        name: Rufo
+        entry: bin/bundle exec rufo app
+        language: system
+        pass_filenames: false
```
thomasleese, line 36: I'm wondering about this configuration for Rufo. I think the preference would be to follow our existing rules, where only `.erb` files are passed to `rufo`. Is there a way to filter the list similar to how `.rufo` works rather than need to lint all the files each time?

I realise that `bin/lint` currently works as you've set up `pre-commit`, but I think this is a nice opportunity to improve upon our existing setup.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - id: brakeman
+        name: Brakeman
+        entry: bin/bundle exec brakeman --no-pager .
+        language: system
+        always_run: true
+        pass_filenames: false
+        stages: [manual]
```
thomasleese, line 43: I think we would still want `brakeman` to run each time like it does currently in `bin/lint`.

.pre-commit-config.yaml
TheOneFromNorway: Ahh I had an improved setup where it looked at .erb files inside app directory (since it made some modifications to those outside) but semes to have been dropped in a refactoring. I will re-introduce it 👍

README.md
TheOneFromNorway: There are 2 pre-commits, IIRC the brew/apt pre-commit is for command line `pre-commit <args>` which allows you to run hooks/etc manually from command line and the python pre-commit works with the .pre-commit-config.yaml configuration

.pre-commit-config.yaml
TheOneFromNorway: It does extend the runtime of the pre-commit hooks to ca. 5 seconds since it acts on all files. Are we sure we want this? We can still choose to run it manually or e.g. inside bin/lint

README.md
```diff
@@ -309,14 +309,34 @@ brew install libyaml
...
 export CPPFLAGS="-I$OPENSSL_PATH/include $CPPFLAGS"
 export PKG_CONFIG_PATH="$CMAKE_PREFIX_PATH/lib/pkgconfig"
+```
+
+For ubuntu run
+
+```shell
+apt install curl libyaml-dev libreadline-dev zlib1g-dev \
+libssl-dev libicu-dev cmake pkg-config uuid-dev flex bison \
+pre-commit
```
misaka, line 329: Does `pre-commit` not list all these other packages as dependencies?

TheOneFromNorway: That is actually just what I had to install to run things without pre-commit. I figured I would make a linux section to make future onboarding easier since the mac section did not really work for me

thomasleese:
Closing this as we've decided to try `hk` instead: https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3412

PR #3007, opened 2025-02-19
Setup data for multiple programmes when seeding the db

mikejamesthompson:
So we can use the automatic Heroku deployments to test the doubles changes.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -296,6 +296,44 @@ def create_school_moves(organisation)
   end
 end
 
+def create_organisation_sessions(user, organisation)
+  # HPV-only sessions
+  create_session(user, organisation, programme_types: ["hpv"], completed: false)
+  create_session(user, organisation, programme_types: ["hpv"], completed: true)
+
+  # MenACWY and Td/IPV combined sessions
+  create_session(
+    user,
+    organisation,
+    programme_types: %w[menacwy td_ipv],
+    completed: false,
+    year_groups: [8, 9, 10]
+  )
+  create_session(
+    user,
+    organisation,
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -91,97 +91,110 @@ def attach_specific_school_to_organisation_if_present(organisation:, urn:)
...
-    vaccine: programme.vaccines.active.first
-  )
+def create_session(
+  user,
+  organisation,
+  programmes:,
+  completed: false,
+  year_groups: nil
+)
+  year_groups ||= programmes.first.year_groups
```
thomasleese, line 112: I think we can get the year groups from all the programmes like above:

```suggestion
  year_groups ||= programmes.flat_map(&:year_groups).uniq
```

tvararu (approved):

db/seeds.rb
```diff
@@ -91,97 +91,108 @@ def attach_specific_school_to_organisation_if_present(organisation:, urn:)
...
-    vaccine: programme.vaccines.active.first
-  )
+def create_session(
+  user,
+  organisation,
+  programmes:,
+  completed: false,
+  year_groups: nil
+)
+  year_groups ||= programmes.flat_map(&:year_groups).uniq
```
tvararu, line 112:
```suggestion
  year_groups = programmes.flat_map(&:year_groups).uniq
```

I don't think the `||=` does anything (this method's scope isn't shared with the other one)

tvararu (commented):
Very minor nit

tvararu:
Merging so I can use this in my own testing 🚀

PR #3020, opened 2025-02-21
Add deployment pipelines

bogsi17:
This PR adds two Github worflows for Terraform deployments

1) Deploy Mavis
This workflow deploys a Mavis docker image. It takes the docker image tag as an input, updates the terraform task definition and triggers CodeDeploy to roll out the new version. Before applying the changes, it verifies that no other terraform resources would be touched.

2) Deploy Infrastructure
This workflow updates the infrastructure, while keeping the docker image constant. It verifies the configuration changes and fails if any resources are destroyed that would lead to a downtime. Changes that would involve destroying one of these resources, as well as deploying a new environment from scratch still have to be triggered from the command line. 

Both workflows wait for manual approval before applying any changes if they target a protected environment.


TheOneFromNorway (approved):
Changes cause pipelines to work as expected

PR #3030, opened 2025-02-21
Move programme selection to filter panel

thomasleese:
When viewing a list of children for a session, the programme selection form currently shows at the top of the page. This commit improves the design of this form to match the designs in the prototype where the filter is shown in the filter panel.

## Screenshot

Screenshot 2025-02-21 at 19 34 34


benilovj (approved):

paulrobertlloyd:
Added CSS and tweaked the filter panel so that the radios are the same size as checkboxes, and legends and labels are consistently styled.

Screenshot of filter panel.

PR #3044, opened 2025-02-24
Combine health questions when giving consent

tvararu:
If there are multiple vaccines, this ensures that we use all the available health questions.

### DIFF
--- a/config/vaccines.yml
+++ b/config/vaccines.yml
@@ -101,7 +101,6 @@ gardasil_9:
 menquadfi:
   brand: MenQuadfi
   type: menacwy
-  discontinued: true
   method: injection
   manufacturer: Sanofi
   nivs_name: MenQuadfi
--- a/config/vaccines.yml
+++ b/config/vaccines.yml
@@ -124,6 +124,7 @@ menveo:
 nimenrix:
   brand: Nimenrix
   type: menacwy
+  discontinued: true
   method: injection
   manufacturer: Pfizer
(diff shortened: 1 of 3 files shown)


spec/models/consent_form_spec.rb
```diff
@@ -634,6 +634,35 @@
     expect(consent_form.health_answers).to be_empty
   end
 
+  it "combines health questions from multiple active vaccines" do
+    programme1 = create(:programme, :menacwy)
+
+    # Menacwy vaccine is discontinued in yml, we need >1 active vaccines to test
+    programme1.vaccines.first.update!(discontinued: false)
```
thomasleese, line 641: I think this points to an issue in the config, the `MenQuadfi` should be active as it's the one the SAIS organisation will be vaccinating with.

thomasleese: Yeah I've just checked, it's wrong.

config/vaccines.yml
```diff
@@ -101,7 +101,6 @@ gardasil_9:
 menquadfi:
   brand: MenQuadfi
   type: menacwy
-  discontinued: true
```
thomasleese, line 104: I think Nimenrix is supposed to be the discontinued one, I think I made a mistake in this commit: https://github.com/nhsuk/manage-vaccinations-in-schools/commit/8651dc9d61dc72d14f57f8717f7377cdd35c6309

thomasleese (approved):

PR #3052, opened 2025-02-24
MAV-332: Ensure log group is deleted by tf destroy

TheOneFromNorway:
* In the old setup the log group gets immediately deleted after the log streams
** Log streams take some time to appear.
** Once log streams appear after log group deletion they re-create the group to house them
** Introducing a time delay between log stream/group deletion ensures all streams are finished producing


terraform/app/monitoring.tf
```diff
@@ -10,6 +10,11 @@ resource "aws_cloudwatch_log_group" "vpc_log_group" {
   retention_in_days = var.vpc_log_retention_days
   skip_destroy      = local.is_production
 }
+resource "time_sleep" "wait_to_delete_flowlogs_group" {
+  destroy_duration = "3m"
+  depends_on = [aws_cloudwatch_log_group.vpc_log_group]
```
thomasleese, line 15: From reading the docs it looks to me like the `depends_on` needs to go in the resource that we want to wait before destroying (and it depends on the `time_sleep` resource).

I think right now the resource would still get destroyed immediately.

TheOneFromNorway: In its current configuration terraform will try to destroy aws_flow_log.vpc_flowlogs and time_sleep.wait_to_delete_flowlogs_group at the same time (sinec they both only depend on aws_cloudwatch_log_group.vpc_log_group) and will wait until these are destroyed before proceeding to destroy aws_cloudwatch_log_group.vpc_log_group.
Since time_sleep.wait_to_delete_flowlogs_group takes 3 minutes to destroy and aws_flow_log.vpc_flowlogs is destroyed very quickly the existing configuration achieves the intended effect. This has also been verified in testing. 
To future proof against a case where terraform may not try to destroy aws_flow_log.vpc_flowlogs and time_sleep.wait_to_delete_flowlogs_group at the same time we can add a dependency in aws_flow_log.vpc_flowlogs on time_sleep.wait_to_delete_flowlogs_group. This would then cause the sequence <delete aws_flow_log.vpc_flowlogs> -> <delete time_sleep.wait_to_delete_flowlogs_group (takes 3 minutes)> -> <delete aws_cloudwatch_log_group.vpc_log_group>. I will implement the improvement.

thomasleese (approved):

PR #3085, opened 2025-02-26
Bootstrap cleanup

bogsi17:
Remove code that is no longer needed as part of the terraform bootstrap setup.

* Remove ECR repository from terraform config. Every environment pulls images from mavis/webapp now
* Remove creation of docker build scripts from bootstrap script. The docker build is done by the build.yml Github workflow now and the images don't need to be pushed to container registries with different names anymore


TheOneFromNorway (approved):
Tested new script, looks good.

PR #3097, opened 2025-02-27
Stop using NHS services in copilotmigration

TheOneFromNorway:
- Using up non-prod quota
- The environment should not have these features enabled by default


terraform/app/env/copilotmigration.tfvars
```diff
@@ -17,6 +17,6 @@ rails_master_key_path = "/copilot/mavis/copilotmigration/secrets/RAILS_MASTER_KE
 
 rails_env = "staging"
 
-splunk_enabled = "true"
-cis2_enabled   = "true"
-pds_enabled    = "true"
+splunk_enabled = "false"
```
thomasleese, line 20: I think we can keep Splunk, there's no quota on that as far as I know and it'll be useful to have the logs in there.

thomasleese (approved):

terraform/app/env/copilotmigration.tfvars
TheOneFromNorway: OK fixed and squashed

PR #3099, opened 2025-02-27
MAV-745: Deploy infrastructure on push to main

TheOneFromNorway:
- Call infrastructure deploy workflow
- Also remove explicit github.sha inputs
-- From documentation: `When a reusable workflow is triggered by a caller workflow, the github context is always associated with the caller workflow.`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,8 +44,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-        with:
-          ref: ${{ inputs.github_sha || github.sha }}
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -29,8 +25,6 @@ jobs:
       RAILS_ENV: test
     steps:
       - uses: actions/checkout@v4
-        with:
-          ref: ${{ inputs.github_sha || github.sha }}
```
bogsi17, line 33: These two lines can be removed from the `jest` and `playwright` jobs further down in the workflow too

bogsi17 (approved):

PR #3100, opened 2025-02-27
Temporarily change consent given template

thomasleese:
We'd like to change the content in 2.0.0 so we need to point to a template with the new content.


thomasleese:
This has been tested and is passing.

PR #3109, opened 2025-02-27
MAV-745: Give artifacts unique names

TheOneFromNorway:
- Artifacts with same name cannot exist in one workflow 
  - Therefore the continuous_deployment workflow failed

### DIFF
--- a/.github/workflows/deploy_mavis.yml
+++ b/.github/workflows/deploy_mavis.yml
@@ -76,7 +76,7 @@ jobs:
       - name: Upload artifact
         uses: actions/upload-artifact@v4
         with:
-          name: tfplan_mavis
+          name: tfplan_app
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy_mavis.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       - name: Upload artifact
         uses: actions/upload-artifact@v4
         with:
-          name: tfplan
+          name: tfplan_mavis
```
thomasleese, line 79:
```suggestion
          name: tfplan_app
```

Could we call one task deploying the infrastructure and one task deploying the app? I think that'll help to clarify what these jobs do as Mavis, in my mind at least, is the whole service, including the infrastructure and the app.

thomasleese (approved):

PR #3131, opened 2025-03-04
Add Terraform backup module

bogsi17:
This PR enables automated resource backups using the AWS Backup service. To achieve that, the terraform-aws-backup module provided by NHSDigital (https://github.com/NHSDigital/terraform-aws-backup) is used.

The only backed up resource is the Aurora DB. Snapshots will be done twice a day at 7am and 7pm and stored for 60 days.

The module supports also automated copying of the backups into a separate AWS account. However, this will be enabled at a later stage as it's not possible to copy an RDS snapshot that is encrypted with a managed KMS key.


hk.pkl
```diff
@@ -32,7 +32,7 @@ local linters = new Mapping<String, Group> {
 
             ["tf_lint"] = new Step {
                 glob = "*.tf"
-                check = "tflint --chdir=terraform --config=$(pwd)/terraform/.tflint.hcl --recursive"
+                check = "tflint --chdir=terraform/app --config=$(pwd)/terraform/.tflint.hcl --recursive && tflint --chdir=terraform/backup --config=$(pwd)/terraform/.tflint.hcl --recursive && tflint --chdir=terraform/bootstrap --config=$(pwd)/terraform/.tflint.hcl --recursive && tflint --chdir=terraform/modules --config=$(pwd)/terraform/.tflint.hcl --recursive"
```
thomasleese, line 35: Does this mean we're currently not linting `terraform/app`?

.github/workflows/terraform-backup-module.yml
thomasleese, line 1: I think we should try and name the file similarly to the name of the action, `deploy-backup-infrastructure.yml`, for example.

terraform/app/variables.tf
```diff
@@ -229,6 +229,12 @@ variable "backup_retention_period" {
   description = "The number of days to retain backups for the RDS cluster."
 }
 
+variable "enable_backup_to_vault" {
```
thomasleese, line 232: I think we can add `nullable = false` to this.

terraform/backup/destination/env/dev.tfvars
```diff
@@ -0,0 +1,2 @@
+source_account_id          = "393416225559"
+source_account_environment = "dev"
```
thomasleese, line 2: Is it possible to name these `development` and `production`, I'm thinking just to keep it in sync with our normal environments.

thomasleese (commented):
Overall looks good, just got a couple of minor comments.

terraform/modules/s3/variables.tf
```diff
@@ -0,0 +1,18 @@
+variable "bucket_name" {
+  description = "The name of the S3 bucket to create"
+  type        = string
+  nullable    = false
+}
+
+variable "logging_target_bucket_id" {
```
thomasleese, line 7: Do we also want to set `nullable = false` on these?

thomasleese (approved):

PR #3132, opened 2025-03-04
Build new session consent tab

thomasleese:
This updates the design of the session consent page to match the latest designs in the prototype, where a single tab for consent is shown and the nurses are able to filter the patients by consent status and the status is shown per programme.

To build this I've refactored the way we handle patient session consent statuses, with the intention to follow this approach with the triage, register, record and outcome status.

## Screenshots

Screenshot 2025-03-04 at 13 39 28
Screenshot 2025-03-04 at 13 57 25
Screenshot 2025-03-04 at 13 57 13
Screenshot 2025-03-04 at 13 57 03


spec/models/patient_session/consent_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+    end
+
+    context "with both an invalidated and not provided consent" do
+      before do
+        create(:consent, :invalidated, patient:, programme:)
+        create(:consent, :not_provided, patient:, programme:)
+      end
+
+      it { should be(described_class::NONE) }
+    end
```
benilovj, line 40: I might be super paranoid here, but is it worth adding a spec for "refusal (invalidated) + given" here?

thomasleese: Yep I think that's a good idea, I've added that in.

benilovj (approved):
Looks great. The `PatientSession::Consent` refactoring is really nice.

PR #3146, opened 2025-03-05
Mav 787 implement review changes

TheOneFromNorway:
- log retention time
- listener rules
- db isntance naming

### DIFF
--- a/terraform/app/loadbalancer.tf
+++ b/terraform/app/loadbalancer.tf
@@ -189,6 +189,6 @@ module "dns_route53" {
   source             = "./modules/dns"
   dns_name           = aws_lb.app_lb.dns_name
   zone_id            = aws_lb.app_lb.zone_id
-  domain_name        = var.domain_name
-  domain_name_prefix = var.environment
+  zone_name          = var.zone_name
+  domain_names       = tolist(toset(values(var.http_hosts)))
 }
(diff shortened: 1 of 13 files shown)


terraform/app/loadbalancer.tf
```diff
@@ -12,7 +12,7 @@ resource "aws_security_group_rule" "lb_ingress_http" {
   from_port         = 80
   to_port           = 80
   protocol          = "tcp"
-  cidr_blocks       = ["0.0.0.0/0"]
+  cidr_blocks = ["0.0.0.0/0"]
```
misaka, line 15: It's just whitespace, but I would've expected this line to stay as-is.

terraform/app/variables.tf
```diff
@@ -20,14 +21,27 @@ variable "account_id" {
...
+  type = object({
+    MAVIS__HOST                        = string
+    MAVIS__GIVE_OR_REFUSE_CONSENT_HOST = string
+  })
+  description = "Https host names"
+  nullable    = true
+}
+
+locals {
+  host_headers = concat(tolist(toset(values(var.http_hosts))), [for v in tolist(toset(values(var.http_hosts))) : "www.${v}"])
```
bogsi17, line 44: Is the second "tolist" really needed or could terraform iterate over the set?
Could you also extract `toset(values(var.http_hosts)` to a new locale, e.g. "unique_host_names"?

bogsi17 (approved):

PR #3147, opened 2025-03-05
Sort patients when fuzzy searching

thomasleese:
When searching for patients by name, we can return the results in order of similarity to the original query to help the nurses find the right patient when looking for patients.

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -131,6 +131,11 @@ class Patient < ApplicationRecord
                 "similarity(given_name, :query) > 0.3 OR " \
                 "similarity(family_name, :query) > 0.3",
               query:
+            ).order(
+              Arel.sql(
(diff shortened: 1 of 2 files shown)


benilovj (approved):

benilovj:
@thomasleese when searching by name through the global children search, I think the similarity ordering gets overridden by ordering by name in the controller:

https://github.com/nhsuk/manage-vaccinations-in-schools/blob/v2.1.0-wip/app/controllers/patients_controller.rb#L14

I suspect that's not the intent?

thomasleese:
> @thomasleese when searching by name through the global children search, I think the similarity ordering gets overridden by ordering by name in the controller:
> 
> I suspect that's not the intent?

My understanding is that the ordering is additive, meaning it'll sort by similarity first and then alphabetically if the similarity value is the same.

https://guides.rubyonrails.org/active_record_querying.html#ordering

PR #3148, opened 2025-03-05
Don't record a dose sequence for Td/IPV

thomasleese:
Td/IPV (or other multi-dose vaccines) shouldn't have a dose sequence number recorded on them by default as we can't be sure which dose sequence will be being administered on the day.


benilovj:
@thomasleese does the guidance need to change when uploading vaccination records, to say that dose sequence isn't required for doubles?

image

thomasleese:
> @thomasleese does the guidance need to change when uploading vaccination records, to say that dose sequence isn't required for doubles?

Good point, actually that column is completely optional for all programmes now.

misaka (approved):

PR #3153, opened 2025-03-06
Refactor status outcomes

thomasleese:
This makes a number of changes to the outcome classes (`Consent`, `Triage`, `Register`, `Record` and `Outcome`) to improve their usability, code quality and make some of them accessible without the need for a `PatientSession`. The behaviour of the service should be the same before and after this PR, but the changes are necessary to support upcoming work.

I've tried to keep the individual refactors in their own commits so reviewing commit by commit probably makes the most sense.

### `Outcome` suffix

The classes have been renamed to include a `Outcome` suffix (for example `Consent` has changed to `ConsentOutcome`). This is to avoid clashes with the model class naming. Specifically, the existence of a `Patient::Consent` model caused the `has_many :consents` relationship to break.

### Standardise interface

The interface of the classes has been standardised such that they all have a `status` and `latest` method, and most also have an `all` method. All of these methods return a hash mapping programmes to the relevant status, latest records or all records. This has been done so that the `Outcome` classes can be inserted in different parts of the code and the caller doesn't need to know what kind of outcome class it is. This also helps to make the outcome classes that come off the `Patient` and the ones that come off the `PatientSession` have the same interface. A future enhancement could be to introduce a superclass or concern that handles common logic.

### Predicate methods

Convenience methods have been added to the outcome classes, for example `given?(programme)` on `ConsentOutcome` as an alternative to `status[programme] == ConsentOutcome::GIVEN`. This is simply to improve the usability and readability of the code.

### Move consent, triage and programme outcome to patient-level

Three of the outcome classes (`TriageOutcome`, `ConsentOutcome` and `ProgrammeOutcome`) have been refactored such that they don't need a `PatientSession` to get an outcome. This is necessary to build the programme's children tab which needs to fetch the programme outcome for a patient regardless of the session. This refactor didn't require any changes to the data model, as consent, triage and vaccination records are only attached to the patient, not the patient session.


config/locales/patient_session.en.yml
benilovj, line 1: Not sure whether this change should be in a separate commit? Or in the very least, mentioned in the commit message?

benilovj (approved):
Lots and lots here but the end result seems cleaner and more readable so 👍

config/locales/patient_session.en.yml
benilovj: oh never mind, i see now.

PR #3166, opened 2025-03-07
Add audit associations

misaka:
This updates the auditing setup on our models. By associating models for audits, it makes it easier for us to get a more complete picture of the changes done to a record with the `own_and_associated_audits`. Here's a minor example of a patient that has been removed from a cohort:

image

The general approach used to decide how and what auditing to add:

- Typically auditing is for tracking updates to records or, in rarer cases, deletion of records.
  - Only enable auditing for those records where we want to track those activities.
  - Join tables don't typically need auditing.
- Models can be associated with one of their `belongs_to` relationships, with a view on usefulness in operational settings, where it's most useful to see associated changes easily.

### DIFF
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -25,7 +25,8 @@
 #  fk_rails_...  (organisation_id => organisations.id)
 #
 class Session < ApplicationRecord
-  audited
+  audited associated_with: :location
+  has_associated_audits
 
   belongs_to :organisation
   belongs_to :location
--- a/app/models/batch.rb
+++ b/app/models/batch.rb
@@ -27,7 +27,7 @@
 class Batch < ApplicationRecord
   include Archivable
 
-  audited
+  audited associated_with: :vaccine
 
   belongs_to :organisation
   belongs_to :vaccine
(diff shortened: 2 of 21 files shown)


app/models/session.rb
```diff
@@ -25,7 +25,8 @@
 #  fk_rails_...  (organisation_id => organisations.id)
 #
 class Session < ApplicationRecord
-  audited
+  audited associated_with: :location
```
thomasleese, line 28: I wonder if associating these with the organisation would be more useful, what do you think?

misaka: Ya, could be. The reason I chose `location` in the end was mainly because I thought how locations can have multiple sessions, so trying to understand what's happened to a location and it's sessions could be useful from an ops pov.  There's also the association link that goes `organisation` <- `team` <- `location` <- `session`, which while long at least is there. We could conceivably write something that pulls up all the associated audits pretty easily since there won't be any loops.

thomasleese: Yeah that's fair enough. I think one tricky situation could be if a school changes management to a different SAIS organisation, but whether that would ever happen I'm not sure.

thomasleese (approved):

app/models/session.rb
misaka: Oof. That's an interesting one. It feels like that might be a bit in the future, happy to revisit/revise this as we go.

PR #3167, opened 2025-03-07
Build new pre-screening designs

thomasleese:
This makes a number of changes to the pre-screening form to match the latest designs in the prototype. Specifically this changes the radio buttons to checkboxes, adds two new questions asking about medication and pregnancy, and pre-populates the answers when vaccinating a patient for multiple vaccines in the same session.

## Screenshot

Screenshot 2025-03-07 at 16 43 25

Screenshot 2025-03-07 at 16 56 38

### DIFF
--- a/app/components/app_activity_log_component.rb
+++ b/app/components/app_activity_log_component.rb
@@ -174,7 +174,8 @@ class AppActivityLogComponent < ViewComponent::Base
         title: "Completed pre-screening checks",
         body: pre_screening.notes,
         at: pre_screening.created_at,
-        by: pre_screening.performed_by
+        by: pre_screening.performed_by,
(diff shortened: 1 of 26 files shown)


benilovj (approved):
LGTM

Not a blocker, but I wonder whether it's worth adding some spec coverage for the pre-population feature on the second vaccine, otherwise might be easy to regress it.

tvararu (approved):

thomasleese:
> Not a blocker, but I wonder whether it's worth adding some spec coverage for the pre-population feature on the second vaccine, otherwise might be easy to regress it.

I've added a feature test that covers this, and a couple of other scenarios around doubles.

PR #3215, opened 2025-03-12
MAV-818: Make listener rule idempotent

TheOneFromNorway:
- Re-creation of listener rule will cause problems if blue-green deployment is on green cycle
  - Terraform will not know that ECS service is not wired to blue target group
  - Terraform will then switch listener to blue but ECS still connected to green
- To fix the state control-missmatch we need to ensure listener can get up to date information on ECS' target group
- Because this feature is not supported by AWS provider we need to use the AWS-CLI directly. Specifically there are 2 issues
  - The terraform aws_ecs_service "data" object does not have a "loadbalancer" field
  - We need to handle the data object not existing gracefullly
    - Default AWS behaviour is to throw error if data object cannot be fetched
      - A feature request for the providerto override this behaviour has been submitted:	issue 41788

### DIFF
--- a/terraform/app/loadbalancer.tf
+++ b/terraform/app/loadbalancer.tf
@@ -139,17 +139,24 @@ resource "aws_lb_listener" "app_listener_https" {
 }
 
 resource "aws_lb_listener_certificate" "https_sni_certificates" {
-  count = length(local.additional_sni_certificates)
-  listener_arn = aws_lb_listener.app_listener_https.arn
+  count           = length(local.additional_sni_certificates)
+  listener_arn    = aws_lb_listener.app_listener_https.arn
   certificate_arn = local.additional_sni_certificates[count.index]
+}
+
+data "external" "active_target_group" {
+  program = ["bash", "-c", "aws ecs describe-services --cluster ${aws_ecs_cluster.cluster.arn} --service ${local.ecs_service_name} | jq -r '.services[0].loadBalancers[0] | {targetGroupArn}'"]
 }
 
 resource "aws_lb_listener_rule" "forward_to_app" {
   listener_arn = aws_lb_listener.app_listener_https.arn
   priority     = 50000
-  action {
-    type             = "forward"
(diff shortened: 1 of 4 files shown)


terraform/app/loadbalancer.tf
```diff
@@ -139,17 +139,24 @@ resource "aws_lb_listener" "app_listener_https" {
...
+}
+
 resource "aws_lb_listener_rule" "forward_to_app" {
   listener_arn = aws_lb_listener.app_listener_https.arn
   priority     = 50000
-  action {
-    type             = "forward"
-    target_group_arn = aws_lb_target_group.blue.arn
+  dynamic "action" {
+    for_each = [data.external.active_target_group.result["targetGroupArn"]]
```
bogsi17, line 155: Does it really need to be a dynamic action? There should always be only one active target group

terraform/app/loadbalancer.tf
```diff
@@ -139,17 +139,24 @@ resource "aws_lb_listener" "app_listener_https" {
...
 resource "aws_lb_listener_certificate" "https_sni_certificates" {
-  count = length(local.additional_sni_certificates)
-  listener_arn = aws_lb_listener.app_listener_https.arn
+  count           = length(local.additional_sni_certificates)
+  listener_arn    = aws_lb_listener.app_listener_https.arn
   certificate_arn = local.additional_sni_certificates[count.index]
 }
 
+data "external" "active_target_group" {
+  program = ["bash", "-c", "aws ecs describe-services --cluster ${aws_ecs_cluster.cluster.arn} --service ${local.ecs_service_name} | jq -r '.services[0].loadBalancers[0] | {targetGroupArn}'"]
```
bogsi17, line 148:
```suggestion
  program = ["bash", "-c", "aws ecs describe-services --cluster ${aws_ecs_cluster.cluster.arn} --service ${local.ecs_service_name} | jq -r '.services[0].loadBalancers[0]'"]
```

PR #3224, opened 2025-03-12
Small improvements to outstanding programmes banner

thomasleese:
This makes a number of small improvements to the banner listing outstanding programmes to ensure that only programmes without an outcome are shown and only to nurse users.

### DIFF
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -140,11 +140,13 @@ class PatientSession < ApplicationRecord
     end
   end
 
-  def programmes_ready_for_vaccinator
+  def outstanding_programmes
     # If this patient hasn't been seen yet by a nurse for any of the programmes,
     # we don't want to show the banner.
     return [] if programmes.all? { session_outcome.none?(it) }
 
-    programmes.select { ready_for_vaccinator?(programme: it) }
+    programmes.select do
+      ready_for_vaccinator?(programme: it) && session_outcome.none?(it)
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/patient_session.rb
```diff
@@ -140,11 +140,13 @@ def ready_for_vaccinator?(programme: nil)
...
 
-  def programmes_ready_for_vaccinator
+  def outstanding_programmes
     # If this patient hasn't been seen yet by a nurse for any of the programmes,
     # we don't want to show the banner.
     return [] if programmes.all? { session_outcome.none?(it) }
 
-    programmes.select { ready_for_vaccinator?(programme: it) }
+    programmes.select do
+      ready_for_vaccinator?(programme: it) && session_outcome.none?(it)
```
misaka, line 149: I find the `none?` here confusing as I'd assumed it was the opposite of `any?`, and `it` was being compared as: `it === session_outcome`. Couldn't see how this works until I realised that `none?` is a custom method.

I'm not sure if you'd meant to override `none?`, if not it might be better to use a different name ... e.g. `none_for_programme?`, even though that'll make some of these lines less concise.

misaka (approved):

app/models/patient_session.rb
misaka: Probably not a fix for this PR.  🥾🥫

misaka: Just looking at the other PR where you add `already_had?`, maybe `none_for?` would be sufficient here.

thomasleese: Yeah, agreed. I've been wondering about this too because Rubocop also gets confused sometimes about the usage of `none?`.

I'll refactor this in a follow up PR.

thomasleese: https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3229

PR #3226, opened 2025-03-12
Add ticks to programme tabs

thomasleese:
When viewing a patient session, if that programme has an vaccinated outcome for this session, we should show a tick next to the programme so the nurses know that there's nothing further to be done for that programme.

## Screenshot

Screenshot 2025-03-12 at 17 31 40


misaka (approved):
🚀

PR #3230, opened 2025-03-13
Show ticks according to programme outcome

thomasleese:
If a patient was vaccinated in a clinic and they are being viewed in the context of a school session, the programme navigation item should be ticked to show that the patient is vaccinated, even if it wasn't in that session.


tvararu (approved):
👨‍🎨

PR #3233, opened 2025-03-13
Enable Splunk logging in qa

misaka:
This is particularly useful to investigate performance-related issues.

### DIFF
--- a/copilot/webapp/manifest.yml
+++ b/copilot/webapp/manifest.yml
@@ -59,7 +59,7 @@ environments:
       rolling: recreate # Disables blue-green deployment for speed
     variables:
       RAILS_ENV: staging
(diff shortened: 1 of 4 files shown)


copilot/webapp/manifest.yml
```diff
@@ -59,7 +59,7 @@ environments:
       rolling: recreate # Disables blue-green deployment for speed
     variables:
       RAILS_ENV: staging
-      SENTRY_ENVIRONMENT: preview
```
thomasleese, line 62: I think we'll need to configure Sentry to use `DEPLOY_ENV` as by default it picks up that env var.

thomasleese:
@misaka I think we've got Splunk enabled now in QA, do we still need this, or maybe we can enable Splunk in all the non-production environments?

PR #3234, opened 2025-03-13
Don't invalidate self-consent if vaccinated

thomasleese:
If the patient was vaccinated after getting self-consent, we shouldn't invalidate that consent as it was valid when the vaccination was given. A future enhancement might be to add a concept of "lifespan" to consents, to capture exactly at which point they are invalid.

At the moment, the nurses see a patient that was vaccinated, with supposedly no valid consent, which is not accurate.

[Jira Issue - MAV-865](https://nhsd-jira.digital.nhs.uk/browse/MAV-865)

### DIFF
--- a/app/jobs/invalidate_self_consents_job.rb
+++ b/app/jobs/invalidate_self_consents_job.rb
@@ -4,29 +4,36 @@ class InvalidateSelfConsentsJob < ApplicationJob
   queue_as :consents
 
   def perform
-    team_ids = Team.pluck(:id)
-    programme_ids = Programme.pluck(:id)
+    academic_year = AcademicYear.current
 
-    team_ids
-      .product(programme_ids)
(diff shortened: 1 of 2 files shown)


benilovj:
Is there any way that looking at an invalidated consent, SAIS staff can more easily see why it was invalidated (in the middle of the night), and that it happened automatically, not by someone going in at 2am?

thomasleese:
> Is there any way that looking at an invalidated consent, SAIS staff can more easily see why it was invalidated (in the middle of the night), and that it happened automatically, not by someone going in at 2am?

What we could do is track the user who invalidated the consent and triage (`invalidated_by_user_id`), but allow it to be `nullable`. Then we can show this in the activity log, so if a user did it we can include that in the entry, and if not we know it happened automatically.

tvararu (approved):

PR #3254, opened 2025-03-18
Create and destroy infrastructure pipelines

TheOneFromNorway:
- With changes to bootstrap no unique create pipeline is needed
      - Will be included in normal deploy infrastructure
- Destroy pipeline deletes both
    - TF managed resources
    - Backend configuration for environment
       - Note S3 Bucket/DynamoDB table are not deleted as they are shared
         among environments in an account


thomasleese (commented):
Functional changes look good, I've made some suggesting so the names are consistent with the changes I made in https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3255

I think it would be good to squash the linting commits in with the relevant commit that added the change originally.

PR #3260, opened 2025-03-19
GraphRecords UI: An ops tool to visualise how objects are related in the database

alistair-white-horne-tng:
An endpoint which serves a UI which can be used to interactively navigate through the database, viewing objects and useful information about them.

This is a tool designed for use by the Ops Support people, to visualise the data structure without having to use the console.

The information displayed for each object type is as follows:

```ruby
consent: %i[response route created_at],
session: %i[clinic?],
session_attendance: %i[attending],
triage: %i[status created_at],
vaccination_record: %i[outcome performed_at],
programme: %i[type],
vaccine: %i[nivs_name],
organisation: %i[name],
location: %i[name],
cohort_import: %i[csv_filename created_at status rows_count],
class_import: %i[csv_filename created_at status rows_count year_groups],
session_date: %i[value],
team: %i[name]
```

image


app/controllers/concerns/authentication_concern.rb
```diff
@@ -63,18 +63,24 @@ def store_user_location!
...
+
+    def perform_basic_auth
+      authenticated =
+        authenticate_with_http_basic do |username, password|
+          username == Rails.application.credentials.support_username &&
+            password == Rails.application.credentials.support_password
+        end
+
+      unless authenticated
+        request_http_basic_authentication "Application", <<~MESSAGE
```
misaka, line 81: iirc we decided we don't need this anymore, is that right?

app/controllers/inspect/graph/graphs_controller.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+module Inspect
+  module Graph
+    class GraphsController < ApplicationController
+      skip_after_action :verify_policy_scoped
+      skip_before_action :authenticate_user!
+      if Rails.env.production?
+        before_action :authenticate_basic_always, only: :show
+      end
```
misaka, line 10: I think we're handling this in the routes file instead now and don't need this.

app/controllers/inspect/graph/graphs_controller.rb
```diff
@@ -0,0 +1,111 @@
...
+
+        graph_params
+      end
+
+      def safe_get_primary_type
+        object_type = params[:object_type].to_s.downcase.singularize
+        allowed_types = GraphRecords::ALLOWED_TYPES.map(&:to_s)
+        return nil unless allowed_types.include?(object_type)
+        object_type.to_sym
+      end
```
misaka, line 108: Just a suggestion here, you could tidy this up to have less symbol <-> string conversions if you change `GraphRecords::DEFAULT_NODE_ORDER` to be strings. Then this could be:

```suggestion
      def safe_get_primary_type
        object_type = params[:object_type].downcase.singularize
        return nil unless GraphRecords::ALLOWED_TYPES.include?(object_type)
        object_type.to_sym
      end
```

lib/graph_records.rb
```diff
@@ -1,23 +1,206 @@
...
+    fill:#aaffc3,color:black
+    fill:#808000,color:white
+    fill:#ffd8b1,color:black
+    fill:#000075,color:white
+    fill:#a9a9a9,color:white
+    fill:#ffffff,color:black
+    fill:#000000,color:white
+  ].freeze
+
+  DEFAULT_NODE_ORDER = %i[
```
misaka, line 38: As part of the comment above on `Inspect::Graph::GraphController#safe_get_primary_type`, it may simplify other parts of the code if these are strings:

```suggestion
  DEFAULT_NODE_ORDER = %w[
```

lib/graph_records.rb
```diff
@@ -86,49 +303,87 @@ def inspect(obj)
...
   end
 
-  def load_association(association_name, associated_objects)
-    if respond_to?("#{association_name}_loader")
-      send("#{association_name}_loader", associated_objects)
-    else
-      associated_objects
+  def order_nodes(*nodes)
+    nodes.sort_by do |node|
+      @node_order.index(node.class.name.underscore.to_sym) || Float::INFINITY
```
misaka, line 333: If we were to change `DEFAULT_NODE_ORDER` to be strings instead of symbols, this could be slightly simplified to:

```suggestion
      @node_order.index(node.class.name.underscore) || Float::INFINITY
```

app/controllers/inspect/graph/graphs_controller.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+module Inspect
+  module Graph
```
misaka, line 4: I think this extra `Graph` module might be redundant. We could remove it and instead of `Inspect::Graph::GraphController` we'd have `Inspect::GraphController`. Is there a reason to leave it in?

app/controllers/inspect/graph/graphs_controller.rb
```diff
@@ -0,0 +1,111 @@
...
+      layout "full"
+
+      def show
+        @primary_type = safe_get_primary_type
+        if @primary_type.nil?
+          render plain:
+                   "You don't have permission to view object type: #{params[:object_type].to_s.downcase.singularize}",
+                 status: :bad_request and return
+        end
+        @primary_id = params[:object_id].to_i
```
misaka, line 21: Don't need to use the `to_i` here, it even stands out a little bit.

```suggestion
        @primary_id = params[:object_id]
```

app/controllers/inspect/graph/graphs_controller.rb
```diff
@@ -0,0 +1,111 @@
...
+              primary_type: @primary_type,
+              clickable: true
+            )
+            .graph(**@graph_params)
+            .join("\n")
+      end
+
+      private
+
+      def build_traversals_config
```
misaka, line 51: Would it make sense to be more specific about the purpose of this method? `list_types_used` or similar? Might help set expectations for future viewer(s).

lib/graph_records.rb
```diff
@@ -27,41 +210,71 @@ def graph(**objects)
...
-          class_name.classify.constantize.where(id: ids)
+    if @primary_type.nil?
+      @primary_type =
+        (
+          if objects.keys.size >= 1
+            objects.keys.first.to_s.singularize.to_sym
+          else
+            :patient
+          end
         )
```
misaka, line 221: Ruby has a handy operator `||=` that allows you to assign a variable if it's not already assigned a value.

```suggestion
    @primary_type ||=
      if objects.keys.size >= 1
        objects.keys.first.to_s.singularize.to_sym
      else
        :patient
      end
```

lib/graph_records.rb
```diff
@@ -27,41 +210,71 @@ def graph(**objects)
...
+    end
+
+    begin
```
misaka, line 224: Is the `begin` here just for the `rescue` block further down? If so, I don't think you need it, especially since you have an `else` block in the rescue that re-raises any unexpected errors. You can remove this `begin` which'll reduce the nesting level for most of the code in this method.

lib/graph_records.rb
```diff
@@ -27,41 +210,71 @@ def graph(**objects)
...
-      patient: "fill:#c2e598",
-      parent: "fill:#faa0a0",
-      consent: "fill:#fffaa0",
-      class_import: "fill:#7fd7df",
-      cohort_import: "fill:#a2d2ff",
-      patient_focused: "fill:#c2e598,stroke:#000,stroke-width:3px",
-      parent_focused: "fill:#faa0a0,stroke:#000,stroke-width:3px"
-    }.with_indifferent_access
+    object_types =
+      @nodes.to_a.map { |node| node.class.name.underscore.to_sym }.uniq
```
misaka, line 260: You can `map` on a `Set`.

```suggestion
      @nodes.map { |node| node.class.name.underscore.to_sym }.uniq
```

lib/graph_records.rb
```diff
@@ -74,7 +287,11 @@ def render_edges
     @edges.map { |from, to| "  #{node_name(from)} --> #{node_name(to)}" }
   end
 
-  def inspect(obj)
+  def render_clicks
+    @nodes.to_a.map { "  click #{node_name(it)} \"#{node_link(it)}\"" }
```
misaka, line 291: Also don't need the `to_a` here.

```suggestion
    @nodes.map { "  click #{node_name(it)} \"#{node_link(it)}\"" }
```

lib/graph_records.rb
```diff
@@ -27,41 +210,71 @@ def graph(**objects)
...
-      patient_focused: "fill:#c2e598,stroke:#000,stroke-width:3px",
-      parent_focused: "fill:#faa0a0,stroke:#000,stroke-width:3px"
-    }.with_indifferent_access
+    object_types =
+      @nodes.to_a.map { |node| node.class.name.underscore.to_sym }.uniq
+
+    styles =
+      object_types.each_with_object({}) do |type, hash|
+        color_index =
+          Digest::MD5.hexdigest(type.to_s).to_i(16) % BOX_STYLES.length
```
misaka, line 265: Interesting idea, I like it.

app/controllers/inspect/graph/graphs_controller.rb
alistair-white-horne-tng: I was under the impression that that would change the endpoint to be `/inspect/{object_type}/{object_id}`. If this change were also applied to Lakshmi's timeline, then it could possibly conflict with her endpoint `/inspect/{patient_id}`.

- Am I right in thinking that this is intrinsically linked to the endpoint? Thinking about it now, I'd expect this to be controlled entirely in `routes.rb`.
- Would it make sense to have both Lakshmi's and my modules under a parent module called `Inspect`, or is it simpler to just have them independent?

app/controllers/inspect/graph/graphs_controller.rb
alistair-white-horne-tng: Ah yes, I think this is actually an issue with the variable inside this variable being ambiguously named; this function does in fact construct the `traversals_config`, and doesn't simply list the types used. I'll rename the dodgy variable instead of the function.

lib/graph_records.rb
alistair-white-horne-tng: I don't quite understand... Are you proposing that:
- I remove the `begin`, but leave the `rescue`? Isn't invalid syntax?
- or that I rewrite the function so it doesn't need to catch an error?

app/controllers/inspect/graph/graphs_controller.rb
alistair-white-horne-tng: Ok, never mind, I think I've addressed it. Hopefully that all ties together as it should... If you have time, might be worth having another quick look at the structure to check that I haven't done anything which is going to cause issue

lib/graph_records.rb
misaka: > - I remove the begin, but leave the rescue? Isn't invalid syntax?

This one. You can use a `rescue` inside a method definition without needing a `begin`. You'd only use that to isolate the code that might raise a particular exception, which I don't think you need here.

app/views/inspect/graphs/show.html.erb
```diff
@@ -0,0 +1,56 @@
...
+      <% end %>
+    <% end %>
+
+    <%= form.govuk_submit "Update graph", name: nil %>
+  <% end %>
+<% end %>
+
+<%= render AppDetailsComponent.new(summary: "Raw mermaid", expander: true, open: false) do %>
+  <pre class="nhsuk-u-font-family-monospace nhsuk-u-margin-0" style="font-size: small;"><%= @mermaid %></pre>
+<% end %>
```
misaka, line 56: Surprised the linter isn't bringing this up, but this file is missing a newline at the very end.

misaka (approved):
Looking good, nice work!

lib/graph_records.rb
alistair-white-horne-tng: Ah ye, I see. Fixed now

PR #3262, opened 2025-03-20
Improve performance generating outcome statuses

thomasleese:
This refactors the way outcome statuses are generated to avoid instantiating ActiveRecord model objects and instead pluck values from the database directly and process the raw values.

Locally, I’m seeing session pages load in about half the time than they would normally with 10,000, 50,000 and 100,000 patients. Further work is necessary to improve performance, likely requiring some sort of state machine or caching of the various statuses so they can be queried directly in SQL.

I’ve tried to make the commits in this PR as small as possible while avoiding breaking the functionality between the commits, but in general the refactors are wide reaching and therefore touch lots of part of the code. The feature tests haven’t required any changes.


app/components/app_programme_session_table_component.rb
```diff
@@ -6,48 +6,60 @@ def initialize(sessions, programme:)
...
 
-    count = session.patient_sessions.select { yield it }.length
+    count = patient_sessions.fetch(session.id, []).select { yield it }.length
 
     number_to_percentage(count / total_count.to_f * 100.0, precision: 0)
   end
 
   def no_response_count(session:)
-    number_stat(session:) { it.patient.consent_outcome.no_response?(programme) }
+    number_stat(session:) do
```
jhenderson, line 43: I'm a bit concerned by the `it` syntax (not just here). It's not apparent what `it` stands for. I had to jump into the method to see what it returns. I think we should use a block variable with a suitable name to make the code more readable (where appropriate).

thomasleese:
This has been superseded by https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3263, https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3268, https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3272, https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3273 and https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3275 which caches the statuses in the database.

PR #3263, opened 2025-03-21
Cache consent status in database

thomasleese:
This is a work in progress/proof of concept to store the consent status of each patient/programme in the database to improve performance.


app/lib/patient_sessions_status_factory.rb
```diff
@@ -0,0 +1,44 @@
...
+      Patient::ConsentStatus.import!(
+        consent_statuses,
+        on_duplicate_key_ignore: true
+      )
+
+      consent_statuses.select(&:persisted?).each(&:refresh!)
+    end
+  end
+
+  def self.call(*args, **kwargs)
```
jhenderson, line 27:
```suggestion
def self.call(...)
  new(...).call
end
```

app/lib/patient_sessions_status_factory.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class PatientSessionsStatusFactory
+  def initialize(patient_session_scope)
+    @patient_sessions =
+      patient_session_scope.eager_load(:patient).preload(
```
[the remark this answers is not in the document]
thomasleese: Yeah, I think that's fair enough. In the past I've got used to specifying `eager_load` explicitly when I want a `JOIN` that's then used in another part of the query, but here it's not needed.

app/lib/patient_merger.rb
```diff
@@ -115,4 +117,12 @@ def self.call(*args, **kwargs)
   private
 
   attr_reader :patient_to_keep, :patient_to_destroy
+
+  def refresh_statuses
+    Patient
+      .includes(consents: :parent)
```
jhenderson, line 123: I think we can remove `includes(consents: :parent)`. We're only fetching a single patient record, so the benefit of eager loading appears to be unnecessary, unless I'm missing a trick somewhere.

thomasleese: I agree in this case, but the `refresh!` method is designed to work with multiple preloaded patients via the `latest_consents` method, and if we don't preload the `consents` then a strict loading exception is raised.

thomasleese: I've updated this PR and introduced a `StatusUpdater` class that handles the calling of `refresh!` which I think helps to clarify the usage.

app/models/patient/consent_status.rb
```diff
@@ -24,8 +24,66 @@ class Patient::ConsentStatus < ApplicationRecord
   belongs_to :patient
   belongs_to :programme
 
+  has_many :consents,
+           -> do
+             not_invalidated.response_provided.eager_load(:parent, :patient)
```
misaka, line 29: Just wondering out loud, but is there any reason for the `consents` association not to always be scoped to `programme`? It seems odd that you could pull up the consents and get unrelated ones for this object. Or is there a technical reason not to?

I also see that `ConsentGrouper` does it's own filtering as part of it's interface, which makes sense, but even so it feels odd to me that the `ConsentStatus` for a particular programme could return unrelated consents.

thomasleese: Basically I tested both approaches, and scoping this to the program ends up needing to make more queries and is slower overall.

If we scope it to the programme, Rails ends up pre-loading the consents 3 times (for each programme) making 3 queries to the database for each batch of consents statuses we want to update.

Whereas if instead we handle the programme filtering in memory, only one query to the database is made, and overall the performance is better and memory usage is lower.

Originally I wanted to scope these (and there's a few others in the other status classes) but I found this to the most performant approach in the end.

misaka (approved):

app/models/patient_session.rb
```diff
@@ -99,6 +99,17 @@ class PatientSession < ApplicationRecord
           order("LOWER(patients.family_name)", "LOWER(patients.given_name)")
         end
 
+  scope :has_consent_status,
```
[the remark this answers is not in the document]
thomasleese: Yeah, it's definitely an improvement on what was there before!

app/components/app_consent_component.rb
```diff
@@ -14,7 +14,7 @@ def initialize(patient_session:, programme:)
 
   def consents
     @consents ||=
-      patient.consent_outcome.all[programme].sort_by(&:created_at).reverse
+      patient.consents.includes(:consent_form, :parent).order(created_at: :desc)
```
misaka, line 17: Does this need to be scoped to `programme`?

thomasleese: Yes, it does, I've amended the commit.

misaka (approved):
🚀  Lots of edge cases, I wonder if we need to beef up our testing.

PR #3265, opened 2025-03-21
Timeline records UI

murugapl:
An endpoint which serves a UI used to view events related to a patient record.

This is a tool designed for use by the Ops Support people, to visualise the patient data without having to use the console.

The information that can be displayed for each 'event' is as follows:
```
cohort_imports: %i[rows_count status uploaded_by_user_id],
class_imports: %i[rows_count year_groups uploaded_by_user_id session_id],
patient_sessions: %i[session_id],
school_moves: %i[source school_id home_educated],
school_move_log_entries: %i[user_id school_id home_educated],
consents: %i[programme_id response route parent_id withdrawn_at invalidated_at],
triages: %i[status performed_by_user_id programme_id invalidated_at],
vaccination_records: %i[outcome performed_by_user_id programme_id session_id vaccine_id]
```

And the following audited changes are allowed:
```
patient_id
session_id
programme_id
vaccine_id
organisation_id
school_id
gp_practice_id

uploaded_by_user_id
performed_by_user_id
user_id
parent_id

status
outcome
response
route

date_of_death_recorded_at
restricted_at
invalidated_at
withdrawn_at

rows_count
year_groups
home_educated
source
```

image

### DIFF
--- a/app/assets/stylesheets/_timeline.scss
+++ b/app/assets/stylesheets/_timeline.scss
@@ -1,0 +1,74 @@
+.timeline {
+  position: relative;
+  margin-bottom: govuk-spacing(4);
+  margin-left: govuk-spacing(2);
+
+  &::before {
+    content: "";
+    position: absolute;
+    top: govuk-spacing(2);
+    left: 0;
+    width: 5px;
+    height: 100%;
+    background-color: $govuk-brand-colour;
+  }
+}
+
+.timeline__item {
+  position: relative;
+  padding-bottom: govuk-spacing(6);
+  padding-left: govuk-spacing(4);
+
+  &::before {
+    content: "";
+    position: absolute;
+    top: govuk-spacing(2);
+    left: 0;
+    width: 15px;
+    height: 5px;
+    background-color: $govuk-brand-colour;
+  }
+}
+
+.timeline__item--date {
+  position: relative;
+  padding-left: govuk-spacing(4);
+  padding-bottom: govuk-spacing(6);
+  margin-top: govuk-spacing(1);
+  margin-bottom: 0;
+  @include govuk-font($size: 16);
+
+  &::before {
+    content: "";
+    position: absolute;
+    left: -9.5px;
+    top: 50%;
+    transform: translateY(-125%);
+    display: block;
+    width: 24px;
+    height: 24px;
+    border-radius: 50%;
+    background-color: $govuk-brand-colour;
+  }
+}
+
+.timeline__header {
+  margin-bottom: govuk-spacing(1);
+}
+
+.timeline__title {
+  display: inline;
+  @include govuk-font($size: 16);
+}
+
+.timeline__byline {
+  display: inline;
+  margin: 0;
+  color: $govuk-secondary-text-colour;
+  @include govuk-font($size: 16);
+}
+
+.timeline__description {
+  margin-top: govuk-spacing(4);
+  @include govuk-font($size: 16);
+}
--- a/app/components/app_timeline_filter_component.html.erb
+++ b/app/components/app_timeline_filter_component.html.erb
@@ -1,0 +1,174 @@
+<%= render AppCardComponent.new(filters: true) do |card| %>
+    <% card.with_heading { "Customise timeline" } %>
+    <%= form_with url: @url,
+                  method: :get,
+                  data: { module: "autosubmit",
+                          turbo: "true",
+                          turbo_action: "replace" },
+                  builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+        <%= f.govuk_fieldset legend: { text: "Events to display:", size: "s" } do %>
+            <% event_options.keys.each do |value| %>
+                <%= f.govuk_check_boxes_fieldset :event_names, legend: { hidden: true } do %>
+                    <%= f.govuk_check_box :event_names,
+                                          value,
+                                          label: { text: value.to_s.humanize },
+                                          checked: value.to_s.in?(params[:event_names] || event_options.keys.map(&:to_s)),
+                                          "data-autosubmit-target": "field",
+                                          "data-action": "autosubmit#submit",
+                                          "data-turbo-permanent": "true" %>
+                    
+                    <% available_fields = timeline_fields[value.to_sym] || [] %>
+                    <% if available_fields.any? && value.to_s.in?(params[:event_names]) %>
+                        <div class="nhsuk-checkboxes__conditional nhsuk-u-margin-bottom-2">
+                            <% available_fields.each do |field| %>
+                                <%= f.govuk_check_box "detail_config[#{value}]",
+                                                      field,
+                                                      small: true,
+                                                      label: { text: field },
+                                                      checked: field.to_s.in?(params.dig("detail_config", value) || []),
+                                                      "data-autosubmit-target": "field",
+                                                      "data-action": "autosubmit#submit",
+                                                      "data-turbo-permanent": "true" %>
+                            <% end %>
+                        </div>
+                    <% end %>
+                <% end %>
+            <% end %>
+
+            <%= f.govuk_check_boxes_fieldset :audit_config, legend: { hidden: true } do %>
+                <%= f.govuk_check_box :event_names, "audits",
+                                      label: { text: "Audits" },
+                                      checked: "audits".in?(params[:event_names]),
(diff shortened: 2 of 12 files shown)


app/components/app_timeline_table_component.rb
```diff
@@ -0,0 +1,38 @@
...
+    mapping = {
+      "CohortImport" => "blue",
+      "ClassImport" => "purple",
+      "PatientSession" => "green",
+      "Consent" => "yellow",
+      "Triage" => "red",
+      "VaccinationRecord" => "grey",
+      "SchoolMove" => "orange",
+      "SchoolMoveLogEntry" => "pink"
+    }
```
thomasleese, line 35: This could be defined as a constant in this class, meaning it wouldn't need to be instanciated in memory each time, for example:

```ruby
# outside the method
MAPPING = {
   ...
}

# in the method
MAPPING.fetch(...)
```

app/controllers/inspect/timeline/patients_controller.rb
```diff
@@ -0,0 +1,128 @@
...
+          hash[event_type.to_sym] = selected_fields
+        end
+      end
+
+      def show
+        params.reverse_merge!(event_names: DEFAULT_EVENT_NAMES)
+        params[:audit_config] ||= {}
+
+        params.dig(:audit_config, :include_associated_audits)
+        params.dig(:audit_config, :include_filtered_audit_changes)
```
thomasleese, line 81: These lines don't do anything on their own, as the return value from `dig` is ignored, can these be removed or do they need to be used?

app/views/inspect/timeline/patients/show.html.erb
```diff
@@ -0,0 +1,77 @@
...
+                  heading: "Invalid patient ID",
+                  description: "Patient-#{params[:manual_patient_id]} doesn't exist",
+                ) %>
+          </div>
+        <% elsif @no_events_compare_message %>
+          <div class="app-timeline__column">
+            <%= render AppWarningCalloutComponent.new(
+                  heading: "No events found for comparison patient",
+                  description: "Patient-#{@compare_patient.id} doesn't have the following events recorded: " +
+                               "#{params[:event_names].mapTimelineRecords::AVAILABLE_DETAILS_CONFIG.join(", ")}".html_safe,
```
thomasleese, line 47: Something doesn’t quite look right here with `#{params[:event_names].mapTimelineRecords::`, so we need to call `map` with an argument?

lib/timeline_records.rb
```diff
@@ -0,0 +1,298 @@
...
+    withdrawn_at
+    rows_count
+    year_groups
+    home_educated
+    source
+  ].freeze
+
+  def initialize(patient_id, detail_config: {}, audit_config: {})
+    @patient = Patient.find(patient_id)
+    @patient_id = patient_id
```
thomasleese, line 72:
```suggestion
  def initialize(patient, detail_config: {}, audit_config: {})
    @patient = patient
    @patient_id = patient.id
```

It looks like everywhere we call `TimelineRecords.new` we've got a `Patient` available so we don't need to fetch it each time.

lib/timeline_records.rb
```diff
@@ -0,0 +1,298 @@
...
+    when :audits
+      @events += audits_events
+    else
+      puts "No handler for event type: #{event_type}"
+    end
+  end
+
+  def org_cohort_imports_events
+    @additional_events[:cohort_imports].map do |cohort_import_id|
+      cohort_import = CohortImport.find(cohort_import_id)
```
thomasleese, line 193:
```suggestion
    CohortImport.where(id: @additional_events[:cohort_imports]).map do |cohort_import|
```

I would suggest doing something like this (and in other similar places) to make only 1 query to the database and fetch all the imports, rather than needing to make a query to the database for each one.

lib/timeline_records.rb
```diff
@@ -0,0 +1,298 @@
...
+        end
+      )
+
+    audit_events.map do |audit|
+      filtered_changes =
+        audit
+          .audited_changes
+          .transform_keys(&:to_s)
+          .each_with_object({}) do |(key, value), hash|
+            if ALLOWED_AUDITED_CHANGES.include?(key.to_sym)
```
thomasleese, line 233: Here we're transforming the keys to strings (`to_s`), and then later on back to symbols again (`to_sym`), is there a way we can avoid this extra processing?

I'm wondering if `audited_changes` will already have string keys since it'll be coming from the database.

app/views/inspect/timeline/patients/show.html.erb
murugapl: Yep I have no clue how/when that happened! Have fixed

thomasleese (approved):
Looks great!

paulrobertlloyd:
@murugapl Few comments from me:

- What does `Patient.find(88)` refer to? Should it be here?
- It looks like the grid markup is slightly off (going by margin on left-hand side)
- Let’s replicate the SAIS-facing UI by using a card heading for the filters
- Remove bottom margin on checkbox fieldsets in the filter
- Wondering if it also might make sense to:
  - Change page title to ‘Timeline for patient 88’
  - Use a table/panel per date (so in the above example, there would be 2 cards, 19 March 2025 and 18 March 2025)

app/assets/stylesheets/_timeline.scss
paulrobertlloyd, line 1: Is this file needed? Do the existing styles for the SAIS-facing UI not provide the intended formatting? (We need to be careful adding CSS… arguably there’s already too much)

app/components/app_timeline_filter_component.html.erb
```diff
@@ -0,0 +1,174 @@
...
+<%= form_with url: @url,
+              method: :get,
+              class: "app-timeline__filters",
+              data: { module: "autosubmit",
+                      turbo: "true",
+                      turbo_action: "replace" },
+              builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+    <%= f.govuk_fieldset legend: { text: "Events to display:", size: "s" } do %>
+        <% event_options.keys.each do |value| %>
+            <%= f.govuk_check_boxes_fieldset :event_names, legend: { hidden: true } do %>
```
paulrobertlloyd, line 11: Not sure if this’ll work, but should hopefully reduce the gap between each set of checkboxes

```suggestion
            <%= f.govuk_check_boxes_fieldset :event_names, legend: { hidden: true }, classes: "nhsuk-u-margin-bottom-2" do %>
```

paulrobertlloyd (changes requested):

paulrobertlloyd:
> It looks like the grid markup is slightly off

Oh, looks like this is not using the grid system; this’ll help you layout the page: https://service-manual.nhs.uk/design-system/styles/layout#grid

app/components/app_timeline_filter_component.html.erb
murugapl: This works when added to the conditional checkboxes instead of the fieldset :)

app/assets/stylesheets/_timeline.scss
murugapl: Gotten rid of it now that I have discovered the grid layout - thank you!

misaka (approved):
Let's get this 🚢 ed

misaka:
Ugh. I think the linting tests are failing because Rubocop and Prettier are having a disagreement on using `do ... end` vs `{ ... }` for chained blocks. Annoying, I think we need a longer term fix for this.

image

For now I think @murugapl you want to add the following into [`AppTimelineTableComponent#format_event_details`](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3265/files#diff-d6917ebd6406bdb089c800bd41df7fc27245793a362213daab1edf7d1e032c21R26)

```
# stree-ignore
```

This disables the Prettier formatting in that method. See [this search of the repo for `stree-ignore` for examples](https://github.com/search?q=repo:nhsuk/manage-vaccinations-in-schools+stree-ignore&type=code).

paulrobertlloyd:
There a few tiny tweaks I’d make (and I mean _really_ tiny), but don’t let me stop this from 🚢ing!

This is sooo cool. Hope it helps with support!

misaka:
Issue accepted in SonarQube, just waiting for @paulrobertlloyd to approve.

paulrobertlloyd (approved):

PR #3268, opened 2025-03-24
Cache triage status in database

thomasleese:
This introduces a new model, `Patient::TriageStatus`, which stores the triage status of a patient-programme pair in the database so it can queried and rendered quickly. Locally, I'm seeing the triage tab on sessions with 50,000 patients load almost instantly even when filtering on statuses. The session overview page has a slight improvement, but that will only load quickly once the other statuses are cached.

I've decided not to implement a state machine as had been previously discussed, but this could be built on top of this change to enhancement the functionality. For now, this effectively caches the existing logic that runs on the fly to the database.

I've also decided not to implement this using Rails callbacks and instead be explicit where the status is refreshed. This allows us to optimise for bulk refreshes, as is necessary when adding new programmes to sessions, running the seeds, or generally updating the status of an entire session.

This follows up from #3263 which applies the same to the consents.


app/components/app_patient_page_component.html.erb
```diff
@@ -72,14 +72,14 @@
       ) %>
 <% end %>
 
-<% if patient.triage_outcome.all[programme].any? %>
```
misaka, line 75: This needs to be scoped to the programme.

misaka (approved):

PR #3275, opened 2025-03-25
Cache register status in database

thomasleese:
This introduces a new model, `PatientSession::RegistrationStatus`, which stores the vaccination status of a patient-session date pair in the database so it can queried and rendered quickly. Locally, I'm seeing the register outcomes page on programmes with 50,000 patients load almost instantly even when filtering on statuses. The session overview page has a slight improvement, but that will only load quickly once the other statuses are cached.

I've decided not to implement a state machine as had been previously discussed, but this could be built on top of this change to enhancement the functionality. For now, this effectively caches the existing logic that runs on the fly to the database.

I've also decided not to implement this using Rails callbacks and instead be explicit where the status is refreshed. This allows us to optimise for bulk refreshes, as is necessary when adding new programmes to sessions, running the seeds, or generally updating the status of an entire session.

This follows up from https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3273 which applies the same to the session status.


app/lib/status_updater.rb
```diff
@@ -141,6 +170,25 @@ def patient_statuses_to_import
...
   end
 
+  def patient_session_registration_statuses_to_import
+    @patient_session_registration_statuses_to_import ||=
+      patient_sessions
+        .eager_load(:patient)
+        .includes(session: %i[session_dates])
+        .find_each
+        .flat_map do |patient_session|
+          unless programme_ids_per_year_group.key?(
```
jhenderson, line 180:
```suggestion
year_group = patient_session.patient.year_group
        
next [] unless programme_ids_per_year_group.key?(year_group)
```

misaka (approved):

app/lib/status_updater.rb
thomasleese: The code has changed a bit since this comment, it should be more readable now.

PR #3277, opened 2025-03-26
Prepare training environment migration

bogsi17:
* Add backend config
* Add resource identifiers for import script


TheOneFromNorway (approved):
Resources look accurate after checking via aws console

PR #3281, opened 2025-03-26
Adapt continuous deployment workflow for terraform

bogsi17:
* Enable automated deployment to QA and Test on every push to main
* Enable manual trigger to allow one-click deployments to any branch


.github/workflows/continuous-deployment.yml
```diff
@@ -1,23 +1,32 @@
...
+run-name: Continuous deployment of ${{ github.ref_name }} to QA and Test
 
 on:
   push:
-    branches: [main]
+    branches: [MAV-921_continuous_deployment_job]
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Deployment environment"
```
thomasleese, line 10:
```suggestion
        description: Deployment environment
```

thomasleese (commented):
It looks to me like the `deploy-mavis.yml` and `continuous-deployment.yml` workflows could be combined in to a single workflow.

Since `continuous-deployment.yml` is the only workflow that uses the `deploy-mavis.yml` workflow, I think the steps in `deploy-mavis.yml` can be moved in to `continuous-deployment.yml`.

I think it's potentially confusing at the moment as they would both show up in the GitHub UI as jobs that can be run manually and it's not immediately clear what the difference is. I suppose once the migration is over and we delete the old `deploy.yml` and `_deploy-template.yml` workflows, which one of these that we keep can be renamed to `deploy.yml` which should help clarify what it's for.

.github/workflows/continuous-deployment.yml
bogsi17: After the last change, their purposes are more separated:

`continuous-deployment.yml` is now purely for continuous deployment to QA and test and cannot be run manually anymore.

`deploy-mavis.yml` is intended to be triggered manually. This is the standard workflow to perform a manual deployment to any environment, including production

.github/workflows/continuous-deployment.yml
```diff
@@ -11,13 +11,12 @@ jobs:
...
+  deploy-mavis:
+    needs: build-and-push-image
+    strategy:
+      fail-fast: false
+      matrix:
+        environment: ["qa", "test"]
+    uses: ./.github/workflows/deploy-mavis.yml
     with:
-      environment: copilotmigration
+      environment: ${{ matrix.environment }}
```
TheOneFromNorway, line 22: Also need to include image_tag as an input (using ${{ github.sha }} should work)

TheOneFromNorway (approved):

PR #3290, opened 2025-03-31
Rename `disable_pds` variable 

thomasleese:
This name wasn't very clear as to exactly what would be enabled/disabled by setting the value. Instead, I've renamed the variable in Terraform and the associated variable in Rails to make it clear that it's referring only the jobs which update patients in bulk to reduce the number of unnecessary PDS API calls.

### DIFF
--- a/app/jobs/bulk_update_patients_from_pds_job.rb
+++ b/app/jobs/bulk_update_patients_from_pds_job.rb
@@ -8,7 +8,7 @@ class BulkUpdatePatientsFromPDSJob < ApplicationJob
   good_job_control_concurrency_with perform_limit: 1
 
   def perform
-    return unless Settings.pds.perform_jobs
+    return unless Settings.pds.enqueue_bulk_updates
 
     patients = Patient.with_nhs_number.not_invalidated.not_deceased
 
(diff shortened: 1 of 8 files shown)


terraform/app/env/training.tfvars
```diff
@@ -19,10 +19,10 @@ rails_env             = "staging"
 rails_master_key_path = "/copilot/mavis/secrets/STAGING_RAILS_MASTER_KEY"
 splunk_enabled        = "false"
 cis2_enabled          = "false"
-pds_enabled           = "false"
+perform_bulk_pds_jobs = "false"
```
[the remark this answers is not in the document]
thomasleese: Ah, that's what I'd called the variable initially, I actually meant to remove it from this file as `false` is the default.

app/jobs/bulk_update_patients_from_pds_job.rb
```diff
@@ -8,7 +8,7 @@ class BulkUpdatePatientsFromPDSJob < ApplicationJob
   good_job_control_concurrency_with perform_limit: 1
 
   def perform
-    return unless Settings.pds.perform_jobs
+    return unless Settings.pds.perform_bulk_jobs
```
misaka, line 11: What about `enqueue_bulk_updates`? Just because there's multiple jobs here, so it's not immediately clear which one it refers too.

thomasleese: Yeah that's a good point, I've amended the commit and used your suggestion.

tvararu (approved):

PR #3291, opened 2025-03-31
Refactor Terraform feature enable variables

thomasleese:
This makes a number of small changes to the three variables related to enabling/disabling features in various environments to improve the maintainability and understandability of the variables.

Specifically, this changes the data type to boolean, and inverses the logic such that all the features are enabled by default (ready for production) and need to be explicitly disabled in non-production environments.

### DIFF
--- a/terraform/app/variables.tf
+++ b/terraform/app/variables.tf
@@ -143,23 +143,23 @@ variable "image_digest" {
   nullable    = false
 }
 
-variable "cis2_enabled" {
+variable "enable_cis2" {
   type        = bool
-  default     = false
+  default     = true
   description = "Boolean toggle to determine whether the CIS2 feature should be enabled."
   nullable    = false
 }
 
-variable "pds_enabled" {
+variable "enable_pds" {
   type        = bool
-  default     = false
+  default     = true
   description = "Boolean toggle to determine whether the PDS feature should be enabled."
   nullable    = false
 }
 
-variable "splunk_enabled" {
+variable "enable_splunk" {
   type        = bool
-  default     = false
+  default     = true
   description = "Boolean toggle to determine whether the Splunk feature should be enabled."
   nullable    = false
 }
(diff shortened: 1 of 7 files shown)


terraform/app/variables.tf
```diff
@@ -194,15 +195,15 @@ locals {
     },
     {
       name  = "MAVIS__CIS2__ENABLED"
-      value = var.cis2_enabled
+      value = var.disable_cis2 ? "false" : "true"
```
TheOneFromNorway, line 198: Double negative logic is always a pain-point in code. Having an input in terraform be .._disabled and turning it into .._ENABLED also seems confusing. I would rather we update both the terraform configuration and the variable's usage in the application

thomasleese: Yeah, I didn't love it either. I think changing this with no downtime requires more work as we'll need to deploy with both environment variables, update the code to support the new variable, and then deploy again with only one environment variable.

I agree about the change, but I would prefer to do it a separate PR to avoid increasing the scope too much.

TheOneFromNorway: I don't understand the deployment issue, if we deploy an image that has variables which align with the terraform variables then it should just work unless I miss something.
IMO I think the scope is "variable change", I think it is more clear to do this at once instead of spread over PRs. Alternatively we could move all the formatting to a separate PR if the issue is too many edits.

TheOneFromNorway: > I imagined a scenario where perhaps the old application code would be deployed under the new environment variables, but maybe that's not the case.

No this should not happen

>  needing lots of code like: if !Settings.cis2.disabled or unless Settings.pds.dont_enqueue_bulk_updates

That is definitely worse! The other option would be to only add these elements if they differ from the default (but that is very messy to do in terraform I believe), or simply not use this negative logic at all (e.g. stick with ..._ENABLE as variable names but have their default be "true", and only overwrite them to "false" if need be. This achieves the same effect functionally. We do not set any "disable" variables (which would be nice) but it avoids the confusing logic of setting false a variable which is true in the application.

thomasleese: >or simply not use this negative logic at all (e.g. stick with ..._ENABLE as variable names but have their default be "true", and only overwrite them to "false" if need be.

I'm happy with this approach, actually it's what I had originally, but after discussion with @misaka we liked the idea of calling out when a feature is disabled. Maybe it's not worth it for this potential confusion.

TheOneFromNorway: I too like .._disabled, just not the implementation in the end. Then I think it makes sense to stick with the original values but swap the default logic 👍

thomasleese: I've dropped the commit which inverted the logic from this PR.

PR #3294, opened 2025-03-31
Allow build and deployment with single workflow

bogsi17:
By including the build step into `deploy-mavis.yml`, it's possible to deploy branches within a single workflow. Previously, this required to run two workflow, first the build-and-push-image workflow and then the deploy-mavis workflow.

The `build-and-push-image.yml` workflow now checks if an image already exists and only builds it if it's not yet present in both ECR repositories. Apart from general efficiency, this allows us to run the workflow as part of production deployments where we generally don't want to build a new image and rather deploy an image that already exists and was tested on other environments.

### DIFF
--- a/.github/workflows/build-and-push-image.yml
+++ b/.github/workflows/build-and-push-image.yml
@@ -5,7 +5,43 @@ on:
   workflow_call:
 
 jobs:
+  check-image-presence:
+    name: Check if images already exist
+    runs-on: ubuntu-latest
+    outputs:
+      build_needed: ${{ steps.check_image.outputs.BUILD_NEEDED }}
+    steps:
+      - name: Configure AWS Dev Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::393416225559:role/GitHubActionsRole
+          aws-region: eu-west-2
+      - name: Check if dev image exists
+        run: |
+          if aws ecr describe-images --repository-name mavis/webapp --image-ids imageTag=${{ github.sha }} > /dev/null 2>&1; then
+            echo "Dev image with given tag already exists"
+          else
+            echo "Dev image does not exist. Build needed"
+            echo "BUILD_NEEDED=true" >> $GITHUB_ENV
+          fi
+      - name: Configure AWS Prod credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::820242920762:role/GitHubActionsRole
+          aws-region: eu-west-2
+      - name: Check if prod image exists
+        id: check_image
+        run: |
+          if [ -e $BUILD_NEEDED ] && aws ecr describe-images --repository-name mavis/webapp --image-ids imageTag=${{ github.sha }} > /dev/null 2>&1; then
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-push-image.yml
```diff
@@ -4,8 +4,49 @@ on:
...
 
+concurrency:
+  group: build-and-push-image-${{ github.sha }}
+
 jobs:
+  check-image-presence:
+    name: Check if images already exist
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
```
thomasleese, line 15: Do we need this permission, I can't see it being used in this job?

.github/workflows/build-and-push-image.yml
```diff
@@ -4,8 +4,49 @@ on:
...
+          role-to-assume: arn:aws:iam::820242920762:role/GithubDeployMavisAndInfrastructure
+          aws-region: eu-west-2
+      - name: Check if prod image exists
+        id: check_image
+        run: |
+          if [ -e $BUILD_NEEDED ] && aws ecr describe-images --repository-name mavis/webapp --image-ids imageTag=${{ github.sha }} > /dev/null 2>&1; then
+            echo "Prod and dev images with given tag already exist. No build needed"
+          else
+            echo "At least one image does not exist. Build needed"
+            echo "BUILD_NEEDED=true" >> $GITHUB_OUTPUT
```
thomasleese, line 44:
```suggestion
            echo "build-needed=true" >> $GITHUB_OUTPUT
```

I think the convention with GitHub Actions is typically lowercase variables with dashes.

.github/workflows/build-and-push-image.yml
```diff
@@ -4,8 +4,49 @@ on:
...
+            echo "Dev image does not exist. Build needed"
+            echo "BUILD_NEEDED=true" >> $GITHUB_ENV
+          fi
+      - name: Configure AWS Prod credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::820242920762:role/GithubDeployMavisAndInfrastructure
+          aws-region: eu-west-2
+      - name: Check if prod image exists
+        id: check_image
```
thomasleese, line 38:
```suggestion
        id: check-image
```

.github/workflows/build-and-push-image.yml
```diff
@@ -4,8 +4,49 @@ on:
...
+          else
+            echo "Dev image does not exist. Build needed"
+            echo "BUILD_NEEDED=true" >> $GITHUB_ENV
+          fi
+      - name: Configure AWS Prod credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::820242920762:role/GithubDeployMavisAndInfrastructure
+          aws-region: eu-west-2
+      - name: Check if prod image exists
```
thomasleese, line 37:
```suggestion
      - name: Check if production image exists
```

.github/workflows/build-and-push-image.yml
bogsi17: It's needed for the configure-aws-credentials action

thomasleese (approved):

PR #3300, opened 2025-04-01
Improve patient search result ordering

alistair-white-horne-tng:
Order search results by match similarity before alphabetical name

[MAV-951](https://nhsd-jira.digital.nhs.uk/browse/MAV-951)

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -9,10 +9,7 @@ class PatientsController < ApplicationController
   before_action :record_access_log_entry, only: %i[show log]
 
   def index
-    patients =
-      @form.apply(
-        policy_scope(Patient).includes(:school).not_deceased.order_by_name
-      )
+    patients = @form.apply(policy_scope(Patient).includes(:school).not_deceased)
 
     @pagy, @patients = pagy(patients)
 
--- a/spec/forms/search_form_spec.rb
+++ b/spec/forms/search_form_spec.rb
@@ -106,6 +106,54 @@ describe SearchForm do
         expect(form.apply(scope, programme:)).to include(patient)
       end
     end
+
+    context "searching on name" do
+      let(:consent_status) { nil }
+      let(:date_of_birth_day) { nil }
+      let(:date_of_birth_month) { nil }
+      let(:date_of_birth_year) { nil }
+      let(:missing_nhs_number) { nil }
+      let(:programme_status) { nil }
+      let(:q) { nil }
+      let(:register_status) { nil }
+      let(:session_status) { nil }
+      let(:triage_status) { nil }
+      let(:year_groups) { nil }
+
+      let(:patient_a) do
+        create(:patient, given_name: "Harry", family_name: "Potter")
+      end
+      let(:patient_b) do
+        create(:patient, given_name: "Hari", family_name: "Potter")
+      end
+      let(:patient_c) do
+        create(:patient, given_name: "Arry", family_name: "Pott")
+      end
+      let(:patient_d) do
+        create(:patient, given_name: "Ron", family_name: "Weasley")
+      end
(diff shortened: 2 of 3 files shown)


app/controllers/patients_controller.rb
```diff
@@ -9,10 +9,7 @@ class PatientsController < ApplicationController
   before_action :record_access_log_entry, only: %i[show log]
 
   def index
-    patients =
-      @form.apply(
-        policy_scope(Patient).includes(:school).not_deceased.order_by_name
-      )
+    patients = @form.apply(policy_scope(Patient).includes(:school).not_deceased)
```
thomasleese, line 12: I think we still need the `order_by_name`, without this, if the user isn't searching for a name, the ordering won't be guaranteed.

alistair-white-horne-tng: This is already present [here](https://github.com/alistair-white-horne-tng/manage-vaccinations-in-schools/blob/41996c9d50db11c330b953b1d14b40f75b3c753d/app/forms/search_form.rb#L43)

spec/forms/search_form_spec.rb
```diff
@@ -106,6 +106,83 @@
         expect(form.apply(scope, programme:)).to include(patient)
       end
     end
+
+    context "searching on name" do
```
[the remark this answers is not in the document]
alistair-white-horne-tng: Good point. I've now moved them into `patient_spec.rb`, and added some feature tests where these were.

spec/forms/search_form_spec.rb
```diff
@@ -106,6 +106,60 @@
...
+        create(:patient, given_name: "Ginny", family_name: "Weasley")
+      end
+
+      context "with no search query" do
+        let(:q) { nil }
+
+        it "sorts alphabetically by name" do
+          [patient_a, patient_b, patient_c, patient_d, patient_e] # forces the creation of the patients in memory
+
+          expect(form.apply(scope).count).to eq(5)
```
thomasleese, line 145: I don't think we need this expectation since the line below it checks that the array matches exactly an array of length 5. Plus by only doing the line below the patients will be executed before `form.apply` avoiding the need for line 143.

spec/forms/search_form_spec.rb
```diff
@@ -106,6 +106,60 @@
...
+        end
+      end
+
+      context "with some search query" do
+        let(:q) { "Harry Potter" }
+
+        it "sorts by similarity" do
+          [patient_a, patient_b, patient_c, patient_d, patient_e] # forces the creation of the patients in memory
+
+          expect(form.apply(scope).count).to eq(3)
```
thomasleese, line 158: Similar comment to above.

spec/forms/search_form_spec.rb
```diff
@@ -106,6 +106,60 @@
...
+      end
+      let(:patient_e) do
+        create(:patient, given_name: "Ginny", family_name: "Weasley")
+      end
+
+      context "with no search query" do
+        let(:q) { nil }
+
+        it "sorts alphabetically by name" do
+          [patient_a, patient_b, patient_c, patient_d, patient_e] # forces the creation of the patients in memory
```
thomasleese, line 143: Usually when we need to do this, we can use `let!` instead of `let` which evaluates immediately rather than lazily.

alistair-white-horne-tng: Ahhhh! This makes so much more sense, I feel like I now understand the lazy loading situation. I did a few checks before the `form.apply(scope)` line, and I thought it was successfully matching `[]` with the desired array. Didn't seem right but I wasn't able to prove it wrong. I suppose I wasn't clear as to which line actually triggered the population of the array. Anyway, all sorted now

thomasleese (approved):

PR #3305, opened 2025-04-01
Leicestershire onboarding

murugapl:
The responses which the yaml is based off) from Leicestershire are in the Microsoft form shared on slack


thomasleese:
The training configuration was added in #3306, we can re-open this when we're ready to add the configuration in production and know whether Leicestershire will be using clinics.

PR #3308, opened 2025-04-02
Support importing SystmOne anatomical sites

thomasleese:
This updates the vaccination record importer to support the anatomical sites that we export in the SystmOne format. This should improve compatibility when it comes to importing vaccination records with organisations that use SystmOne.


thomasleese:
Turns out SystmOne exports don't include this information.

PR #3309, opened 2025-04-02
Support importing SystmOne gender codes

thomasleese:
This updates the vaccination record importer to support the gender codes that we export in the SystmOne format. This should improve compatibility when it comes to importing vaccination records with organisations that use SystmOne.

Screenshot 2025-04-02 at 11 05 13


thomasleese:
Turns out SystmOne exports don't consume the gender code in the same way that it is exported.

PR #3312, opened 2025-04-02
Support batches without expiration dates

thomasleese:
When uploading historical vaccination records it's not guaranteed that the batch expiration date will be provided with the batch number. Therefore, we need to handle batches without expiration dates.

This makes a number of changes to the service to allow for batches without expiration dates (specifically to ensure they will render correctly without the user seeing an error) and also updates the import process to support importing historical vaccination records without batches. Offline recording of vaccinations still requires a batch expiration date to be provided.

We also can't be sure that batches with the same number won't be reused, so we allow batches to exist with the same number, one with an expiration date and one without an expiration date. In the future we might be able to consolidate these batches in to one.

### DIFF
--- a/spec/factories/batches.rb
+++ b/spec/factories/batches.rb
@@ -32,10 +32,13 @@ FactoryBot.define do
     vaccine
 
     name { "#{prefix}#{Faker::Number.number(digits: 4)}" }
-    expiry { Faker::Time.forward(days: 50, period: :morning) }
 
     trait :expired do
-      expiry { Date.yesterday }
+      expiry { Faker::Time.backward(days: 50, period: :evening) }
+    end
+
+    trait :not_expired do
+      expiry { Faker::Time.forward(days: 50, period: :morning) }
     end
 
     trait :archived do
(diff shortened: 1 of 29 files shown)


spec/factories/batches.rb
```diff
@@ -32,7 +32,10 @@
     vaccine
 
     name { "#{prefix}#{Faker::Number.number(digits: 4)}" }
-    expiry { Faker::Time.forward(days: 50, period: :morning) }
+
+    trait :not_expired do
```
misaka, line 36: Is this `not_expired` or maybe `with_expiry`? Just thinking that `not_expired` sounds like a batch that has expired in the past, but I believe this is setting an expiry date in the future.

misaka: Actually, ignore that. It suddenly makes more sense with the `expired` trait below.

misaka (approved):

spec/factories/batches.rb
thomasleese: Yeah, I did wonder about `with_expiry`, but then `expired` is also `with_expiry`. I might swap them around though in the file, I think it's more natural for expired to come first.

PR #3313, opened 2025-04-03
Add `ImmunisationImport::RowParser`

thomasleese:
This refactors the code related to parsing the data for vaccination imports to separate the concerns related to parsing the raw values from the CSV file in to suitable Ruby objects, separate to the validation and creation of objects in the database.

This will make no difference to the functionality of the service, but it will allow us to extend the parser to support different file formats without needing to change any code related to validation and importing in to the database.


thomasleese:
This is going to be replaced with a different approach to refactoring the CSV imports.

PR #3314, opened 2025-04-03
Download school moves

jhenderson:
### Context
This PR introduces functionality to export school moves data to a CSV format. This feature is designed to assist new users in keeping their existing systems up-to-date during the transition period to full adoption.

### Changes proposed in this pull request
- New service object to generate CSV exports of school move data
- New controller and views to handle date range input and confirmation page

### Screenshots
| School Moves (1) | Date Range Input (2) | Confirmation and download (3) |
|------------------|---------------------------|---------------------|
|  Screenshot 2025-04-04 at 09 55 36 | Screenshot 2025-04-04 at 09 57 20 | Screenshot 2025-04-04 at 09 57 57 |

### DIFF
--- a/app/models/school_move_export.rb
+++ b/app/models/school_move_export.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+class SchoolMoveExport
+  include RequestSessionPersistable
+  include WizardStepConcern
+
+  attribute :date_from, :date
+  attribute :date_to, :date
+
+  def wizard_steps
+    %i[dates confirm].freeze
+  end
+
+  delegate :count, to: :school_moves
+
+  def self.request_session_key
+    "school_move_export"
+  end
+
+  def csv_data
+    Reports::CSVSchoolMoves.call(school_moves)
+  end
+
+  def csv_filename
+    name_parts = ["school_moves_export"]
+    name_parts << date_from.to_fs(:govuk) if date_from.present?
+    name_parts << "to" if date_from.present? && date_to.present?
+    name_parts << date_to.to_fs(:govuk) if date_to.present?
+
+    "#{name_parts.join("_")}.csv"
+  end
+
+  def date_from_formatted
+    if date_from.present?
+      date_from.strftime("%d %B %Y")
+    else
+      "Earliest recorded vaccination"
+    end
+  end
+
+  def date_to_formatted
+    if date_to.present?
+      date_to.strftime("%d %B %Y")
+    else
+      "Latest recorded vaccination"
+    end
+  end
+
+  private
+
+  def school_moves
+    scope =
+      SchoolMoveLogEntry.where(
+        patient: @current_user.selected_organisation.patients
+      )
+
+    if date_from.present?
+      scope =
+        scope.where(
+          "school_move_log_entries.created_at >= ?",
+          date_from.beginning_of_day
+        )
+    end
+
+    if date_to.present?
+      scope =
+        scope.where(
+          "school_move_log_entries.created_at <= ?",
+          date_to.end_of_day
+        )
+    end
+
+    scope
+  end
+
+  def reset_unused_fields
+  end
+end
--- a/app/views/school_moves/index.html.erb
+++ b/app/views/school_moves/index.html.erb
@@ -1,4 +1,14 @@
-<%= h1 t(".title"), size: "xl" %>
+<%= h1 "School moves", size: "xl" %>
+
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-two-thirds">
+    <p>When imported records or a new consent response indicates that a child has changed
+      school, Mavis flags this as a school move.</p>
+    <p>You can then review the new information and confirm the school move or ignore it.</p>
+
+    <%= govuk_button_to "Download records", school_move_exports_path, class: "app-button--secondary nhsuk-u-margin-bottom-5" %>
+  </div>
+</div>
 
 <% if @school_moves.any? %>
   <div class="nhsuk-table__panel-with-heading-tab">
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -190,6 +190,12 @@ Rails.application.routes.draw do
   end
 
   resources :school_moves, path: "school-moves", only: %i[index show update]
+  resources :school_move_exports,
+            path: "school-moves/exports",
+            controller: "school_moves/exports",
+            only: %i[create show update] do
+    get "download", on: :member
+  end
 
   resources :sessions, only: %i[edit index show], param: :slug do
     resource :consent, only: :show, controller: "sessions/consent"
--- a/spec/features/download_school_moves_spec.rb
+++ b/spec/features/download_school_moves_spec.rb
@@ -1,0 +1,106 @@
+# frozen_string_literal: true
+
+describe "Download school moves" do
+  scenario "no dates given" do
+    given_i_am_signed_in
+    and_school_moves_exist
+    and_i_go_school_moves
+    when_i_click_on_download_records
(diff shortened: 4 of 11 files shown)


app/services/csv_school_moves.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class CSVSchoolMoves
+  include ServicePattern
```
jhenderson, line 4: The `ServicePattern` is something we did at DfE https://github.com/DFE-Digital/register-trainee-teachers along with a `services/` directory. We don't have to follow it, but I thought it's worth putting up for show and get some feedback.

app/services/csv_school_moves.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  HEADERS = %i[
+    full_name
```
jhenderson, line 7: These headers are just stubs.

config/locales/en.yml
```diff
@@ -822,6 +822,37 @@ en:
...
+      download: Download records
+      description:
+        - >-
+          When imported records or a new consent response indicates that a child has changed
+          school, Mavis flags this as a school move.
+        - >-
+          You can then review the new information and confirm the school move or ignore it.
+    download:
+      dates:
+        caption: Schools
```
paulrobertlloyd, line 834:
```suggestion
        caption: School moves
```

config/locales/en.yml
```diff
@@ -822,6 +822,37 @@ en:
...
+        date_to_legend: To (optional)
+        date_range_heading: Select a date range
+        date_hint: For example, %{example}
+        submit: Continue
+        description:
+          You can download a record of school moves as a CSV file. Only school
+          moves that have been reviewed and confirmed will be included.
+        unconfirmed_moves: There are %{count} unconfirmed school moves in your cohort.
+      confirm:
+        caption: Schools
```
paulrobertlloyd, line 846:
```suggestion
        caption: School moves
```

paulrobertlloyd (dismissed):
Caption should match the section name

app/lib/service_pattern.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module ServicePattern
```
thomasleese, line 3: I think we should add this in a separate PR, we already have some service-like classes in `app/lib` so I think if we add this module (which I like) we should also apply it to those classes.

app/policies/school_move_log_entry_policy.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class SchoolMoveLogEntryPolicy < ApplicationPolicy
```
thomasleese, line 3: Since this is only used in the `SchoolMoveExport` I would be tempted to move it straight in to there, we can always pull it out in to a policy if we decided we need to show the log entries in more places.

app/services/csv_school_moves.rb
thomasleese, line 1: I think for now this should go in `app/lib/reports` with the other classes that export reports. We can move it to `app/services` in a separate PR if we want.

app/services/csv_school_moves.rb
```diff
@@ -0,0 +1,75 @@
...
+    DES_NUMBER
+  ].freeze
+
+  def initialize(school_move_log_entries)
+    @school_move_log_entries = school_move_log_entries
+  end
+
+  def call
+    CSV.generate(headers: HEADERS, write_headers: true) do |csv|
+      school_move_log_entries.each do |log_entry|
```
thomasleese, line 33:
```suggestion
      school_move_log_entries.find_each do |log_entry|
```

app/services/csv_school_moves.rb
```diff
@@ -0,0 +1,75 @@
...
+      nil,
+      school_urn(school, patient),
+      school&.name,
+      school&.created_at&.to_fs(:govuk),
+      nil,
+      nil
+    ]
+  end
+
+  def school_urn(school, patient)
```
thomasleese, line 66: I think we should use the existing method in `Reports::ExportFormatters` to do this: https://github.com/nhsuk/manage-vaccinations-in-schools/blob/688c6bab2308b87ff3ff5c35b7f7cc3e4aacfbfa/app/lib/reports/export_formatters.rb#L6-L14

app/views/school_move_exports/dates.html.erb
```diff
@@ -0,0 +1,27 @@
...
+<p class="nhsuk-body">
+  <%= t("school_moves.download.dates.unconfirmed_moves", count: @school_move_export.count) %>
+</p>
+
+<h3><%= t("school_moves.download.dates.date_range_heading") %></h3>
+
+<%= form_with model: @school_move_export, url: wizard_path, method: :put do |f| %>
+  <%= f.govuk_error_summary %>
+  <%= f.govuk_date_field :date_from,
+                         legend: { text: t("school_moves.download.dates.date_from_legend") },
```
thomasleese, line 19: Rather than specifying the locale strings here, we can take advantage of the automatic locale strings stored in `config/locales/helpers.yaml`.

https://govuk-form-builder.netlify.app/building-blocks/localisation/

app/views/school_moves/index.html.erb
```diff
@@ -1,5 +1,15 @@
 <%= h1 t(".title"), size: "xl" %>
 
+<div class="nhsuk-grid-row">
+  <div class="nhsuk-grid-column-two-thirds">
+    <% t(".description").each do |paragraph| %>
```
thomasleese, line 5: In general we've not been using locale strings for hard coded text. Mostly because we don't think it'll need to be translated and it's easier to follow if all the content is in one file (in this case the view). If you think it's better as locale strings then I'm happy with that approach too, just thought I'd highlight that's a little inconsistent.

config/routes.rb
```diff
@@ -190,6 +190,11 @@
   end
 
   resources :school_moves, path: "school-moves", only: %i[index show update]
+  resources :school_move_exports,
+            path: "school-move-exports",
+            only: %i[create show update] do
+    get "download", on: :member
+  end
```
thomasleese, line 197: I would suggest putting this inside the school move resource to keep the paths tidy. Something like:

```rb
resources :school_moves, path: "school-moves", only: %i[index show update] do
  resources :exports, only: %i[create show update] do
    get "download", on: :member
  end
end
```

We can rename the controller to `SchoolMoves::ExportController`.

spec/features/download_school_moves_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    and_i_go_school_moves
+    when_i_click_on_download_records
+    when_i_enter_some_dates
+    then_i_should_see_a_confirmation_page_with_dates
+    when_i_click_download_csv
+    then_i_get_a_csv_file_with_expected_row_count(1)
+  end
+
+  def given_i_am_signed_in
+    programmes = [create(:programme, :hpv)]
```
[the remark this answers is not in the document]
thomasleese: I'm thinking actually you probably don't need to specify the programme in the tests as it's not related to school moves, instead we can let the factories create one at random.

app/models/school_move_export.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def reset_unused_fields
+  end
+
+  def scoped_log_entries
+    scope = SchoolMoveLogEntry
+    organisation = @current_user.selected_organisation
+
+    return scope.none if organisation.nil?
```
thomasleese, line 80: I don't think we need to worry about this, the user won't have got to this page if they haven't selected an organisation.

thomasleese (approved):

PR #3318, opened 2025-04-04
Add `CSVParser` and `CSVParser::Field`

thomasleese:
This adds a new struct which handles the parsing of CSV fields (converting values to integers, dates and times) and stores additional information about the field (the row index, the column index, and the original header). This additional information will be used in a follow up PR to improve the error handling and validation by showing the original header to the user, which allows to easily handle multiple different file formats while presenting to the user the correct headers.

This follows on from #3317.

### DIFF
--- a/app/lib/csv_parser.rb
+++ b/app/lib/csv_parser.rb
@@ -22,21 +22,47 @@ class CSVParser
 
   private_class_method :new
 
+  Field =
+    Struct.new("Field", :value, :column, :row, :header) do
+      delegate :blank?, :present?, to: :value
+
+      alias_method :to_s, :value
+
+      def cell = "#{column}#{row}"
+    end
+
   private
 
   attr_reader :data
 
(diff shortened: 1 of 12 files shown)


app/lib/csv_parser.rb
```diff
@@ -0,0 +1,113 @@
...
+
+        parsed_values.first
+      end
+    end
+
+  private
+
+  attr_reader :data
+
+  ALPHABET = %w[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z].freeze
```
jhenderson, line 77:
```suggestion
  ALPHABET =  ('A'..'Z').to_a.freeze
  ALPHABET + ALPHABET.product(ALPHABET).map(&:join)
```

thomasleese: Ah nice! I've amended the commit to do this instead.

misaka (approved):

PR #3321, opened 2025-04-04
Create ecs service module

TheOneFromNorway:
Create a new module for ecs services. This allows us to simply add additional service types, e.g. such as a background service for good-job

### DIFF
--- a/terraform/app/modules/ecs_service/main.tf
+++ b/terraform/app/modules/ecs_service/main.tf
@@ -1,0 +1,100 @@
+resource "aws_security_group" "this" {
+  name        = "${var.server_type}-service-${var.environment}"
+  description = "Security Group for communication with ECS Service"
+  vpc_id      = var.network_params.vpc_id
+  lifecycle {
+    ignore_changes = [description]
+  }
+}
+
+resource "aws_security_group_rule" "egress_all" {
+  type              = "egress"
+  from_port         = 0
+  to_port           = 0
+  protocol          = "-1"
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = aws_security_group.this.id
+}
+
+resource "aws_ecs_service" "this" {
+  name                              = "mavis-${var.environment}-${var.server_type}"
+  cluster                           = var.cluster_id
+  task_definition                   = aws_ecs_task_definition.this.arn
+  desired_count                     = var.desired_count
+  launch_type                       = "FARGATE"
+  enable_execute_command            = true
+  health_check_grace_period_seconds = 60
+
+  network_configuration {
+    subnets         = var.network_params.subnets
+    security_groups = [aws_security_group.this.id]
+  }
+  deployment_controller {
+    type = var.deployment_controller
+  }
+  dynamic "deployment_circuit_breaker" {
+    for_each = var.deployment_controller == "ECS" ? [1] : []
+    content {
+      enable   = true
+      rollback = true
+    }
+  }
+  dynamic "load_balancer" {
+    for_each = var.loadbalancer != null ? [1] : []
+    content {
+      target_group_arn = var.loadbalancer.target_group_arn
+      container_name   = var.container_name
+      container_port   = var.loadbalancer.container_port
+    }
+  }
+  deployment_minimum_healthy_percent = 100
+  deployment_maximum_percent         = 200
+  lifecycle {
+    ignore_changes = [
+      task_definition,
+      load_balancer,
+      # desired_count TODO: Uncomment once we include autoscaling
+    ]
+    create_before_destroy = true
+  }
+}
+
+resource "aws_ecs_task_definition" "this" {
+  family                   = "mavis-${var.server_type}-task-definition-${var.environment}"
+  requires_compatibilities = ["FARGATE"]
+  network_mode             = "awsvpc"
+  cpu                      = var.task_config.cpu
+  memory                   = var.task_config.memory
+  execution_role_arn       = var.task_config.execution_role_arn
+  task_role_arn            = var.task_config.task_role_arn
+  container_definitions = jsonencode([
+    {
+      name      = var.container_name
+      image     = var.task_config.docker_image
+      essential = true
+      portMappings = [
+        {
+          containerPort = 4000
+          hostPort      = 4000
+        }
+      ]
+      environment = concat(var.task_config.environment, [{ name = "SERVER_TYPE", value = var.server_type }])
+      secrets     = var.task_config.secrets
+      logConfiguration = {
+        logDriver = "awslogs"
+        options = {
+          awslogs-group         = var.task_config.log_group_name
+          awslogs-region        = var.task_config.region
+          awslogs-stream-prefix = "${var.environment}-${var.server_type}-logs"
+        }
+      }
+      healthCheck = {
+        command     = var.task_config.health_check_command
+        interval    = 30
+        timeout     = 5
+        retries     = 3
+        startPeriod = 10
+      }
+    }
+  ])
+}
--- a/terraform/app/modules/ecs_service/variables.tf
+++ b/terraform/app/modules/ecs_service/variables.tf
@@ -1,0 +1,79 @@
+variable "environment" {
+  type        = string
+  description = "Application environment (for example production or staging)"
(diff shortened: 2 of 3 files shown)


terraform/app/modules/ecs_service/main.tf
```diff
@@ -0,0 +1,100 @@
+resource "aws_security_group" "this" {
+  name        = "${var.server_type}-service-${var.environment}"
+  description = "Security Group for communication with Background ECS Service"
```
thomasleese, line 3:
```suggestion
  description = "Security Group for communication with ${var.server_type} service"
```

terraform/app/modules/ecs_service/main.tf
```diff
@@ -0,0 +1,100 @@
...
+      task_definition,
+      load_balancer,
+      # desired_count TODO: Uncomment once we include autoscaling
+    ]
+    create_before_destroy = true
+  }
+}
+
+resource "aws_ecs_task_definition" "this" {
+  family                   = "${var.server_type}-task-definition-${var.environment}"
```
thomasleese, line 63: Wondering about the naming convention here, for the service we have a `mavis-` prefix but here we don't, should we be consistent with the naming?

terraform/app/modules/ecs_service/variables.tf
```diff
@@ -0,0 +1,71 @@
+variable "environment" {
```
thomasleese, line 1: Should this also be `nullable`?

terraform/app/modules/ecs_service/variables.tf
```diff
@@ -0,0 +1,71 @@
...
+  type        = string
+  description = "String literal for the environment"
+}
+
+variable "server_type" {
+  type        = string
+  description = "Type of server to be deployed"
```
thomasleese, line 8: I wonder if we should clarify here what type means, and how it's exposed to an env var?

terraform/app/modules/ecs_service/variables.tf
```diff
@@ -0,0 +1,71 @@
+variable "environment" {
+  type        = string
+  description = "String literal for the environment"
```
thomasleese, line 3:
```suggestion
  description = "Application environment (for example production or staging)"
```

terraform/app/modules/ecs_service/main.tf
```diff
@@ -0,0 +1,100 @@
...
+    content {
+      enable   = true
+      rollback = true
+    }
+  }
+  dynamic "load_balancer" {
+    for_each = var.loadbalancer != null ? [1] : []
+    content {
+      target_group_arn = var.loadbalancer.target_group_arn
+      container_name   = "mavis-application"
```
thomasleese, line 46:
```suggestion
      container_name   = "application"
```

Do we need the prefix?

terraform/app/modules/ecs_service/main.tf
```diff
@@ -0,0 +1,100 @@
...
+    create_before_destroy = true
+  }
+}
+
+resource "aws_ecs_task_definition" "this" {
+  family                   = "${var.server_type}-task-definition-${var.environment}"
+  requires_compatibilities = ["FARGATE"]
+  network_mode             = "awsvpc"
+  cpu                      = 1024
+  memory                   = 2048
```
thomasleese, line 67: Should these be coming from `var.task_config`?

terraform/app/modules/ecs_service/main.tf
```diff
@@ -0,0 +1,100 @@
...
+  execution_role_arn       = var.task_config.execution_role_arn
+  task_role_arn            = var.task_config.task_role_arn
+  container_definitions = jsonencode([
+    {
+      name      = "mavis-application"
+      image     = var.task_config.docker_image
```
thomasleese, line 73: I wonder if it would be nicer to only need to pass in the image tag to the ECS service, I imagine all our services will want to be on the same image URL.

terraform/app/modules/ecs_service/main.tf
TheOneFromNorway: yes, will implement

terraform/app/modules/ecs_service/main.tf
TheOneFromNorway: Definitely not needed, the question is just where we add the mavis prefix and were we dont

thomasleese: I would suggest avoid a `mavis-` prefix unless we really need one, I think it's obvious that all of this relates to Mavis.

terraform/app/modules/ecs_service/variables.tf
TheOneFromNorway: Can add some more description

terraform/app/modules/ecs_service/main.tf
thomasleese: Yeah that's fair enough, I think once we have a single repository we can tidy up a few things like this.

terraform/app/modules/ecs_service/main.tf
TheOneFromNorway: Should definitely drop Background in the description that is a remnant. However the service name already contains the type so I would not duplicate it for the description and rather keep it general (because you cannot modify descriptions with terraform outside the AWS Console without re-creating the resource)

terraform/app/modules/ecs_service/main.tf
TheOneFromNorway: Sure will add

terraform/app/modules/ecs_service/main.tf
```diff
@@ -0,0 +1,100 @@
...
+      task_definition,
+      load_balancer,
+      # desired_count TODO: Uncomment once we include autoscaling
+    ]
+    create_before_destroy = true
+  }
+}
+
+resource "aws_ecs_task_definition" "this" {
+  family                   = "mavis-${var.server_type}-task-definition-${var.environment}"
```
thomasleese, line 63: Do we need the `mavis-` prefix here?

terraform/app/modules/ecs_service/main.tf
```diff
@@ -0,0 +1,100 @@
...
+  requires_compatibilities = ["FARGATE"]
+  network_mode             = "awsvpc"
+  cpu                      = var.task_config.cpu
+  memory                   = var.task_config.memory
+  execution_role_arn       = var.task_config.execution_role_arn
+  task_role_arn            = var.task_config.task_role_arn
+  container_definitions = jsonencode([
+    {
+      name      = "mavis-application"
```
thomasleese, line 72: Is it worth including the server type in this name?

terraform/app/modules/ecs_service/main.tf
thomasleese: Okay, that makes sense. I think we should be consistent with the naming either way, so I think if we want a prefix we need to add it to the security group and log stream.

TheOneFromNorway: I will create another ticket/PR for handling naming chances of log-groups/etc to keep it out of this scope. Would be good to do but I want to restrict this to just the ECS (and naming changes there-in)

terraform/app/modules/ecs_service/main.tf
TheOneFromNorway: Not really, it is only ever visible within a task and only used to indicate to the service which container (in case of multiple) in the task the load balancer it should send traffic to. That being said this should also have its mavis- prefix removed

terraform/app/modules/ecs_service/variables.tf
```diff
@@ -0,0 +1,79 @@
+variable "environment" {
+  type        = string
+  description = "Application environment (for example production or staging)"
+  nullable    = false
+}
+
+variable "server_type" {
+  type        = string
+  description = "Type of server to be deployed. This is set as an environment variable in the main container, and is used to determine how the application is launched"
```
bogsi17, line 9: Can we pass the server_type as part of the `task_config` environment?

bogsi17: As discussed, we won't do it since the server type is also referenced in resource names

PR #3322, opened 2025-04-04
Create modularized ecs services

TheOneFromNorway:
Create service to handle background tasks as well as create a user-facing service (thrust). This is also used as an opportunity to rename the existing service and move it into the module setup.
This is part of a PR chain with https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3321 as parent

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,7 +80,9 @@ USER 1000:1000
 # Entrypoint prepares the database.
 ENTRYPOINT ["/rails/bin/docker-entrypoint"]
 
-# Start server via Thruster by default, this can be overwritten at runtime
+# Start web server by default, this can be overwritten by environment variable
 EXPOSE 4000
 ENV HTTP_PORT=4000
-CMD ["./bin/thrust", "./bin/rails", "server"]
+ENV GOOD_JOB_PROBE_PORT=4000
+ENV SERVER_TYPE=web
+CMD ["./bin/docker-start"]
--- a/bin/docker-start
+++ b/bin/docker-start
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+if [ "$SERVER_TYPE" == "web" ]; then
+  echo "Starting web server..."
+  exec "$SCRIPT_DIR"/thrust "$SCRIPT_DIR"/rails server
+elif [ "$SERVER_TYPE" == "good-job" ]; then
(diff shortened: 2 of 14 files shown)


Dockerfile
```diff
@@ -80,7 +80,9 @@ USER 1000:1000
 # Entrypoint prepares the database.
 ENTRYPOINT ["/rails/bin/docker-entrypoint"]
 
-# Start server via Thruster by default, this can be overwritten at runtime
+# Start server via Thruster by default, this can be overwritten by environment variable
 EXPOSE 4000
 ENV HTTP_PORT=4000
-CMD ["./bin/thrust", "./bin/rails", "server"]
+ENV GOOD_JOB_PROBE_PORT=4000
+ENV SERVER_TYPE=thrust
```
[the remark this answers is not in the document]
TheOneFromNorway: If we want to go the route of using "good-job" to describe the backend service I would suggest we follow the same strategy for the user-facing service (also because we may have more user-facing service types in the future).

terraform/app/codedeploy.tf
```diff
@@ -116,8 +116,8 @@ resource "aws_s3_object" "appspec_object" {
   key    = "appspec.yaml"
   acl    = "private"
   content = templatefile("templates/appspec.yaml.tpl", {
-    task_definition_arn = aws_ecs_task_definition.task_definition.arn
-    container_name      = jsondecode(aws_ecs_task_definition.task_definition.container_definitions)[0].name
+    task_definition_arn = module.web_service.task_definition.arn
+    container_name      = module.web_service.task_definition.arn
```
bogsi17, line 120: If the variables are the same, can we remove the `container_name` from the template?

bogsi17: It's changed again in the following PR https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3323

bogsi17 (approved):

bin/docker-start
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+if [ "$SERVER_TYPE" == "web" ]; then
+  echo "Starting web server..."
+  exec "$SCRIPT_DIR"/thrust "$SCRIPT_DIR"/rails server
+elif [ "$SERVER_TYPE" == "good-job" ]; then
+  echo "Starting background server..."
```
thomasleese, line 9:
```suggestion
  echo "Starting good-job..."
```

thomasleese (approved):

PR #3323, opened 2025-04-04
Update deployment pipeline to work with modularized services

TheOneFromNorway:
- Option to deploy all or only one type of services
  - When deploying all deploy in parallel to speed up process
- Default to deploy all for deploying mavis
- Always deploy all for continuous deployment

This is part of a PR chain with https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3322 as parent

### DIFF
--- a/.github/workflows/deploy-application.yml
+++ b/.github/workflows/deploy-application.yml
@@ -20,6 +20,15 @@ on:
         description: Docker image tag
         required: false
         type: string
+      server_types:
+        description: Server types to deploy
+        required: true
+        type: choice
+        options:
+          - all
+          - web
+          - good-job
+        default: all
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-application.yml
```diff
@@ -157,3 +175,54 @@ jobs:
         run: |
           aws deploy wait deployment-successful --deployment-id $deployment_id
           echo "Deployment successful"
+
+  create-good-job-deployment:
```
thomasleese, line 179: How similar is this to `create-thrust-deployment`? Could it be pulled out in to a shared template and then called twice for each type of deployment?

bogsi17 (approved):

PR #3324, opened 2025-04-04
Use only today's prescreening data for VaccinateForm initialisation

murugapl:
This change updates the logic for initialising VaccinateForm so that it only considers PreScreening records with a SessionDate of Date.current. Only vaccine-independent questions will be copied over from previous pre-screenings, the others (and the notes field) will be blank. 
If no PreScreening from today's session is available, the form will remain blank.

Changes include:
• Implementing a data migration to associate PreScreening with a SessionDate.
• Updating the query logic to join on session_date and filter by today's date.
• Being selective about which fields of the PreScreening to copy over.
• Tests to verify that only today's screening data is used.

### DIFF
--- a/app/components/app_patient_page_component.rb
+++ b/app/components/app_patient_page_component.rb
@@ -57,16 +57,16 @@ class AppPatientPageComponent < ViewComponent::Base
 
   def default_vaccinate_form
     pre_screening =
-      patient_session.pre_screenings.order(created_at: :desc).first
+      patient_session
+        .pre_screenings
+        .joins(:session_date)
+        .merge(SessionDate.today)
+        .order(created_at: :desc)
+        .first
 
     VaccinateForm.new(
       feeling_well: pre_screening&.feeling_well,
-      knows_vaccination: pre_screening&.knows_vaccination,
-      no_allergies: pre_screening&.no_allergies,
-      not_already_had: pre_screening&.not_already_had,
-      not_pregnant: pre_screening&.not_pregnant,
-      not_taking_medication: pre_screening&.not_taking_medication,
-      pre_screening_notes: pre_screening&.notes || ""
+      not_pregnant: pre_screening&.not_pregnant
     )
   end
 end
--- a/app/forms/vaccinate_form.rb
+++ b/app/forms/vaccinate_form.rb
@@ -81,8 +81,17 @@ class VaccinateForm
         not_taking_medication: not_taking_medication || false,
         notes: pre_screening_notes,
         performed_by: current_user,
-        programme_id:
+        programme_id:,
+        session_date_id: session_date_id
       )
+  end
+
+  def session_date_id
+    patient_session
+      .session
+      .session_dates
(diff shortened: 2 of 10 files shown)


db/migrate/20250403155909_add_session_date_id_to_pre_screenings.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class AddSessionDateIdToPreScreenings < ActiveRecord::Migration[6.1]
+  def up
+    add_reference :pre_screenings, :session_date, foreign_key: true, null: true
+
+    PreScreening.reset_column_information
```
jhenderson, line 7: I don't think you need to call `reset_column_information` because you're using `update_column` which bypasses validations and callbacks, working directly with the database. `reset_column_information `.

app/forms/vaccinate_form.rb
```diff
@@ -82,10 +82,19 @@ def pre_screening
         notes: pre_screening_notes,
         patient_session:,
         performed_by: current_user,
-        programme_id:
+        programme_id:,
+        session_date_id: session_date_id
```
jhenderson, line 86:
```suggestion
        session_date_id:
```

app/forms/vaccinate_form.rb
```diff
@@ -82,10 +82,19 @@ def pre_screening
...
       )
   end
 
+  def session_date_id
+    patient_session
+      .session
+      .session_dates
+      .find { |sd| sd.value == Date.current }
```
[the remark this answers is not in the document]
thomasleese: Agree, I think another option would be to [use the `today` scope](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/688c6bab2308b87ff3ff5c35b7f7cc3e4aacfbfa/app/models/session_date.rb#L28):

```rb
session.session_dates.today.first
```

app/models/session_date.rb
```diff
@@ -23,6 +23,7 @@ class SessionDate < ApplicationRecord
   belongs_to :session
 
   has_many :session_attendances, dependent: :restrict_with_error
+  has_many :pre_screenings
```
thomasleese, line 26: I think we should add `restrict_with_error` to this as well, if a session date is deleted and there exists pre-screenings should we prevent the deletion of the session date.

app/components/app_patient_page_component.rb
```diff
@@ -57,16 +57,21 @@ def vaccination_records
...
+        .first
 
     VaccinateForm.new(
       feeling_well: pre_screening&.feeling_well,
-      knows_vaccination: pre_screening&.knows_vaccination,
-      no_allergies: pre_screening&.no_allergies,
-      not_already_had: pre_screening&.not_already_had,
+      knows_vaccination: nil,
+      no_allergies: nil,
+      not_already_had: nil,
```
thomasleese, line 71: We don't need to specify these as `nil` is the default.

thomasleese (approved):

PR #3326, opened 2025-04-05
Reorganise ImmunisationImportRow specs

thomasleese:
This reorganises the specs so that similar examples are next to each other, the hierarchy is correctly represented from the vaccination record, to the patient, etc.


thomasleese:
> LGTM? Sorry but this one's really hard to review 😵

Yeah sorry, if you remove hide whitespace changes you can at least see that the blocks of code are mostly the same but have just been moved around.

PR #3327, opened 2025-04-05
Add support for uploading SystmOne vaccination records

thomasleese:
This updates the import process to allow users to upload a file in the SystmOne format which has a different set of fields to the existing format. We haven't got a real example of this type of file for testing, so this is a best guess at how the format should be parsed.

We haven't updated any parts of the UI in this change, we might want to do this in a follow up PR to describe the two different formats.

To build this, the import mechanism required a substantial refactor, which is covered by:

- https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3312
- https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3317
- https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3318
- https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3325
- https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3326

### DIFF
--- a/app/models/immunisation_import_row.rb
+++ b/app/models/immunisation_import_row.rb
@@ -27,7 +27,7 @@ class ImmunisationImportRow
            :validate_session_id,
            :validate_time_of_vaccination,
            :validate_uuid,
-           :validate_vaccine_name
+           :validate_vaccine
 
   CARE_SETTING_SCHOOL = 1
   CARE_SETTING_COMMUNITY = 2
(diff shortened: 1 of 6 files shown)


app/models/immunisation_import_row.rb
```diff
@@ -253,7 +253,8 @@ def school
   end
 
   def programme
-    @programme ||= programmes_by_name[programme_name&.to_s]
+    @programme ||=
+      programmes_by_name[programme_name&.to_s] || vaccine&.programme
```
misaka: Does this change our spec for the imports? Just wondering if we've considered the wider implication of making what seems like a small improvement.

thomasleese: I'm not too worried about it because the priority is given to if the programme as provided explicitly in the files if it's present (we only rely on the vaccine as a fallback). I've also added a validation to check the vaccine is relevant for the programme if specified.

misaka (approved):

PR #3334, opened 2025-04-07
Remove obsolete Copilot-related code

TheOneFromNorway:
Now that copilot is no longer in use in any prod or non-prod environments it is time to remove copilot specific code


thomasleese (approved):
🗑️

PR #3336, opened 2025-04-07
Add mock data generators for testing

misaka:
The generators here were used for generating large volumes of data for importing into Mavis, and for automatically giving consents, performing triage and recording vaccinations for data in Mavis. This was used for the performance tests but can also be useful for other areas where test data is required. These generators include specs to try to ensure that they stay up-to-date with any code changes.

### DIFF
--- a/lib/generate/consents.rb
+++ b/lib/generate/consents.rb
@@ -1,0 +1,102 @@
+# frozen_string_literal: true
+
+module Generate
+  class Consents
+    attr_reader :organisation, :programme
+
+    def initialize(
+      organisation:,
+      programme: nil,
+      session: nil,
+      refused: 0,
+      given: 0,
+      given_needs_triage: 0
+    )
+      @organisation = organisation
+      @programme = programme || organisation.programmes.sample
+      @session = session
+      @refused = refused
+      @given = given
+      @given_needs_triage = given_needs_triage
+    end
+
+    def call
+      create_consent_with_response(:refused, @refused)
+      create_consent_with_response(:given, @given)
+      create_consent_given_needs_triage(@given_needs_triage)
+    end
+
+    def self.call(...) = new(...).call
+
+    private
+
+    def patients
+      (@session.presence || organisation)
+        .patients
+        .includes(:parents, :consents, consents: :parent)
+        .in_programmes([programme])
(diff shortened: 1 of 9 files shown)


lib/generate/patient_imports.rb
```diff
@@ -0,0 +1,193 @@
...
+      @programme = Programme.find_by(type: programme)
+      @urns =
+        urns ||
+          @organisation
+            .locations
+            .select { it.urn.present? }
+            .sample(3)
+            .pluck(:urn)
+      @patient_count = patient_count
+      @nhs_number_prefix = "99"
```
thomasleese, line 34: Do we need this?

lib/generate/consents.rb
```diff
@@ -0,0 +1,119 @@
...
+    def session_for(patient)
+      patient
+        .sessions
+        .eager_load(:location)
+        .merge(Location.school)
+        .has_programme(programme)
+        .sample
+    end
+
+    def create_consent_refused(count)
```
thomasleese, line 61: This method along with `create_consent_given` are very similar, I wonder if they could be refactored in to a single method?

lib/generate/triages.rb
```diff
@@ -0,0 +1,81 @@
...
+        .tap do
+          raise "Not enough patients to generate triages" if it.size < count
+        end
+    end
+
+    def user
+      @user ||= organisation.users.includes(:organisations).sample
+    end
+
+    def create_triage_ready_to_vaccinate(count)
```
thomasleese, line 51: Could the trait be passed in and avoid the need for two similar methods?

thomasleese (approved):
Just a couple of small comments, but I've approved anyway as this will be used for testing.

lib/generate/patient_imports.rb
misaka: Hrm. Ya, this was an old attempt to avoid collisions with existing patient data. I backed it out, but I don't remember why and can't find any hints in the reflog. ¯\_(ツ)_/¯

misaka:
There's a bunch of tests failing due to dates, I think it's because of a clash between BST and GMT. There are 7 failures with this message:

> Validation failed: Expiry Enter an expiry date after 8 April 2025

Will re-run these tests in the morning.

PR #3337, opened 2025-04-07
Normalise spaces during imports

alistair-white-horne-tng:
Implement and apply functions which normalise strings for all relevant fields of uploads to reduce match reviews on upload (eg addresses having double spaces, NHS numbers having zero-width joiners instead of spaces).

It both:
- normalises incoming string values
- normalises names when querying/matching patients in the database

[MAV-926](https://nhsd-jira.digital.nhs.uk/browse/MAV-926)


thomasleese:
This has been superseded by #3477.

PR #3349, opened 2025-04-08
Set `MAVIS__PDS__WAIT_BETWEEN_JOBS` environment variable

thomasleese:
This adds a new environment variable to the task definition that allows us to configure the wait time between when PDS jobs are enqueued without needing a deployment, by storing the value in the parameter store.

https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3329 is the change to the application code which added the setting.

### DIFF
--- a/terraform/app/ssm_parameters.tf
+++ b/terraform/app/ssm_parameters.tf
@@ -1,0 +1,12 @@
+resource "aws_ssm_parameter" "pds_wait_between_jobs" {
+  name = "/${var.environment}/pds_wait_between_jobs"
+  type = "String"
+
+  # This value is the default, but can be customised in the AWS console
+  # directly and isn't managed by Terraform.
+  value = "2"
+
+  lifecycle {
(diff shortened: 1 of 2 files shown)


terraform/app/ssm_parameters.tf
```diff
@@ -0,0 +1,12 @@
+resource "aws_ssm_parameter" "pds_wait_between_jobs" {
+  name = "/copilot/mavis/${var.environment}/pds_wait_between_jobs"
```
TheOneFromNorway, line 2: Lets not have /copilot in the string (this is just a remnant from the copilot days. I would like to remove it from the rails master key in the future as well)

thomasleese: I wondered about this too, I decided to keep it in to be consistent, but if you're happy then I'll remove it and we can base the others off this naming convention.

TheOneFromNorway (approved):

PR #3357, opened 2025-04-08
Don't send clinic invitations incorrectly

thomasleese:
This fixes a bug where patients who are eligible for multiple programmes were being sent clinic invitations where they shouldn't have been. The specific scenario is where a patient is vaccinated for one programme and consent refused for another, therefore the parents shouldn't receive an invitation to the clinic, but the code was incorrectly deciding that the patient should.

The code wasn't treating the programmes individually, so it was seeing one programme as unvaccinated and another programme as consent given and deciding therefore that the patient should be sent the invitation.

### DIFF
--- a/app/jobs/concerns/send_clinic_invitations_concern.rb
+++ b/app/jobs/concerns/send_clinic_invitations_concern.rb
@@ -19,17 +19,6 @@ module SendClinicInvitationsConcern
 
     return false unless patient.send_notifications?
 
-    eligible_programmes = patient_session.programmes & programmes
-
-    return false if eligible_programmes.empty?
-
-    all_vaccinated =
-      eligible_programmes.all? do |programme|
-        patient.vaccination_status(programme:).vaccinated?
-      end
-
-    return false if all_vaccinated
-
     already_sent_notification =
       patient_session.session_notifications.any? do
         _1.session_date == session_date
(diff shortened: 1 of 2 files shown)


jhenderson (approved):
LGTM

app/jobs/concerns/send_clinic_invitations_concern.rb
```diff
@@ -19,32 +19,24 @@ def should_send_notification?(patient_session:, programmes:, session_date:)
...
       eligible_programmes.all? do |programme|
-        patient.consent_status(programme:).refused?
+        patient.vaccination_status(programme:).vaccinated? ||
+          patient.consent_status(programme:).refused?
       end
 
-    return false if all_consent_refused
-
-    true
+    !all_vaccinated_or_consent_refused
```
jhenderson, line 39: I ran the spec without the negation operator and it still passed.

thomasleese:
@jhenderson Your comment seems to have disappeared, but good catch, I've amended the commit to make sure the tests are actually working as expected.

jhenderson:
> @jhenderson Your comment seems to have disappeared, but good catch, I've amended the commit to make sure the tests are actually working as expected.

I deleted it because it did fail when I ran the full spec, so I wasn't 100% sure, but yeah it would make sense for the specific test you wrote to fail if we changed the logic.

Also had this suggestion if you wanted to get rid of the `!all_vaccinated_or_consent_refused`:

```
eligible_programmes.any? do |programme|
  status = patient.vaccination_status(programme:)
  consent = patient.consent_status(programme:)

  !status.vaccinated? && !consent.refused?
end
```

misaka (approved):

PR #3361, opened 2025-04-09
Accept import capitalisation differences

alistair-white-horne-tng:
Implement some extra logic when importing, which:
- automatically accepts any capitalisation differences in relevant import fields, without staging the changes for review
- matches patients whose postcode only differs by capitalisation
- saves all incoming parental information to the existing matched parent

[MAV-926](https://nhsd-jira.digital.nhs.uk/browse/MAV-926)

### DIFF
--- a/app/models/concerns/pending_changes_concern.rb
+++ b/app/models/concerns/pending_changes_concern.rb
@@ -6,15 +6,31 @@ module PendingChangesConcern
   included { attribute :pending_changes, :jsonb, default: {} }
 
   def stage_changes(attributes)
+    need_save = false
+
     new_pending_changes =
       attributes.each_with_object({}) do |(attr, new_value), staged_changes|
         current_value = public_send(attr)
-        staged_changes[attr.to_s] = new_value if new_value != current_value
+
+        if normalise_for_comparison(new_value) ==
+             normalise_for_comparison(current_value)
+          if new_value != current_value
+            public_send("#{attr}=", new_value)
+            pending_changes.delete(attr.to_s)
+            need_save = true
+          end
+        else
+          staged_changes[attr.to_s] = new_value
+        end
       end
 
-    if new_pending_changes.any?
+    if new_pending_changes.any? || need_save
       update!(pending_changes: pending_changes.merge(new_pending_changes))
     end
+  end
+
+  def normalise_for_comparison(value)
+    value.respond_to?(:downcase) ? value.downcase : value
   end
 
   def with_pending_changes
--- a/spec/models/class_import_row_spec.rb
+++ b/spec/models/class_import_row_spec.rb
@@ -177,6 +177,22 @@ describe ClassImportRow do
         expect(parents.first.full_name).to eq("John Smith")
       end
     end
+
+    context "when uploading different caps name" do
+      let!(:existing_parent) do
+        create(:parent, full_name: "JENNY SMITH", email: "jenny@example.com")
+      end
+
+      let(:capitalised_parent_2_data) do
+        {
+          "PARENT_2_EMAIL" => "jenny@example.com",
+          "PARENT_2_NAME" => "Jenny Smith"
+        }
+      end
+      let(:data) { valid_data.merge(capitalised_parent_2_data) }
+
+      it { should include(existing_parent) }
+    end
   end
 
   describe "#to_patient" do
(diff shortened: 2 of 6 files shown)


app/models/concerns/pending_changes_concern.rb
```diff
@@ -9,14 +9,28 @@ def stage_changes(attributes)
...
+        end
       end
 
     if new_pending_changes.any?
       update!(pending_changes: pending_changes.merge(new_pending_changes))
     end
   end
 
+  def normalise_for_comparison(value)
+    # Normalise by case
```
jhenderson, line 30: Can remove the comment as the method name is self explanatory

app/models/patient_import_row.rb
```diff
@@ -122,6 +123,8 @@ def to_parents
       parent.phone = attributes[:phone] if attributes[:phone]
       parent.phone_receive_updates = false if parent.phone.blank?
 
+      parent.save!
```
jhenderson, line 126: `parent.save!` in a `to_parents` method does seem out of place, as this method should ideally just return parent objects without persisting them to the database. Should we persist later on down the chain? @thomasleese

thomasleese: Agreed, we should find a way to avoid this as it'll be a performance hit on the import. Currently we commits parents to the database in bulk.

alistair-white-horne-tng: Good point. On further investigation, this was already the functionality. I've added a few tests to check this.

app/models/concerns/pending_changes_concern.rb
```diff
@@ -9,14 +9,24 @@ def stage_changes(attributes)
...
+        if normalise_for_comparison(new_value) ==
+             normalise_for_comparison(current_value)
+          public_send("#{attr}=", new_value) if new_value != current_value
+        else
+          staged_changes[attr.to_s] = new_value
+        end
       end
 
     if new_pending_changes.any?
       update!(pending_changes: pending_changes.merge(new_pending_changes))
```
thomasleese, line 22: I think there's a scenario where this method won't work correctly.

If all the values coming in are the same when ignoring capitalisation, `new_pending_changes` will be left empty and therefore the `pending_changes` column won't be set on the model instance.

alistair-white-horne-tng: Isn't that the expected behaviour? If there are no pending changes, then `pending_changes` should be blank... I must be missing something...?

thomasleese: `new_pending_changes` would be empty that's true, but then the conditional below it `if new_pending_changes.any?` would be `false`, meaning that we're not saving the model (since `update!` isn't called) and we're leaving `pending_changes` as whatever it used to be on the model rather than clearing it out. I think it's an edge case, but is still a possible scenario.

alistair-white-horne-tng: Do you mean that you think it should always update `pending_changes` based on the most recent import? Because that's not currently the logic, right?

thomasleese: I think the problem is that the model won't get saved to the database in this method, so we'll end up with the old capitalisation versions rather than taking on the new ones.

thomasleese: I've thought of another potential scenario that we might want to consider.

Let's say we have a patient with the name "thomas" and a pending change to make it "tomas". If we then get a new import in that changes it to "Thomas", this will be accepted automatically since it's a capitalisation change only, however the pending change to "tomas" will stay.

I'm wondering if we would want to clear out the pending change in this case since the more recent import is presumably the most accurate?

alistair-white-horne-tng: Oh! That makes more sense. How's this new logic?

alistair-white-horne-tng: I've now added `pending_changes.delete(attr.to_s)` to solve that.

app/models/concerns/pending_changes_concern.rb
```diff
@@ -6,17 +6,33 @@ module PendingChangesConcern
   included { attribute :pending_changes, :jsonb, default: {} }
 
   def stage_changes(attributes)
```
thomasleese, line 8: I think there's potential to refactor this method as I'm finding it a bit complicated/hard to read. I've got a suggestion below which I think works and I find easier to understand:

```rb
def stage_changes(attributes)
  attributes.each do |attr, new_value|
    current_value = public_send(attr)

    if normalised(new_value) == normalised(current_value)
      public_send("#{attr}=", new_value)
      pending_changes.delete(attr.to_s)
    else
      pending_changes[attr.to_s] = new_value
    end
  end

  save!
end

private

def normalised(value)
  value.respond_to?(:downcase) ? value.downcase : value
end
```

What do you think?

thomasleese:
Hmm it seems like merging `v2.2.2-wip` in to `main` closed this, I think because it's a fork. You'll probably want to re-open it @alistair-white-horne-tng.

alistair-white-horne-tng:
New PR for this change: https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3588

PR #3362, opened 2025-04-09
Deploy and build scripts for Terraformlandia

misaka:
These are updated to work with our new Terraform infrastructure. I've used these scripts already so I think they could be ready for merging, but I've left this PR in draft because I'd like input from others on what may be missing and generally how they should work.


thomasleese (commented):
The scripts themselves look fine, I wonder if these are better placed in `script` rather than `bin`. I think the approach we vaguely wanted to take with `bin` was to stick with the Rails default scripts as much as possible.

misaka:
> The scripts themselves look fine, I wonder if these are better placed in `script` rather than `bin`. I think the approach we vaguely wanted to take with `bin` was to stick with the Rails default scripts as much as possible.

Does Rails have any guidance about when to use `script` or `bin`? I'm easy about where it goes, just wondering if there's a thing about maybe `script` is for things that run within the app environment, e.g. with `rails runner`.

thomasleese:
> Does Rails have any guidance about when to use `script` or `bin`? I'm easy about where it goes, just wondering if there's a thing about maybe `script` is for things that run within the app environment, e.g. with `rails runner`.

https://blog.saeloun.com/2025/02/10/rails-8-adds-script-folder-by-default-and-generator/ is the best article I can see about it. https://github.com/rails/rails/pull/52335/ is the PR that added it.

> Sometime developers use the bin/ folder but it is primarily for system-level scripts and placing project-specific scripts here often led to confusion.

I'm not really sure what system-level vs project-specific means here, but I would lean more towards these being project-specific scripts.

TheOneFromNorway (approved):

misaka:
> I'm not really sure what system-level vs project-specific means here, but I would lean more towards these being project-specific scripts.

Ya, fair. I think project-specificity makes sense as a differentiator.

thomasleese (approved):

PR #3364, opened 2025-04-09
Update number of background tasks

TheOneFromNorway:
- We want to ensure we have 2 tasks for availability reasons
- Make a variable that is easy to tune without config changes

### DIFF
--- a/terraform/app/env/copilotmigration.tfvars
+++ b/terraform/app/env/copilotmigration.tfvars
@@ -20,5 +20,5 @@ enable_splunk                   = false
 enable_cis2                     = false
 enable_pds_enqueue_bulk_updates = false
 
-minimum_replicas = 3
-appspec_bucket   = "nhse-mavis-appspec-bucket-copilotmigration"
+appspec_bucket        = "nhse-mavis-appspec-bucket-copilotmigration"
+minimnum_web_replicas = 2
--- a/terraform/app/variables.tf
+++ b/terraform/app/variables.tf
@@ -234,16 +234,22 @@ variable "enable_autoscaling" {
   description = "Boolean toggle to determine whether the ECS service should have autoscaling enabled."
 }
 
-variable "minimum_replicas" {
+variable "minimum_web_replicas" {
   type        = number
-  default     = 2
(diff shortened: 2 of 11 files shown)


terraform/app/variables.tf
```diff
@@ -234,16 +234,22 @@ variable "enable_autoscaling" {
...
   type        = number
   default     = 2
-  description = "Minimum amount of allowed replicas"
+  description = "Minimum amount of allowed replicas for web service"
 }
 
-variable "maximum_replicas" {
+variable "maximum_web_replicas" {
   type        = number
   default     = 2
```
thomasleese, line 245: Shouldn't we have 3 web instances?

misaka:
This PR also has changes for the web autoscaling ... do we want that?

terraform/app/variables.tf
TheOneFromNorway: I have 2 for playground environments (e.g. poc), the configuration in the .tfvars file sets it as 3 for all qa/test/.../prod environments

thomasleese: Ah, oops, should've checked there as well. I like the defaults being production-first and then non-production environments have it customised, I wonder if we can refactor that here?

misaka (approved):

terraform/app/variables.tf
TheOneFromNorway: Sure, reformatted to use default of 3

thomasleese (approved):

terraform/app/env/copilotmigration.tfvars
```diff
@@ -20,5 +20,6 @@ enable_splunk                   = false
 enable_cis2                     = false
 enable_pds_enqueue_bulk_updates = false
 
-minimum_replicas = 3
-appspec_bucket   = "nhse-mavis-appspec-bucket-copilotmigration"
+minimum_replicas      = 3
```
thomasleese, line 23: I think this shouldn't be here.

TheOneFromNorway: Right, removed minimum_replicas 👍

thomasleese (approved):

PR #3365, opened 2025-04-09
Log queries in dev reset endpoint

misaka:
Previously the dev reset endpoint wouldn't show any output and it was hard to understand what was happening. This changes it to output the query being performed and the time it took as a log message to give more feedback.

image


app/controllers/dev_controller.rb
```diff
@@ -20,61 +22,76 @@ def reset
...
   def random_consent_form
     Faker::Config.locale = "en-GB"
 
-    session = Session.includes(programmes: :vaccines).find(params[:session_id])
+    session =
+      if params[:slug].present?
+        Session.includes(programmes: :vaccines).find_by(slug: params[:slug])
+      else
+        Session.includes(programmes: :vaccines).find(slug: params[:session_id])
+      end
```
thomasleese, line 94: Do we need to include this change here?

misaka: Probably not. I just found it easier to use, do you think it's worth removing?

thomasleese (approved):

app/controllers/dev_controller.rb
thomasleese, line 1: I think it would make sense for this class to be a concern now that it has no controller methods of itself, overall I really like this refactor though.

thomasleese (approved):

app/controllers/dev_controller.rb
misaka: That's a good shout, and should be fast to do.

thomasleese (approved):

PR #3366, opened 2025-04-09
There was a typo in the pipeline

TheOneFromNorway:
- Causing good-job service not to update
  - Essentially we just re-deployed existing image multiple times


thomasleese (approved):
Do we know why the workflow didn't fail?

misaka (approved):

PR #3371, opened 2025-04-09
Update `script/shell.sh` for connecting to running servers.

misaka:
Addresses a couple issues:

- `getopt` on macos isn't compatible with GNU `getopt`
- minor stylistic things such as using lower-casing script variables and dasherising cmdline options

### DIFF
--- a/script/shell.sh
+++ b/script/shell.sh
@@ -21,23 +21,23 @@ if [ $# -lt 1 ]; then
     exit 1
 fi
 
-REGION="eu-west-2"
-SERVICE_NAME=""
-TASK_ID=""
-TASK_IP=""
+region="eu-west-2"
+service_name=""
+task_id=""
+task_ip=""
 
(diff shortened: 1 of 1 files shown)


tvararu (approved):

thomasleese (approved):

script/shell.sh
```diff
@@ -21,94 +21,84 @@ if [ $# -lt 1 ]; then
...
     case "$1" in
         --service)
-            SERVICE_NAME="$2"
+            service_name="$2"
             shift 2
             ;;
-        --task_id)
-            TASK_ID="$2"
+        --task-id)
+            echo "HERE"
```
TheOneFromNorway, line 36: Probably left over from debugging?

TheOneFromNorway (approved):

script/shell.sh
misaka: Yup. Good eyes!

PR #3378, opened 2025-04-10
Make max ACU capacity for rds a variable amount

TheOneFromNorway:
- Allows us to increase capacity of specific environments
  - e.g. production which handles larger/more tasks

### DIFF
--- a/terraform/app/variables.tf
+++ b/terraform/app/variables.tf
@@ -256,6 +256,12 @@ variable "minimum_good_job_replicas" {
   description = "Minimum amount of allowed replicas for good-job service"
 }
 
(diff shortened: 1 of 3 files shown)


terraform/app/variables.tf
```diff
@@ -256,6 +256,12 @@ variable "minimum_good_job_replicas" {
   description = "Minimum amount of allowed replicas for good-job service"
 }
 
+variable "max_acu_capacity" {
```
thomasleese, line 259:
```suggestion
variable "max_database_acu_capacity" {
```

Might be worth just making clear that this refers to the database.

thomasleese (approved):

PR #3379, opened 2025-04-10
Make `MAVIS__PDS__ENQUEUE_BULK_UPDATES` configurable

thomasleese:
This makes this variable configurable by adding it as a parameter and then exposing the variable as an environment variable.

This would then allow us to enable/disable this part of the service per environment while also being able to temporarily turn it on and off.

This will start by enabling it in all environments and then we can turn it off.

### DIFF
--- a/terraform/app/ssm_parameters.tf
+++ b/terraform/app/ssm_parameters.tf
@@ -11,6 +11,18 @@ resource "aws_ssm_parameter" "good_job_max_threads" {
   }
 }
 
+resource "aws_ssm_parameter" "pds_enqueue_bulk_jobs" {
+  name = "/${var.environment}/pds_enqueue_bulk_jobs"
+  type = "String"
+
+  # This value is the default, but can be customised in the AWS console
+  # directly and isn't managed by Terraform.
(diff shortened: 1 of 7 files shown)


terraform/app/ssm_parameters.tf
```diff
@@ -11,6 +11,18 @@ resource "aws_ssm_parameter" "good_job_max_threads" {
   }
 }
 
+resource "aws_ssm_parameter" "pds_enqueue_bulk_jobs" {
+  name = "/${var.environment}/pds_enqueue_bulk_jobs"
+  type = "String"
+
+  # This value is the default, but can be customised in the AWS console
+  # directly and isn't managed by Terraform.
+  value = "true"
```
TheOneFromNorway, line 20: We want to set it to true on all environments?

thomasleese: The default should be production-ready, and then we'll turn it off in the parameter store as necessary.

TheOneFromNorway: We could have kept the terraform variable and used it directly

PR #3380, opened 2025-04-10
Refactor use of ssm parameters

TheOneFromNorway:
- Ensure we use variable configs to be unique to an environment
  - For code to reflect deployed configurations
- One resource with for-each instead of multiple blocks (DRY)
- Remove lifecycle ignore_changes
  - Ensures regular sync between code and deployed configuration
  - Any persisted changes will end up being version controlled
- Disable PDS/CIS2/SPLUNK for POC environment

### DIFF
--- a/terraform/app/ssm_parameters.tf
+++ b/terraform/app/ssm_parameters.tf
@@ -1,38 +1,7 @@
-resource "aws_ssm_parameter" "good_job_max_threads" {
-  name = "/${var.environment}/good_job_max_threads"
-  type = "String"
+resource "aws_ssm_parameter" "environment_config" {
+  for_each = local.parameter_store_variables
+  name     = "/${var.environment}/env/${each.key}"
+  type     = "String"
 
-  # This value is the default, but can be customised in the AWS console
-  # directly and isn't managed by Terraform.
-  value = "5"
-
-  lifecycle {
-    ignore_changes = [value]
-  }
-}
-
-resource "aws_ssm_parameter" "pds_enqueue_bulk_jobs" {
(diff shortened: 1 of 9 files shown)


terraform/scripts/bootstrap.sh
```diff
@@ -106,8 +106,13 @@ http_hosts = {
...
   MAVIS__GIVE_OR_REFUSE_CONSENT_HOST = "$ENV.mavistesting.com"
 }
-minimum_replicas     = 3
-appspec_bucket       = "nhse-mavis-appspec-bucket-$ENV"
+
+enable_splunk                   = false
+enable_cis2                     = false
+enable_pds_enqueue_bulk_updates = false
+
+minimum_replicas = 3
```
thomasleese, line 114:
```suggestion
minimum_web_replicas = 3
```

terraform/app/ssm_parameters.tf
```diff
@@ -1,38 +1,7 @@
-resource "aws_ssm_parameter" "good_job_max_threads" {
-  name = "/${var.environment}/good_job_max_threads"
-  type = "String"
+resource "aws_ssm_parameter" "environment_config" {
```
thomasleese, line 1:
```suggestion
resource "aws_ssm_parameter" "environment_configs" {
```

Not sure what the convention is with Terraform, but when using `fore_each` would we normally pluralise the resource name?

TheOneFromNorway: But if there is a strong preference/existing trend in the repository for this behavior I am also not so fussed

thomasleese (approved):

PR #3388, opened 2025-04-11
Add Essex schools to Hertfordshire onboarding

thomasleese:
This adds the Essex school URNs to the Hertfordshire organisation via the onboarding configuration, however in production we won't be able to use this file as the organisation is already onboarded.

To support this, we've split the organisation in two separate teams which handles the different schools.

### DIFF
--- a/config/onboarding/hertfordshire-production.yaml
+++ b/config/onboarding/hertfordshire-production.yaml
@@ -1,24 +1,29 @@
 organisation:
-  name: Norfolk, Suffolk and Cambridgeshire and Peterborough Community and School Age Immunisation Service
-  email: hct.csaiscambspb@nhs.net
+  name: Hertfordshire and East Anglia Community School Age Immunisation Service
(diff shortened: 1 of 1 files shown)


config/onboarding/hertfordshire-production.yaml
```diff
@@ -30,9 +35,105 @@ schools:
     - 142902
     - 145051
     - 145271
+  essex:
+    - 115235
```
benilovj, line 39: I've just tried the first URN and it turns out it's a closed establishment on GIAS 🤦 
It does have a linked school – I'll see if I can clean this list up over the weekend.

benilovj: Addressed in [0733ad8](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3388/commits/0733ad879fcc8c285a6f9f4394ef04f17066e4c9)

benilovj (approved):

PR #3392, opened 2025-04-14
Add linting to script and ci

TheOneFromNorway:
- Ensure proper and consistent formatting for PRs


bin/lint
```diff
@@ -9,3 +9,4 @@ if [ -z "$*" ]
 then
    bin/brakeman --quiet --no-summary --no-pager
 fi
+terraform fmt -recursive
```
thomasleese, line 12: Currently this linting script allows passing file paths as arguments to only lint those files, is it possible to include this for Terraform formatting like the others?

thomasleese (approved):
Code looks good, could we reference specifically that this adds `terraform fmt` in the commit message?

PR #3393, opened 2025-04-14
Overwrite db/data/dfe-schools.zip

thomasleese:
When downloading the latest version of the schools data we should make sure to overwrite the version already contained in the repo otherwise running, for example, `schools:import` will use the file in the repo and not the latest version just downloaded.

### DIFF
--- a/lib/tasks/schools.rake
+++ b/lib/tasks/schools.rake
@@ -45,7 +45,8 @@ namespace :schools do
       download_button = download_form.button_with(value: "Results.zip")
       puts "'Results.zip' link found, downloading the file..."
       download_file = agent.click(download_button)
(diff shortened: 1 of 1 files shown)


misaka (approved):

lib/tasks/schools.rake
```diff
@@ -45,7 +45,7 @@ namespace :schools do
       download_button = download_form.button_with(value: "Results.zip")
       puts "'Results.zip' link found, downloading the file..."
       download_file = agent.click(download_button)
-      download_file.save("db/data/dfe-schools.zip")
```
misaka, line 48: Is it worth putsing "Overwriting ..."?

thomasleese: I've added a line about that to the task.

PR #3395, opened 2025-04-14
Include TFLint into the CI pipeline

TheOneFromNorway:
- Also update the bin/lint script to include this feature
  - Looping allows for only running tflint against passed file arguments

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -70,3 +70,16 @@ jobs:
           terraform_version: 1.10.5
       - name: Check formatting
         run: terraform fmt -check -recursive
+
+  tflint:
+    name: TFLint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: terraform-linters/setup-tflint@v4
+      - name: Run TFLint
+        working-directory: terraform
+        run: |
+          tflint --init
+          tflint --recursive
--- a/bin/lint
+++ b/bin/lint
@@ -11,3 +11,5 @@ if [ -z "$*" ]; then
 else
    terraform fmt $*
 fi
+tflint --chdir=terraform --init
+tflint --chdir=terraform --recursive
--- a/terraform/.tflint.hcl
+++ b/terraform/.tflint.hcl
@@ -1,0 +1,9 @@
+plugin "terraform" {
+  enabled = true
+}
+
+plugin "aws" {
+  enabled = true
+  version = "0.37.0"
+  source  = "github.com/terraform-linters/tflint-ruleset-aws"
+}
--- a/terraform/app/modules/dns/main.tf
+++ b/terraform/app/modules/dns/main.tf
@@ -1,3 +1,13 @@
+terraform {
+  required_version = "~> 1.10.5"
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 5.87"
+    }
+  }
+}
+
 data "aws_route53_zone" "this" {
   name         = var.zone_name
   private_zone = false
--- a/terraform/app/modules/dns/outputs.tf
+++ b/terraform/app/modules/dns/outputs.tf
@@ -1,3 +1,4 @@
 output "certificate_arn" {
-  value = aws_acm_certificate.this.arn
+  value       = aws_acm_certificate.this.arn
+  description = "ARN of the generated ACM certificate"
 }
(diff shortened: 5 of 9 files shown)


.github/workflows/lint.yml
```diff
@@ -70,3 +70,16 @@ jobs:
...
+  tflint:
+    name: TFLint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: terraform-linters/setup-tflint@v4
+      - name: Run TFLint
+        run: |
+          cd terraform
```
thomasleese, line 83: There's also the option of `working-directory: terraform` which I think is nicer than relying on `cd`.

.github/workflows/lint.yml
```diff
@@ -70,3 +70,16 @@ jobs:
...
+          tflint --init
+          tflint --config=$(pwd)/.tflint.hcl --recursive
```
thomasleese, line 85: According to the documentation for `tflint`, the default value for `--config` is `.tflint.hcl` so I don't think we need to pass this in.

bin/lint
```diff
@@ -8,6 +8,12 @@ bin/bundle exec rufo ${*:-app}
 if [ -z "$*" ]; then
    bin/brakeman --quiet --no-summary --no-pager
    terraform fmt -recursive
+   tflint --config="$(pwd)/terraform/.tflint.hcl" --chdir=terraform --recursive
```
thomasleese, line 11: Do we need to specify `--config` here? I think it will default to `.tflint.hcl` in the current working directory.

bin/lint
```diff
@@ -8,6 +8,12 @@ bin/bundle exec rufo ${*:-app}
...
 else
    terraform fmt $*
+   for file in "$@"; do
```
thomasleese, line 14: I realise that this is the way `tflint` works, but I'm not sure about this because it means you can't do, for example:

```sh
$ bin/lint terraform/app/modules
$ bin/lint terraform/app/modules/dns/*
```

I'd almost be tempted to show a warning message instead, it doesn't feel like `tflint` is setup to support this use case very well.

Out of interest, is this an area where pre-commit makes it easier to run the linting rules on just one file, one directory or one glob?

TheOneFromNorway: Pre-commit would handle this OOTB for all files by running with option `--all-files`, a list of files by passing `--files`, or the default run on staged files. The non-pre-commit-version I agree does not support our specific use-case very well.
I see 2 easy-win options:
1. Just run it for all files in the terraform directory regardless (it does not do any writes anyway so worst case it complains about a file you didnt touch)
2. Do not run it unless you specify running the linting scripts for all files. 

If we plan to move to pre-commit in the end then I guess not having a beautiful solution at this stage is less relevant

.github/workflows/lint.yml
TheOneFromNorway: Yeah right, lets do that 👍

.github/workflows/lint.yml
TheOneFromNorway: See comment below regarding the --config parameter

bin/lint
thomasleese: Wow that's really surprising to me. It kind of implies that the more natural way to configure `tflint` would be to have a `.tflint.hcl` file in each directory.

I would certainly expect to be able to do: `$ tflint --chdir=terraform/app/modules/dns` without needing to also specify the path to the configuration file.

How do you feeling about having `.tflint.hcl` in each directory?

thomasleese:
Can we delete https://github.com/nhsuk/manage-vaccinations-in-schools/blob/main/terraform/hooks/lint.sh ?

bin/lint
TheOneFromNorway: Yes I saw it mentioned having a `.tflint.hcl` for each directory containing terraform files. I would prefer not to do this since we have to add it for every module/etc and adding 10+ files seems excessive. You can run `tflint --chdir=terraform/app/modules/dns`, it just will not pick up any unique configuration and so run with default config only, see [ref](https://github.com/terraform-linters/tflint/discussions/1872).

Ultimately, not specifying a config file may be fine anwyays. We install the aws plugin by running tflint --init in the terraform root directory anwyay, so it should therefore it is already part of the checks AFAIK. This would just mean that we loose the ability to add custom rules/behavior but we are not doing that at this stage anyway. I will refactor this out for simplicity at this stage.

bin/lint
```diff
@@ -11,3 +11,5 @@ if [ -z "$*" ]; then
 else
    terraform fmt $*
 fi
+tflint --init --config=terraform/.tflint.hcl
```
thomasleese, line 14:
```suggestion
tflint --chdir=terraform --init
```

This might be nicer as it matches the line below.

thomasleese (approved):

thomasleese (approved):

PR #3407, opened 2025-04-15
Prepare restriction of AWS permissions for github workflows

bogsi17:
We want to restrict the IAM role used by the Github workflows such that it can only be assumed by workflow runs from the main and release branches. This causes the problem that deployment workflows run from tags could not assume the role anymore.

To mitigate this issue and still allow deploying specific tags, this PR allows to specify the git tag that shall be deployed in a dedicated field. While the workflow can be run from main or release, this still allows to deploy a specific git tag. If no value is set, it defaults to the regular behaviour.

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -51,6 +51,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.image_tag || github.sha }}
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
@@ -110,6 +112,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.image_tag || github.sha }}
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy-infrastructure.yml
```diff
@@ -22,6 +25,9 @@ on:
         type: string
   workflow_call:
     inputs:
+      git-sha:
+        description: The git commit sha of the commit to deploy.
+        type: string
```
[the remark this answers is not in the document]
bogsi17: After some changes on the base branch, the input is now called `image_tag`. Ultimately this input serves two purposes: 
- To determine which docker image to deploy
- To checkout the git ref from which the docker image was built (In order to have matching terraform configuration files)

We rely a bit on the convention that the docker image tag can be interpreted as git ref but I think it makes sense to tie both things together. If we want a way to deploy an image A using the terraform config from commit B then we would need two separate inputs

.github/workflows/deploy.yml
```diff
@@ -36,15 +36,35 @@ on:
           - web
           - good-job
         default: all
+      git_tag:
+        description: Git tag to deploy (optional). Will use workflow ref if not provided.
```
misaka, line 40: For clarity, are we able to insert the ref name here?

```suggestion
        description: Git tag to deploy (optional). Will use the reference ${{ github.ref_name}} if not provided.
```

misaka: Or maybe use the `default` option:

```suggestion
        description: Git tag to deploy (optional). Will use workflow ref if not provided.
        default: ${{ github.ref_name }}
```

README.md
```diff
@@ -104,7 +110,7 @@ git checkout release
 git pull origin release
 git reset --hard v1.0.0-rc1
 git push --tags origin release
-```
+````
```
misaka, line 113: Is this closing ````yaml? Should ```sha be closed anywhere?

misaka (approved):

misaka (approved):
Looks good! We agreed we'd leave the new CodeQL warning alone for now, and address in a later PR. 👍

PR #3409, opened 2025-04-15
Accumulate gender on import

alistair-white-horne-tng:
During an import, if the existing patient's gender is `nil`, `not_known` or `not_specified`, and the incoming value is different, then Mavis will now automatically accept the new value without staging it and asking for a review.

### DIFF
--- a/app/models/patient_import_row.rb
+++ b/app/models/patient_import_row.rb
@@ -47,6 +47,12 @@ class PatientImportRow
         existing_patient.registration = attributes.delete(:registration)
       end
 
+      if attributes[:gender_code].in?(%w[male female not_specified]) &&
+           !existing_patient.gender_code.in?(%w[male female not_specified]) &&
+           attributes[:gender_code] != existing_patient.gender_code
+        existing_patient.gender_code = attributes.delete(:gender_code)
+      end
+
       if address_postcode.present? &&
            address_postcode.to_postcode != existing_patient.address_postcode
         attributes.merge!(
--- a/spec/models/class_import_row_spec.rb
+++ b/spec/models/class_import_row_spec.rb
@@ -216,6 +216,62 @@ describe ClassImportRow do
         end
       end
     end
+
+    context "with an existing patient without gender" do
+      let(:data) { valid_data.merge("CHILD_GENDER" => "male") }
+
+      let!(:existing_patient) do
+        create(
+          :patient,
+          address_postcode: "SW1A 1AA",
+          family_name: "Smith",
+          gender_code: "not_known",
+          given_name: "Jimmy",
+          date_of_birth: Date.new(2010, 1, 1)
+        )
+      end
+
+      it { should eq(existing_patient) }
+
+      it "saves the incoming gender" do
+        expect(patient).to have_attributes(gender_code: "male")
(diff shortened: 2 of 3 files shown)


app/models/concerns/pending_changes_concern.rb
```diff
@@ -9,14 +9,24 @@ def stage_changes(attributes)
     new_pending_changes =
       attributes.each_with_object({}) do |(attr, new_value), staged_changes|
         current_value = public_send(attr)
-        staged_changes[attr.to_s] = new_value if new_value != current_value
+
+        if attr == :gender_code && overwrite_gender?(current_value)
+          public_send("#{attr}=", new_value) if new_value != current_value
+        elsif new_value != current_value
+          staged_changes[attr.to_s] = new_value
+        end
```
thomasleese, line 17: The logic looks good, but I don't think it belongs in the `PendingChangesConcern` class. The class is designed to be generic enough that it can applied to any model, at the moment it's only being used by `Patient` and `VaccinationRecord`, but that doesn't mean it will always be the case.

The gender is specific to the patient, and therefore it may not be appropriate to add logic to other classes. I would suggest instead that we apply this logic where the `stage_changes` method is being used.

We already have similar logic related to registration, where it's set automatically in class list imports but not in cohort imports. I think we can probably follow a similar approach for gender: https://github.com/nhsuk/manage-vaccinations-in-schools/blob/3d6d91783a30beb355f1bd8d00b674faaf4455f6/app/models/patient_import_row.rb#L46-L48

alistair-white-horne-tng: That makes sense. Changed that now

app/models/patient_import_row.rb
```diff
@@ -145,6 +150,10 @@ def to_parent_relationships(parents, patient)
       end
   end
 
+  def is_gender_known?(gender_code)
+    gender_code.in?(%w[male female])
```
misaka, line 154: Is there a reason we only check this against `male` and `female`? Our patients model support `not_specified` ...

alistair-white-horne-tng: I was under the impression that `male` or `female` should overwrite `not_specified` - is that incorrect?

I suppose I thought that `not_specified` indicated that the person didn't choose an option, rather than actively chose to keep their gender out of Mavis.

alistair-white-horne-tng: Okiedoke, because we're unsure, I'll leave it for manual review, and not get Mavis to auto-accept if `not_specified` is already there.

thomasleese (approved):
The code looks good, I would suggest the commit messages can be swashed together as I don't think we need the level of granularity in the commit history.

misaka:
> The code looks good, I would suggest the commit messages can be swashed together as I don't think we need the level of granularity in the commit history.

➕1 for swashing commit messages! 😀

app/models/patient_import_row.rb
misaka: Might be worth adding a comment about this behaviour to the Jira ticket in the testing section to ask testers to have a look and see what they think.

misaka (approved):

PR #3411, opened 2025-04-15
Don't require VACCINATED column

thomasleese:
If the file which has been uploaded has a `VACCINE_GIVEN` or `Vaccination type` column, we can assume that the record has been administered.

This was already the case, but only the vaccine that the user provided is valid. If the vaccine was invalid, this lead to a confusing scenario where the user would see a validation error about the vaccine being invalid and a validation error about missing `VACCINATED`. This is wrong, once the vaccine has been fixed, the second validation error about the `VACCINATED` column would go away, so there's no need to show it to the user in the first place.

### DIFF
--- a/app/models/immunisation_import_row.rb
+++ b/app/models/immunisation_import_row.rb
@@ -377,7 +377,8 @@ class ImmunisationImportRow
       elsif "no".start_with?(vaccinated.to_s.downcase)
         false
       end
-    elsif vaccine.present?
+    elsif vaccine_name.present? ||
+          combined_vaccination_and_dose_sequence.present?
       true
     end
   end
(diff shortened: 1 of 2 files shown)


app/models/immunisation_import_row.rb
```diff
@@ -377,7 +377,8 @@ def administered
       elsif "no".start_with?(vaccinated.to_s.downcase)
```
TheOneFromNorway, line 377: Should we do a separate error if the field is not "yes" or "no"?

thomasleese: This scenario is already covered by the validation error message explaining that the value needs to be either `Y` or `N`:

https://github.com/nhsuk/manage-vaccinations-in-schools/blob/3d6d91783a30beb355f1bd8d00b674faaf4455f6/app/models/immunisation_import_row.rb#L461

TheOneFromNorway (approved):

PR #3414, opened 2025-04-15
Accumulate address on import

alistair-white-horne-tng:
During an import, if all the existing patient's address fields are `nil`, then Mavis will now automatically accept the full incoming address without staging it and asking for a review.

### DIFF
--- a/app/models/patient_import_row.rb
+++ b/app/models/patient_import_row.rb
@@ -60,6 +60,10 @@ class PatientImportRow
           address_line_2: address_line_2&.to_s,
           address_town: address_town&.to_s
         )
+      elsif auto_overwrite_address?(existing_patient)
+        existing_patient.address_line_1 = attributes.delete(:address_line_1)
+        existing_patient.address_line_2 = attributes.delete(:address_line_2)
+        existing_patient.address_town = attributes.delete(:address_town)
       end
 
       existing_patient.stage_changes(attributes)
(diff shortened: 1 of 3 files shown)


app/models/patient_import_row.rb
```diff
@@ -47,6 +47,13 @@ def to_patient
         existing_patient.registration = attributes.delete(:registration)
       end
 
+      unless existing_patient.has_address?
+        existing_patient.address_line_1 = attributes.delete(:address_line_1)
+        existing_patient.address_line_2 = attributes.delete(:address_line_2)
+        existing_patient.address_town = attributes.delete(:address_town)
+        existing_patient.address_postcode = attributes.delete(:address_postcode)
+      end
```
thomasleese, line 55: I'm wondering about this approach. We know that in some files only the postcode is included, rather than the full address. If only the postcode is included (and matches the existing patients), is there a risk that we will wipe out other parts of the address that might be correct?

alistair-white-horne-tng: I've changed the logic slightly - what do you think now?

app/models/patient_import_row.rb
```diff
@@ -195,6 +201,13 @@ def nhs_number_value
 
   private
 
+  def auto_overwrite_address?(existing_patient)
+    existing_patient.address_postcode == attributes[:address_postcode].to_postcode &&
```
thomasleese, line 205:
```suggestion
    existing_patient.address_postcode == address_postcode.to_postcode &&
```

thomasleese (approved):

thomasleese (approved):

PR #3418, opened 2025-04-16
Remove duplicated permissions from custom policy

bogsi17:
The custom IAM policy used by the GitHub workflows contains very fine-granular permissions. Whenever a new resource is added, it is quite cumbersome to find out all the detailed permissions that are required to manage this resource.

For this reason, two AWS-managed policies were added to the role that grant generic read-only permission as well as permissions to tag and untag resources. Some fine-granular permissions in the custom policy are now no longer needed.


docs/disaster-recovery.md
```diff
@@ -195,3 +195,14 @@ RAILS_ENV=staging bin/bundle exec \
...
 ```
+
+## Set up a new AWS account from scratch
+
+### Create a new IAM role for GitHub workflows
+
+In the AWS IAM console, create a new role for the GitHub workflows. Create a custom policy from `terraform/resources/github_actions_policy.json`. Also, attach the managed policies
+
+- `ReadOnlyAccess`
+- `ResourceGroupsTaggingAPITagUntagSupportedResources`
```
TheOneFromNorway, line 206: Can we change this to be a cli-command or script? What about creating the federated identity?

TheOneFromNorway: Fair enough, for now lets ensure we document this technical debt somewhere

TheOneFromNorway (approved):

PR #3423, opened 2025-04-16
Rename poc environments

bogsi17:
The names `poc` and `copilotmigration` for the two environments are outdated. To better reflect their purpose, this PR renames them to `sandbox-alpha` and `sandbox-beta`.


terraform/app/env/sandbox-beta.tfvars
```diff
@@ -0,0 +1,24 @@
...
+  MAVIS__HOST                        = "sandbox-beta.mavistesting.com"
+  MAVIS__GIVE_OR_REFUSE_CONSENT_HOST = "sandbox-beta.mavistesting.com"
+}
+
+enable_splunk                   = false
+enable_cis2                     = false
+enable_pds_enqueue_bulk_updates = false
+
+minimum_web_replicas = 2
+appspec_bucket       = "nhse-mavis-appspec-bucket-sandbox-beta"
```
TheOneFromNorway, line 24: Can we change the order so that variables appear in same order consistently between the generated .tfvars files?

TheOneFromNorway (approved):

thomasleese (commented):
Do we need two of these? I'm thinking we could get rid of one of them.

bogsi17:
> Do we need two of these? I'm thinking we could get rid of one of them.

@thomasleese We don't need both environments all the time, but sometimes it has been useful to have two if we were working on different topics at the same time.

PR #3424, opened 2025-04-17
Post release of v2.1.3 remove old service

bogsi17:
- The old service is replaced by identical service with new name
  - This used the module and simplifies setup/running/developing in the future


thomasleese:
This was replaced by #3427

PR #3426, opened 2025-04-17
Use last available answer from today for initialising pre-screening questions

murugapl:
Update pre-screening initialisation logic to retrieve the most recent answer for each question from the same day, instead of only using the last pre-screening performed. This ensures that relevant answers are translated across pre-screenings, even if a specific vaccine does not have a particular question.

### DIFF
--- a/app/components/app_patient_page_component.rb
+++ b/app/components/app_patient_page_component.rb
@@ -56,19 +56,21 @@ class AppPatientPageComponent < ViewComponent::Base
   end
 
   def default_vaccinate_form
-    pre_screening =
+    today_pre_screenings =
       patient_session
         .pre_screenings
         .joins(:session_date)
         .merge(SessionDate.today)
         .order(created_at: :desc)
-        .first
+
+    feeling_well = today_pre_screenings.any?(&:feeling_well) || nil
+    not_pregnant = today_pre_screenings.any?(&:not_pregnant) || nil
 
     VaccinateForm.new(
       patient_session:,
       programme:,
-      feeling_well: pre_screening&.feeling_well,
-      not_pregnant: pre_screening&.not_pregnant
+      feeling_well:,
+      not_pregnant:
     )
   end
 end


app/components/app_patient_page_component.rb
```diff
@@ -56,19 +56,27 @@ def vaccination_records
...
         .pre_screenings
         .joins(:session_date)
         .merge(SessionDate.today)
         .order(created_at: :desc)
-        .first
+
+    feeling_well =
+      today_pre_screenings
+        .detect { |ps| ps.feeling_well.present? }
+        &.feeling_well
```
thomasleese, line 69:
```suggestion
    feeling_well = today_pre_screenings.any?(&:feeling_well)
```

I'm struggling a bit to understand this logic, but I think this suggestion has the same behaviour.

murugapl: Oh much easier!

murugapl: If nothing is present, then it sets the value for "feeling_well" and "not_pregnant" to false rather than nil.. I believe that is incorrect?

murugapl: Yes thanks, silly me!

app/components/app_patient_page_component.rb
```diff
@@ -56,19 +56,21 @@ def vaccination_records
...
+    feeling_well = today_pre_screenings.any?(&:feeling_well) || nil
+    not_pregnant = today_pre_screenings.any?(&:not_pregnant) || nil
 
     VaccinateForm.new(
       patient_session:,
       programme:,
-      feeling_well: pre_screening&.feeling_well,
-      not_pregnant: pre_screening&.not_pregnant
+      feeling_well: feeling_well,
+      not_pregnant: not_pregnant
```
thomasleese, line 73:
```suggestion
      feeling_well:,
      not_pregnant:
```

PR #3440, opened 2025-04-23
Fix consent notification count in programme overview

murugapl:
This fix ensures that only notifications from the user's selected organisation appear in the programme overview statistics, by introducing a ConsentNotificationPolicy.
In addition, programme overview cards have been extracted into a dedicated AppProgrammeStatsComponent, allowing for more isolated tests in the future.


app/policies/consent_notification_policy.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ConsentNotificationPolicy < ApplicationPolicy
+  class Scope < ApplicationPolicy::Scope
+    def resolve
+      scope.joins(:session).where(
+        session: {
+          organisation: user.selected_organisation
+        }
+      )
```
thomasleese, line 10: One possible improvement could be to define a `has_one :organisation, through: :session` in the `ConsentNotification` model, which would then allow it to be queried like `ConsentNotification.where(organisation: user.selected_organisation)` which might read better in this class.

app/components/app_programme_stats_component.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AppProgrammeStatsComponent < ViewComponent::Base
+  def initialize(programme:, patients_count:)
+    super
+    @programme = programme
+    @patients_count = patients_count
```
thomasleese, line 7: For consistency, could we get the patients count in the same way as we do for vaccinations and consent notifications? I quite like the idea of this component not needing any external data and then it handles everything itself.

thomasleese (commented):
Nice refactor, just left a couple of comments.

app/policies/consent_notification_policy.rb
murugapl: Similar to the approach in location_policy.rb, I think the explicit join on the session is needed even if I add the organisation to the ConsentNotification model? I get a "missing FROM clause" error otherwise. Or are you suggesting I use this instead of the policy entirely?

thomasleese: Hmm. I was expecting that Rails would sort that out automatically if `has_one :organisation, through: :session` was defined on the model, but maybe the explicit join is still required.

It's not a blocker for this change, so if it's not easy to do then I wouldn't worry about it.

murugapl: I've added the organisation to the consent notification model, but am unable to get it to work without joining explicitly..

thomasleese (approved):

PR #3465, opened 2025-04-29
Remove brakeman from pre-commit and fail on changed files

TheOneFromNorway:
- Adding a custom check to ensure we do not initiate commit when running linters in fix mode
- Removing brakeman as it runs on all files and takes too long as a pre-commit step

### DIFF
--- a/hk.pkl
+++ b/hk.pkl
@@ -35,13 +35,29 @@ local linters = new Mapping<String, Group> {
                 }
         }
     }
+    local steps_with_staged_fixes = (linters["all"].steps) {
+        for (key, value in linters["all"].steps) {
+            [key] = (value) {
+                stage = List("*")
+            }
+        }
+    }
+    ["pre-commit"] = new Group {
+        steps = steps_with_staged_fixes.toMap().remove("brakeman").toMapping()
+    }
 }
 
 hooks {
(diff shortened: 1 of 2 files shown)


hk.pkl
```diff
@@ -35,13 +35,29 @@ local linters = new Mapping<String, Group> {
                 }
         }
     }
+    local steps_with_staged_fixes = (linters["all"].steps) {
```
thomasleese, line 38: Not being very familiar with Pkl, I'm struggling to understand why we need this and can't do something like, for example:

```pkl
linters["all"].steps.toMap().remove("brakeman").toMapping()
```

thomasleese (approved):

hk.pkl
TheOneFromNorway: We do this as well, but we need to add the `stage = List("*")` to the elements, as otherwise the changes made by fix are not added to the git index and therefore not included in the commit. I assumed we did not want the changes staged if we ran it manually so therefore this separate step to add them only for pre-commit

TheOneFromNorway: See line 46 where we afterwards remove brakeman (we could also have removed brakeman first but this way the `steps_with_staged_fixes` local is more versatile for future development)

PR #3467, opened 2025-04-29
Add step to check if rails tests are needed

TheOneFromNorway:
- Simply check if changes are confined within non-ruby/rails relevant directories
- Exit with 0 to have job succeed but skip following steps

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,9 +5,34 @@ on:
   workflow_call:
 
 jobs:
+  pre-rails-tests:
+    name: Pre Rails Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Determine if rails tests are needed
+        id: check-rails-relevant-changes
+        run: |
+          if [ ${{ github.event_name }} == "pull_request" ]; then
+            git fetch origin ${{ github.event.pull_request.head.ref }}
+            RELEVANT_DIRS=$(git diff ${{ github.event.pull_request.base.sha }} origin/${{ github.event.pull_request.head.ref }}  --name-only)
+            echo "Relevant changes for rails: $RELEVANT_DIRS"
+            if [ -z "$(echo "$RELEVANT_DIRS" | egrep -v "^(.github|adr|aws|docs|storage|terraform|tests|tests-examples)/")" ]; then
+              echo "No relevant changes detected, skipping rails tests"
+              echo "RUN_PRE_RAILS_TESTS=false" >> $GITHUB_OUTPUT
+              exit 0
+            fi
+          fi
+          echo "Not a pull request or detected relevant changes: Running rails tests"
+          echo "RUN_PRE_RAILS_TESTS=true" >> $GITHUB_OUTPUT
+    outputs:
+      run-pre-rails-tests: ${{ steps.check-rails-relevant-changes.outputs.RUN_PRE_RAILS_TESTS }}
+
   rails:
     name: Rails
     runs-on: ubuntu-latest
+    needs: pre-rails-tests
+    if: needs.pre-rails-tests.outputs.run-pre-rails-tests == 'true'
     services:
       postgres:
         image: postgres:17.2


.github/workflows/test.yml
```diff
@@ -25,6 +25,18 @@ jobs:
...
+      - name: Determine if rails test are needed
+        run: |
+          if [ ${{ github.event_name }} == "pull_request" ]; then
+            RELEVANT_DIRS=$(git diff ${{ github.event.pull_request.base.sha }} HEAD --name-only \
+            | egrep -v "^(.github|adr|aws|docs|public|storage|terraform|tests|tests-examples)/")
+            echo $RELEVANT_DIRS
+            if [ -z "$RELEVANT_DIRS" ]; then
+              echo "No relevant changes detected, skipping rails tests"
+              exit 0
+            fi
```
bogsi17, line 37: There is also the `paths-ignore` property that can be defined directly on the trigger level which could simplify the logic, see 
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#example-including-and-excluding-branches-and-tags

TheOneFromNorway: That is for push events, not pull-request and it is only on workflow level AFAIK. Can it work for a specific job and for pull requests?

bogsi17 (approved):

.github/workflows/test.yml
```diff
@@ -5,9 +5,34 @@ on:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Determine if pre-rails tests are needed
+        id: check-rails-relevant-changes
+        run: |
+          if [ ${{ github.event_name }} == "pull_request" ]; then
+            git fetch origin ${{ github.event.pull_request.head.ref }}
+            RELEVANT_DIRS=$(git diff ${{ github.event.pull_request.base.sha }} origin/${{ github.event.pull_request.head.ref }}  --name-only)
+            echo "Relevant changes for rails: $RELEVANT_DIRS"
+            if [ -z "$(echo "$RELEVANT_DIRS" | egrep -v "^(.github|adr|aws|docs|public|storage|terraform|tests|tests-examples)/")" ]; then
```
thomasleese, line 20:
```suggestion
            if [ -z "$(echo "$RELEVANT_DIRS" | egrep -v "^(.github|adr|aws|docs|storage|terraform|tests|tests-examples)/")" ]; then
```

`public` contains files that are served by Rails so we would want to run the tests if any of these change.

thomasleese (changes requested):

.github/workflows/test.yml
```diff
@@ -5,9 +5,34 @@ on:
   workflow_call:
 
 jobs:
+  pre-rails-tests:
+    name: Pre Rails Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Determine if pre-rails tests are needed
```
thomasleese, line 13: I was initially confused about the name: "What's a pre-rails test?"

Do we need "pre-rails" when it could be "Determine if Rails tests need to run"?

TheOneFromNorway: Ahh typo will fix 👍

.github/workflows/test.yml
TheOneFromNorway: Will update

.github/workflows/test.yml
```diff
@@ -5,9 +5,34 @@ on:
...
+            if [ -z "$(echo "$RELEVANT_DIRS" | egrep -v "^(.github|adr|aws|docs|storage|terraform|tests|tests-examples)/")" ]; then
+              echo "No relevant changes detected, skipping rails tests"
+              echo "RUN_PRE_RAILS_TESTS=false" >> $GITHUB_OUTPUT
+              exit 0
+            fi
+          fi
+          echo "Not a pull request or detected relevant changes: Running rails tests"
+          echo "RUN_PRE_RAILS_TESTS=true" >> $GITHUB_OUTPUT
+    outputs:
+      run-pre-rails-tests: ${{ steps.check-rails-relevant-changes.outputs.RUN_PRE_RAILS_TESTS }}
```
thomasleese, line 29: I still think the naming is confusing here, `run-pre-rails-tests`, shouldn't this be `run-rails-tests`?

PR #3470, opened 2025-04-30
Ensure cron jobs reference valid classes

thomasleese:
This adds a test to ensure that the class names references in the cron job configuration are valid and refer to classes that exist.

We're adding this to catch any potential issues in the future where the cron jobs classes are invalid (as was fixed recently in https://github.com/nhsuk/manage-vaccinations-in-schools/commit/623a1e63809ed51e27028a369a526696947aec2c). This is necessary because Good Job itself doesn't check that the classes are valid and will fail silently.

### DIFF
--- a/config/cron_jobs.rb
+++ b/config/cron_jobs.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+CRON_JOBS = {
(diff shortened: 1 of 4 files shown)


benilovj (approved):
Nice!

PR #3473, opened 2025-04-30
Optimise offline session exporter

jhenderson:
This PR significantly improves the performance of the offline session exporter. It addresses concerns that the "Record offline" function was taking too long (up to 10s) and if sessions get bigger (eg with flu) the threshold to cause timeouts will be reached, and it will not be possible for these spreadsheets to be generated by SAIS teams.

While most sessions will have around 1,000 patients, community clinic sessions could potentially contain up to 60,000 patients. For this special case, we may need to implement a background job in the future to handle the export asynchronously, but hopefully, this optimisation should handle the vast majority of use cases.

[JIRA Ticket](https://nhsd-jira.digital.nhs.uk/browse/MAV-994)

Key improvements:
- Added comprehensive eager loading for all required associations
- Cached vaccination records in memory with `to_a` to avoid repeated DB calls
- Memoise `clinic_name_values`

Performance impact based on 1236 patient sessions:
- Total time: 6.45s → 2.47s (62% faster)
- SQL queries: 10,649 → 77 (99.3% reduction)
- SQL time: 1.83s → 0.30s (84% faster DB time)

The most significant improvement came from eliminating repeated queries for parent relationships, location names, and programme lookups which were previously causing thousands of separate database calls.

### DIFF
--- a/app/lib/reports/offline_session_exporter.rb
+++ b/app/lib/reports/offline_session_exporter.rb
@@ -125,17 +125,27 @@ class Reports::OfflineSessionExporter
   end
 
   def patient_sessions
-    session
-      .patient_sessions
-      .includes(
-        patient: [
-          :consent_statuses,
-          :school,
-          { vaccination_records: %i[batch performed_by_user vaccine] }
-        ],
-        session: :programmes
-      )
-      .order_by_name
+    @patient_sessions ||=
+      session
+        .patient_sessions
+        .includes(
+          patient: [
+            :consent_statuses,
+            :school,
+            { parent_relationships: :parent },
+            {
+              vaccination_records: %i[
+                batch
+                performed_by_user
+                vaccine
+                programme
+                session
+              ]
+            }
+          ],
+          session: [{ programmes: :vaccines }, :location]
+        )
+        .order_by_name
   end
 
   def consents
@@ -143,7 +153,7 @@ class Reports::OfflineSessionExporter
       Consent
         .where(patient_id: patient_sessions.select(:patient_id))
         .not_invalidated
-        .includes(:parent, :patient)
+        .includes(:parent, patient: { parent_relationships: :parent })
         .group_by(&:patient_id)
         .transform_values do
           it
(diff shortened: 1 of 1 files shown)


app/lib/reports/offline_session_exporter.rb
```diff
@@ -206,7 +216,9 @@ def rows(patient_session:)
       }
 
       vaccination_records =
-        patient.vaccination_records.select { it.programme_id == programme.id }
+        patient.vaccination_records.to_a.select do
```
thomasleese, line 219: I'm wondering what the `to_a` adds here? I would expect `select` to have already called it.

app/lib/reports/offline_session_exporter.rb
```diff
@@ -125,32 +125,42 @@ def columns
...
+        .order_by_name
   end
 
   def consents
     @consents ||=
       Consent
         .where(patient_id: patient_sessions.select(:patient_id))
         .not_invalidated
-        .includes(:parent, :patient)
+        .includes(:parent, patient: { parent_relationships: :parent })
```
thomasleese, line 156: We also include the parent from the `Consent`, do you know why we also need to include it through the parent relationships? Is there some code elsewhere that could be updated to not need it and instead get the parent from the consent model?

jhenderson: If we don't use `.includes(:parent, patient: { parent_relationships: :parent })` we get the following:

```
1. Count: 844 (91.8% of total)
   SELECT "parent_relationships".* FROM "parent_relationships" WHERE "parent_relationships"."patient_id" = N /*application='X'*/
```

I think I tracked down where these queries are coming from:

Starts here:

```
row[:health_question_answers] = Cell.new(
      health_question_answers(consents: grouped_consents),
      style: {
        alignment: {
          wrap_text: true
        }
      }
    )
```

then `health_question_answers` -> `ConsolidatedHealthAnswers` -> `HasHealthAnswers#who_responded` -> `Consent#parent_relationship`

thomasleese (approved):

PR #3474, opened 2025-04-30
Fix naming of variables in testing github workflow

TheOneFromNorway:
- Fixing a bad refactoring


thomasleese (approved):
I think it's also worth renaming the job to something like "Check if Rails tests need to run" for clarity.

PR #3477, opened 2025-04-30
Normalise spaces during import

alistair-white-horne-tng:
During an import, all string fields now have their whitespace normalised before any other processing occurs. This includes the headers

[MAV-1080](https://nhsd-jira.digital.nhs.uk/browse/MAV-1080)

### DIFF
--- a/lib/core_ext/string/normalise_whitespace.rb
+++ b/lib/core_ext/string/normalise_whitespace.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class String
+  def normalise_whitespace
+    result = self
+
+    # \u200D is a zero-width joiner (ZWJ) which is used in the frontend to display the NHS number
+    result = result.tr("\u200D", "")
+
+    # \u00A0 is a non-breaking space
+    result = result.tr("\u00A0", " ")
+
+    result.strip.gsub(/\s+/, " ").presence
+  end
+end
--- a/app/lib/csv_parser.rb
+++ b/app/lib/csv_parser.rb
@@ -107,13 +107,13 @@ class CSVParser
       row = info.line
       header = unconverted_headers[info.index]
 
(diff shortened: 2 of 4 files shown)


lib/core_ext/string/normalise_whitespace.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class String
+  # Normalises whitespace in a string by removing leading and trailing whitespace,
+  # replacing multiple spaces with a single space, and returning nil if the result is empty.
```
thomasleese, line 5: I'd suggest we probably don't need this comment. In general the approach I take is that if a comment like this is required then the code isn't clear enough. In this case, I think the code is pretty clear in what it does so I don't think the comments adds any value.

spec/lib/core_ext/string/string_normalise_whitespace_spec.rb
thomasleese, line 1: Because we're forced to add the `string_` prefix to appease the linters, I would suggest putting it in `core_ext` without the extra `string` directory.

spec/lib/core_ext/string/string_normalise_whitespace_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect("   ".normalise_whitespace).to be_nil
+    end
+
+    it "returns nil if string is empty" do
+      expect("".normalise_whitespace).to be_nil
+    end
+
+    context "with UTF-8 encoded strings" do
+      it "removes zero-width joiners (ZWJ)" do
+        string_with_zwj = "1234‍567890"
```
thomasleese, line 27: I'd suggest using explicit `\u200D` characters here to ensure they're not missed. Certainly in my editor the characters aren't rendered so it just looks like the same string before and after.

thomasleese (approved):

PR #3480, opened 2025-04-30
Add phone instructions for organisations and teams

murugapl:
Some organisations may need to provide additional guidance when sharing phone numbers (like menu options), and this change allows that information to be stored and consistently displayed throughout the application.

Added a phone_instructions string field to both organisation and team models, with corresponding database migrations. The instructions now appear in brackets next to phone numbers in all user-facing views and are included in the Notify personalisation for emails and text communications.

The onboarding process also supports importing these instructions.

### DIFF
--- a/app/views/parent_interface/consent_forms/cannot_consent_responsibility.html.erb
+++ b/app/views/parent_interface/consent_forms/cannot_consent_responsibility.html.erb
@@ -14,5 +14,5 @@
 
 <p>
   If you have any questions, please contact the local health organisation by calling
-  <%= @team.phone %>, or email <%= mail_to @team.email %>.
+  <%= format_phone_with_instructions(@team) %>, or email <%= mail_to @team.email %>.
 </p>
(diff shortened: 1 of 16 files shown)


app/views/parent_interface/consent_forms/cannot_consent_responsibility.html.erb
```diff
@@ -14,5 +14,5 @@
 
 <p>
   If you have any questions, please contact the local health organisation by calling
-  <%= @team.phone %>, or email <%= mail_to @team.email %>.
+  <%= @team.phone %><%= " (#{@team.phone_instructions})" if @team.phone_instructions.present? %>, or email <%= mail_to @team.email %>.
```
thomasleese, line 17: Since we have this logic in two places I think it's worth creating a helper method for it (see the files in `app/helpers` for other similar ones).

thomasleese (commented):
Looks good. I think we should also update the `generic_team` method on the organisation to create the team with any instructions.

https://github.com/nhsuk/manage-vaccinations-in-schools/blob/17d220839ae5891d62fcca4423769346996f10e3/app/models/organisation.rb#L70-L72

```rb
teams.create_with(email:, phone:, phone_instructions:).find_or_create_by!(name:)
```

thomasleese (approved):

PR #3487, opened 2025-05-01
Confirm school moves in bulk

thomasleese:
Locally the test `spec/features/import_child_records_slow_spec.rb` (which imports 1000 patients) speeds up from about 35 seconds to 25 seconds.


thomasleese:
This isn't going to be picked up any time soon, so closing this for now.

PR #3488, opened 2025-05-02
Add field length validation to import processes

jhenderson:
This PR adds length validation to fields in the import process, following up on the previous PR (#3444) that added similar validations to user input fields. 

**The changes include:**
- Added a `MAX_FIELD_LENGTH` constant (100 characters) to both `ImmunisationImportRow` and `PatientImportRow classes`
- Added validation to ensure batch names, clinic names, patient names, and school names don't exceed the maximum length
- Added tests to verify these validations work correctly


app/models/immunisation_import_row.rb
```diff
@@ -32,6 +32,8 @@ class ImmunisationImportRow
   CARE_SETTING_SCHOOL = 1
   CARE_SETTING_COMMUNITY = 2
 
+  MAX_FIELD_LENGTH = 100
```
thomasleese, line 35:
```suggestion
  MAX_FIELD_LENGTH = 300
```

I think actually we settled on 300 as the length for these kinds of fields, looking at Jake's comment in the Jira ticket.

thomasleese (approved):

PR #3489, opened 2025-05-02
Ensure vaccine is set when saving a vaccination record

thomasleese:
This fixes a regression that was introduced in 2073e01c1d6db2277cf677c5d4f9d00262c34469 where the vaccine isn't being set on vaccination records when recorded in the service itself. I've fixed the regression and then introduced a test to ensure this doesn't break again in the future.

### DIFF
--- a/app/models/draft_vaccination_record.rb
+++ b/app/models/draft_vaccination_record.rb
@@ -170,9 +170,15 @@ class DraftVaccinationRecord
 
   delegate :vaccine, to: :batch, allow_nil: true
 
+  delegate :id, to: :vaccine, prefix: true, allow_nil: true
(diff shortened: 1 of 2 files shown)


benilovj (approved):
👍 

Is there any extra spec coverage we need at higher level?

thomasleese:
> Is there any extra spec coverage we need at higher level?

It's tricky because Mavis handles vaccination records without vaccines correctly in the rest of the service (since they can be imported like that).

However, the CarePlus report stopped working because vaccinations records recorded by a SAIS organisation should always have a vaccine, therefore we would want it to fail as it's not normal state for these vaccination records.

PR #3490, opened 2025-05-02
Improve workflow for adding back into cohort

murugapl:
This change auto-confirms a school move when processing a cohort or class upload for a patient with a nil organisation. The generated school move is immediately confirmed—replicating the behavior of importing a new patient—so the confirmed move isn’t returned for further processing by the user.

### DIFF
--- a/app/models/patient_import.rb
+++ b/app/models/patient_import.rb
@@ -32,7 +32,8 @@ class PatientImport < ApplicationRecord
     @school_moves_to_save ||= Set.new
 
     if (school_move = row.to_school_move(patient))
-      if patient.school.nil? && !patient.home_educated
+      if (patient.school.nil? && !patient.home_educated) ||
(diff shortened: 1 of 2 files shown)


app/models/patient_import_row.rb
```diff
@@ -89,6 +89,13 @@ def to_school_move(patient)
         end
 
       school_move.tap { _1.source = school_move_source }
+
+      if patient.organisation.nil?
```
thomasleese, line 93: I don't think we should be calling this here because they're handled in bulk for performance reasons.

I think we need to add the logic to here instead to say if the organisation is nil then the school move needs to be confirmed: https://github.com/nhsuk/manage-vaccinations-in-schools/blob/17d220839ae5891d62fcca4423769346996f10e3/app/models/patient_import.rb#L35

PR #3491, opened 2025-05-02
Terraform documentation

TheOneFromNorway:
Update documentation to reflect process changes/updates

### DIFF
--- a/docs/terraform.md
+++ b/docs/terraform.md
@@ -29,7 +29,7 @@ sso_region = eu-west-2
 sso_registration_scopes = sso:account:access
 ```
 
-Before running `terraform ...` make sure you set the environment variable to the desired profile
+Before running `terraform ...` make sure you set the environment variable to the desired profile, e.g.
 
 ```bash
 export AWS_PROFILE=default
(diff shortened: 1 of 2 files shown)


docs/terraform.md
```diff
@@ -43,22 +43,31 @@ This repo contains 2 folders with terraform configuration.
...
+
+_Case 1:_ Setting up the first environment in an account
+
+To set up everything from scratch, run `./bootstrap.sh <ENV_NAME>` first in the `terraform/scripts` folder and follow
+any instructions from the output.
+
+_Case 2:_ Adding more environments to an account
+
+To add more environments to an account, run `./bootstrap.sh <ENV_NAME> --environment-only` in the `terraform/scripts`
+folder and follow any instructions from the output.
```
thomasleese, line 56: I wonder if the `bootstrap.sh` script should be changed to add environments to an account by default, and only create an account if you specify `--create-account` or something.

Just thinking that the default case should be the most common, and in this case I imagine the most common case is to not create an account at the same time. Not a blocker for this PR though since it documents the script as it is now.

thomasleese (approved):

docs/terraform.md
TheOneFromNorway: This sounds like a good idea, it will be implemented in a follow up.

PR #3494, opened 2025-05-05
Various linting improvements

thomasleese:
This makes a number of small changes to the linting configuration to fix a number of small issues related to switching to `hk`. See individual commits for more details.

### DIFF
--- a/hk.pkl
+++ b/hk.pkl
@@ -22,10 +22,10 @@ local linters = new Mapping<String, Group> {
             }
 
             ["rufo"] = new Step {
-                glob = List("**/*.erb")
+                glob = "*.erb"
                 exclude = List("**/layouts/*")
                 check = "bin/rufo --check {{ files }}"
-                fix = "bin/rufo {{ files }}"
+                fix = "bin/rufo --simple-exit {{ files }}"
             }
 
             ["terraform"] = (Builtins.terraform) {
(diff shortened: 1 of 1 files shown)


hk.pkl
```diff
@@ -28,11 +28,10 @@ local linters = new Mapping<String, Group> {
                 fix = "bin/rufo --simple-exit {{ files }}"
             }
 
-            ["terraform"] = (Builtins.terraform) {
-            }
+            ["terraform"] = Builtins.terraform
 
             ["tf_lint"] = new Step {
-                glob = List("**/*.tf")
+                glob = "*.tf"
```
TheOneFromNorway, line 34: I believe single `*` will not match slashes or back-slashes in Windows machines, leavin the **/* would probably be safer

thomasleese: I'm trying to get these as close as possible to the builtins (so in the future we could potentially switch over and not use our own custom ones): https://github.com/jdx/hk/blob/444918a0a6a7698dace21fa7c2ec2e7912ff8a88/pkl/builtins/tf_lint.pkl#L4

If it's true that it doesn't work on Windows, do we need to perhaps raise this issue upstream?

thomasleese: It's also how the Terraform builtin works currently, which we are using: https://github.com/jdx/hk/blob/444918a0a6a7698dace21fa7c2ec2e7912ff8a88/pkl/builtins/terraform.pkl#L4

thomasleese: I think I'll merge this in as is, and if it's a problem we can address it then (and fix this upstream for others who might use `hk` as well).

I don't think the service is well supported on Windows as it is.

PR #3504, opened 2025-05-06
Fixing tf_lint stage of hk

TheOneFromNorway:
- Without specifying the configuration file it does not apply correct fixes
  - Pathing is relative to directory hk is running in, therefore with --recursive we must pass absolute paths

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,6 +23,7 @@ jobs:
       - run: yarn install --immutable --immutable-cache --check-cache
       - uses: hashicorp/setup-terraform@v3
       - uses: terraform-linters/setup-tflint@v4
+      - run: tflint --chdir=terraform --init
       - uses: jdx/mise-action@v2
         with:
           install_args: hk pkl
--- a/hk.pkl
+++ b/hk.pkl
@@ -32,8 +32,8 @@ local linters = new Mapping<String, Group> {
 
             ["tf_lint"] = new Step {
                 glob = "*.tf"
-                check = "tflint --chdir=terraform --recursive"
-                fix = "tflint --chdir=terraform --recursive --fix"
+                check = "tflint --chdir=terraform --config=$(pwd)/terraform/.tflint.hcl --recursive"
+                fix = "tflint --chdir=terraform --config=$(pwd)/terraform/.tflint.hcl --recursive --fix"
             }
         }
     }


hk.pkl
```diff
@@ -32,8 +32,8 @@ local linters = new Mapping<String, Group> {
 
             ["tf_lint"] = new Step {
                 glob = "*.tf"
-                check = "tflint --chdir=terraform --recursive"
-                fix = "tflint --chdir=terraform --recursive --fix"
+                check = "tflint --chdir=terraform --config=$(pwd)/terraform/.tflint.hcl --recursive"
```
thomasleese, line 35: I know we had this in the old `bin/lint`, but I still don't understand why it's needed:

- The `--chdir=terraform` will ensure that `tflint` runs with a current working directory of `terraform`.
- The default `--config` value is `.tflint.hcl`, which means it will load the file `.tflint.hcl` in the current working directory of `terraform` which is the same as `--config=$(pwd)/terraform/.tflint.hcl`.

TheOneFromNorway: Because AFAIK tflint looks for .tflint.hcl in each directory it is working in unless you give it another file to look at. --recursive basically means it looks for a different .tflint.hcl file in every sub-directory it is searching. If you change a terraform resource name from snake-case to camelcase and try running `tflint` both ways you will see the difference

.github/workflows/lint.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       - run: yarn install --immutable --immutable-cache --check-cache
       - uses: hashicorp/setup-terraform@v3
       - uses: terraform-linters/setup-tflint@v4
+      - run: tflint --init --config=terraform/.tflint.hcl
```
thomasleese, line 26:
```suggestion
      - run: tflint --chdir=terraform --init
```

I think this is slightly nicer as it's consistent with the commands in `hk.pkl` although I think it's functionally the same.

hk.pkl
thomasleese: Oh right, I see. Needing `$(pwd)` makes me nervous, but it seems like it's the way `tflint` is expected to work.

I wonder if the more usual convention is to have a `.tflint.hcl` file in each module, but then you could end up with inconsistent configuration so I'm not convinced that's any better.

thomasleese (approved):

.github/workflows/lint.yml
TheOneFromNorway: Done

PR #3505, opened 2025-05-06
Reduce spacing between header navigation items above tablet breakpoint

paulrobertlloyd:
`nhsuk-frontend` v9.4 increased this space to 16px which, when accumulated across all navigation items, moves the last item beyond the container on desktop. We’ll override this value to 12px to resolve this issue in the short term.


thomasleese (approved):
Looks good to me locally.

PR #3509, opened 2025-05-06
Automatically accept preferred names

alistair-white-horne-tng:
If a patient has either `preferred_given_name` or `preferred_family_name` set to `nil` in Mavis, but a name (eg "Benny") is uploaded for a matching record (in either a class upload, or cohort upload), this is now automatically accepted, rather than prompting the user to review.

[MAV-1100](https://nhsd-jira.digital.nhs.uk/browse/MAV-1100)

### DIFF
--- a/app/models/patient_import_row.rb
+++ b/app/models/patient_import_row.rb
@@ -57,6 +57,19 @@ class PatientImportRow
         %w[male female not_specified]
       )
 
+      auto_accept_attribute(
+        existing_patient,
+        attributes,
+        :preferred_given_name,
+        :present?
+      )
+      auto_accept_attribute(
+        existing_patient,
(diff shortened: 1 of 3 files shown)


app/models/patient_import_row.rb
```diff
@@ -53,6 +53,18 @@ def to_patient
         existing_patient.gender_code = attributes.delete(:gender_code)
       end
 
+      if attributes[:preferred_given_name].present? &&
```
thomasleese, line 56: I'm wondering if it's time to extract this common logic in to methods, where the attribute name can be passed in. The current method is getting quite long and has got a fair amount of duplication.

thomasleese (approved):

PR #3516, opened 2025-05-07
Add vaccine code column to CarePlus vaccination reports

jhenderson:
CWPT CHIS requested another column for vaccine codes such as FLU, FLUENZ, HPV, 3IN1, ACWYX4 because they cannot upload via Snomed codes at the moment.

[JIRA Ticket](https://nhsd-jira.digital.nhs.uk/browse/MAV-1015)

### DIFF
--- a/spec/lib/reports/careplus_exporter_spec.rb
+++ b/spec/lib/reports/careplus_exporter_spec.rb
@@ -10,66 +10,210 @@ describe Reports::CareplusExporter do
     )
   end
 
-  shared_examples "generates a report" do |programme_type|
-    let(:programme) { create(:programme, programme_type) }
+  shared_examples "generates a report" do
+    let(:programmes) { [programme] }
 
-    context "#{programme_type} programme" do
-      let(:programmes) { [programme] }
+    let(:organisation) do
+      create(:organisation, careplus_venue_code: "ABC", programmes:)
+    end
+    let(:location) do
+      create(
+        :school,
+        gias_local_authority_code: 123,
+        gias_establishment_number: 456
+      )
+    end
+    let(:session) { create(:session, organisation:, programmes:, location:) }
+    let(:parsed_csv) { CSV.parse(csv) }
+    let(:headers) { parsed_csv.first }
+    let(:data_rows) { parsed_csv[1..] }
 
-      let(:organisation) do
-        create(:organisation, careplus_venue_code: "ABC", programmes:)
-      end
-      let(:location) do
-        create(
-          :school,
-          gias_local_authority_code: 123,
-          gias_establishment_number: 456
-        )
-      end
-      let(:session) { create(:session, organisation:, programmes:, location:) }
-      let(:parsed_csv) { CSV.parse(csv) }
-      let(:headers) { parsed_csv.first }
-      let(:data_rows) { parsed_csv[1..] }
+    it "includes the expected headers" do
+      expect(headers).to include(
+        "NHS Number",
+        "Surname",
+        "Forename",
+        "Date of Birth",
+        "Address Line 1",
+        "Person Giving Consent",
+        "Ethnicity",
+        "Date Attended",
+        "Time Attended",
+        "Venue Type",
+        "Venue Code",
+        "Staff Type",
+        "Staff Code",
+        "Attended",
+        "Reason Not Attended",
+        "Suspension End Date"
+      )
 
-      it "includes the expected headers" do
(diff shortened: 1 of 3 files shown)


misaka (approved):

spec/lib/reports/careplus_exporter_spec.rb
```diff
@@ -245,7 +261,20 @@
...
+    it "uses FLUENZ code for nasal delivery" do
+      create_vaccination_record(delivery_method: :nasal_spray, session:)
+      code_index = headers.index("Vaccine Code 1")
+      row = parsed_csv[1]
+
+      expect(row[code_index]).to eq("FLUENZ")
+    end
+  end
+
+  it_behaves_like "generates a report", "hpv"
```
jhenderson, line 276: Changed `include_examples` to `it_behaves_like ` because if for example `hpv` passes, the subsequent shared examples pass as well even though the specs should fail.

I believe `include_examples` simply includes the examples directly in the current context without additional nesting. With `it_behaves_like`, each programme type gets its own isolated test context, preventing potential state leakage between examples.

spec/lib/reports/careplus_exporter_spec.rb
```diff
@@ -10,25 +10,50 @@
     )
   end
 
+  let(:programme) { create(:programme, programme_type) }
```
jhenderson, line 13: Had to move a good bit of state up here so I could test `FLUENZ`

spec/lib/reports/careplus_exporter_spec.rb
thomasleese: Interesting, there are a few other places where we use `includes_examples`. I wonder if it's worth changing those too?

spec/lib/reports/careplus_exporter_spec.rb
```diff
@@ -10,25 +10,50 @@
...
+  def create_vaccination_record(options = {})
+    patient_session =
+      create(
+        :patient_session,
+        :consent_given_triage_not_needed,
+        programmes:,
+        session:
+      )
+
+    vaccine = create(:vaccine, programme:)
```
thomasleese, line 39: I think these can be pulled out in to their own `let` as they don't rely on the `options` passed in.

spec/lib/reports/careplus_exporter_spec.rb
```diff
@@ -10,25 +10,49 @@
...
+      :vaccination_record,
+      {
+        programme:,
+        vaccine:,
+        patient: patient_session.patient,
+        session: patient_session.session,
+        performed_at: 2.weeks.ago
+      }.merge(attributes)
+    )
+  end
```
thomasleese, line 50: I think instead of having a method like this, there could be an option to refactor the `shared_examples`, so they can be used like this:

```rb
it_behaves_like "generates a report", "flu" do
  let(:programme) { create(:programme, :flu) }
  let(:delivery_method) { :intramuscular }
end

it_behaves_like "generates a report", "flu nasal spray" do
  let(:programme) { create(:programme, :flu) }
  let(:delivery_method) { :nasal_spray }
end
```

I think this would be a more conventional way of structuring these tests.

spec/lib/reports/careplus_exporter_spec.rb
jhenderson: I think so

spec/lib/reports/careplus_exporter_spec.rb
jhenderson: I agree 👍

spec/lib/reports/careplus_exporter_spec.rb
```diff
@@ -11,24 +11,38 @@
...
+      create(
+        :patient_session,
+        :consent_given_triage_not_needed,
+        programmes:,
+        session:
+      )
+    end
+    let(:expected_vaccine_code) do
+      described_class::PROGRAMME_TYPE_TO_VACCINE_CODE.fetch(programme_type)
+    end
```
thomasleese, line 43: I think this can all go inside the `context` below it (like some of these used to be) - and I think that might even allow us to go back to `includes_examples` maybe?

thomasleese: Ah yeah, okay, that makes sense because the `let(:programme)` is defined outside the context.

thomasleese (approved):

thomasleese (approved):

thomasleese (approved):

PR #3523, opened 2025-05-08
Persist filters, per user, per view, per browser session

jhenderson:
User feedback suggested that filters maintain state because they often filter patients, move away from the page, and then come back having to apply the same filters again.

This update will ensure each page with filters maintains its own unique state. This state can only be cleared with the "Clear filters" button on the form.

**Test Guidance:**

1. Visit the various pages with filters
2. Apply filters to each of those pages
3. Visit other pages and come back to the filtered page
4. The filters previously set should still be there even though there's no URL params
5. Click the "Clear filters" button to reset the filters (it should only apply to that particular page)

### DIFF
--- a/app/components/app_search_component.rb
+++ b/app/components/app_search_component.rb
@@ -98,7 +98,7 @@ class AppSearchComponent < ViewComponent::Base
           <% if show_buttons_in_details? %>
             <div class="app-button-group">
               <%= f.govuk_submit "Update results", secondary: true, class: "app-button--small" %>
-              <%= govuk_button_link_to "Clear filters", @url, secondary: true, class: "app-button--small" %>
+              <%= govuk_button_link_to "Clear filters", clear_filters_path, class: "app-button--small app-button--secondary" %>
             </div>
           <% end %>
         <% end %>
@@ -106,7 +106,7 @@ class AppSearchComponent < ViewComponent::Base
         <% unless show_buttons_in_details? %>
           <div class="app-button-group">
             <%= f.govuk_submit "Update results", secondary: true, class: "app-button--small" %>
-            <%= govuk_button_link_to "Clear filters", @url, secondary: true, class: "app-button--small" %>
+            <%= govuk_button_link_to "Clear filters", clear_filters_path, class: "app-button--small app-button--secondary" %>
           </div>
         <% end %>
       <% end %>
@@ -158,5 +158,9 @@ class AppSearchComponent < ViewComponent::Base
         register_statuses.any? || session_statuses.any? ||
         triage_statuses.any? || year_groups.any?
     )
+  end
+
+  def clear_filters_path
+    "#{@url}?search_form[clear_filters]=true"
   end
 end
--- a/app/forms/search_form.rb
+++ b/app/forms/search_form.rb
@@ -5,6 +5,9 @@ class SearchForm
   include ActiveModel::Attributes
   include ActiveRecord::AttributeAssignment
 
+  SESSION_KEY = "search_filters"
+
+  attribute :clear_filters, :boolean
   attribute :consent_status, :string
   attribute :date_of_birth_day, :integer
   attribute :date_of_birth_month, :integer
@@ -16,6 +19,13 @@ class SearchForm
   attribute :session_status, :string
   attribute :triage_status, :string
   attribute :year_groups, array: true
+
+  attr_accessor :session, :request_path
+
+  def initialize(params = {})
+    super(params)
+    handle_session_filters
+  end
 
   def year_groups=(values)
     super(values&.compact_blank&.map(&:to_i)&.compact || [])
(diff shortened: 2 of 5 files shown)


app/components/app_search_component.rb
```diff
@@ -98,15 +98,15 @@ class AppSearchComponent < ViewComponent::Base
...
+              <%= govuk_button_link_to "Clear filters", clear_filters_path, class: "app-button--small app-button--secondary" %>
             </div>
           <% end %>
         <% end %>
 
         <% unless show_buttons_in_details? %>
           <div class="app-button-group">
             <%= f.govuk_submit "Update results", secondary: true, class: "app-button--small" %>
-            <%= govuk_button_link_to "Clear filters", @url, class: "app-button--small app-button--secondary" %>
+            <%= govuk_button_link_to "Clear filters", clear_filters_path, class: "app-button--small app-button--secondary" %>
```
thomasleese, line 109: Rather than specifying a path, I think it's possible to set the name of the button to `clear_filters` with a value of `true, I think that's a cleaner approach:

```suggestion
            <%= govuk_button_link_to "Clear filters", clear_filters_path, class: "app-button--small app-button--secondary", name: "clear_filters", value: "true" %>
```

app/forms/search_form.rb
```diff
@@ -17,6 +20,15 @@ class SearchForm
   attribute :triage_status, :string
   attribute :year_groups, array: true
 
+  attr_accessor :session, :request_path
```
thomasleese, line 23: I think these can be private.

app/forms/search_form.rb
```diff
@@ -17,6 +20,15 @@ class SearchForm
...
+
+  def initialize(params = {}, session:, request_path:)
+    @session = session
+    @request_path = request_path
+    super(params)
```
thomasleese, line 28:
```suggestion
  def initialize(params = {})
    super(params)
```

I think `ActiveModel::Model` will handle the assignments for us.

app/forms/search_form.rb
```diff
@@ -62,4 +74,54 @@ def apply(scope, programme: nil)
...
+    end
+  end
+
+  def load_from_session
+    return if session.nil? || session[SESSION_KEY].blank?
+
+    stored_filters = session[SESSION_KEY][path_key]
+    return if stored_filters.blank?
+
+    assign_attributes(stored_filters)
```
thomasleese, line 105: Does this mean that stored attributes will take priority over those passed in as query parameters if this happens at the end of `initialize`?

spec/forms/search_form_spec.rb
thomasleese, line 1: I think we should also consider adding a feature test that covers the full scenario of navigating away from a page with filters, and then going back to the page and seeing the filters still there?

app/components/app_search_component.rb
jhenderson: Unfortunately it doesn't work. It will only add html attributes to the link tag: Example:

`<a class="nhsuk-button app-button--small app-button--secondary" data-module="nhsuk-button" name="clear_filters" value="true" href="/sessions/session-2/consent" data-govuk-button-init="">Clear filters</a>`

app/forms/search_form.rb
jhenderson: `ActiveModel::Model` does indeed handle assignments, but only for defined attributes. I chose to configure the arguments to differentiate between `params` and other collaborator objects such as `session` and `request_path`. If we want to go ahead with the suggested approach, we'll have to define  `session` and `request_path` as attributes.

Then we'd have to change the usage so that we have something like:

`SearchForm.new(**params, session:, request_path:)` or `SearchForm.new(params.merge(session:, request_path:))`.

What do you think?

thomasleese: I think that's fine, we already do a similar approach in our other forms (for example here https://github.com/nhsuk/manage-vaccinations-in-schools/blob/d6ee775bc090b817209d3b9aa1b80ff03fd6860f/app/forms/batch_form.rb#L7) where the `batch` doesn't come from the query parameters, but it's still passed in like the attributes that do.

app/components/app_search_component.rb
thomasleese: Oh yes, sorry, I missed that it was a link and thought it was a `<button>`

app/forms/search_form.rb
jhenderson: No, `handle_session_filters` will check if any of the attributes are present and if so, overrides the stored filters immediately. `load_from_session` only gets called if all the attributes are blank which would only happen if the user visited the page without any filters.

spec/forms/search_form_spec.rb
jhenderson: Added the feature spec. Wasn't sure if I needed to create patients with various statuses and verify the results, but I assumed that's covered elsewhere.

thomasleese (approved):

PR #3532, opened 2025-05-12
Patient cards in session lists should show a status name

jhenderson:
This PR updates the status labels on patient cards to match the corresponding filter labels across different views in the session management interface.

Currently, there's inconsistency between the labels used in filters and those displayed on patient cards.

Modified the `AppPatientSessionSearchResultCardComponent` to use context-specific status labels instead of the generic "Status" label:

[JIRA Ticket](https://nhsd-jira.digital.nhs.uk/browse/MAV-1174)

### Screenshots

| View | Before | After |
| --- | --- | --- |
| Consent | Screenshot 2025-05-12 at 15 57 16 | Screenshot 2025-05-12 at 15 59 26 |
| Triage | Screenshot 2025-05-12 at 15 58 17 | Screenshot 2025-05-12 at 15 59 42 |
| Session outcomes | Screenshot 2025-05-12 at 15 58 36 | Screenshot 2025-05-12 at 15 59 58 |


paulrobertlloyd:
@jhenderson Filter name for `Session outcomes` should be `Session outcome`

app/components/app_patient_session_search_result_card_component.rb
```diff
@@ -25,8 +25,8 @@ class AppPatientSessionSearchResultCardComponent < ViewComponent::Base
 
             if status_tag
               summary_list.with_row do |row|
```
thomasleese, line 27: I think it might be possible to pass the row structure straight in to `with_row`, something like `with_row(status_row)`, maybe?

https://govuk-components.netlify.app/components/summary-list/#building-a-summary-list-directly-from-data

jhenderson: Doesn't appear to work - `with_row` only takes a block.

thomasleese: Hmm I wonder if it only works if all the rows are provided in that hash format?

jhenderson: Looks like the only way is where one does `govuk_summary_list(rows: rows)` and then `rows` looks something like:

```
{
  rows: [
    { key: { text: "Name" }, value: { text: "Hercule Poirot" } },
    {
      key: { text: "Address" },
      value: { text: "Flat 203, 56B Whitehaven Mansions, Charterhouse Square, London" },
      actions: [{ href: "#", visually_hidden_text: "address" }]
    }
  ]
}
```

but our component is a bit more bespoke than that I think.

config/locales/status.en.yml
```diff
@@ -1,5 +1,10 @@
 en:
   status:
+    label:
```
jhenderson, line 3: @thomasleese Wasn't sure where to put these translations. Started out with `I18n.t("status.consent.label.status")` but then thought it didn't read well, so went with `I18n.t("status.label.consent")`. Happy to take other suggestions.

thomasleese: Yeah, I think that works well. By having the dynamic part at the end it also allows us to do things like `I18n.t("consent", scope: %i[status label])` which I think is nice.

thomasleese (approved):

thomasleese (approved):

PR #3541, opened 2025-05-13
Stop requesting triage for already vaccinated patients

alistair-white-horne-tng:
If a patient has already been vaccinated, then they don't need another vaccination. This means that they don't need to be triaged.

Previously, Mavis would've requested triage for this patient. Now it won't.

[MAV-1089](https://nhsd-jira.digital.nhs.uk/browse/MAV-1089)

### DIFF
--- a/app/models/patient/triage_status.rb
+++ b/app/models/patient/triage_status.rb
@@ -68,24 +68,33 @@ class Patient::TriageStatus < ApplicationRecord
   def vaccination_history_requires_triage?
     vaccination_records.any? do
       it.programme_id == programme_id && it.administered?
-    end && !VaccinatedCriteria.call(programme:, patient:, vaccination_records:)
+    end && !vaccinated?
   end
 
   private
 
+  def vaccinated?
+    @vaccinated ||=
+      VaccinatedCriteria.call(programme:, patient:, vaccination_records:)
+  end
+
   def status_should_be_safe_to_vaccinate?
+    return false if vaccinated?
     latest_triage&.ready_to_vaccinate?
   end
 
   def status_should_be_do_not_vaccinate?
+    return false if vaccinated?
     latest_triage&.do_not_vaccinate?
   end
 
   def status_should_be_delay_vaccination?
+    return false if vaccinated?
     latest_triage&.delay_vaccination?
   end
 
   def status_should_be_required?
+    return false if vaccinated?
     return true if latest_triage&.needs_follow_up?
 
     return false if latest_consents.empty?
--- a/app/controllers/draft_vaccination_records_controller.rb
+++ b/app/controllers/draft_vaccination_records_controller.rb
@@ -121,8 +121,6 @@ class DraftVaccinationRecordsController < ApplicationController
 
     send_vaccination_confirmation(@vaccination_record) if should_notify_parents
 
-    @vaccination_record.triage_patient_as_do_not_vaccinate!
-
     # In case the user navigates back to try and edit the newly created
     # vaccination record.
     @draft_vaccination_record.update!(editing_id: @vaccination_record.id)
--- a/app/models/vaccination_record.rb
+++ b/app/models/vaccination_record.rb
@@ -175,33 +175,6 @@ class VaccinationRecord < ApplicationRecord
     performed_at.to_date.academic_year
   end
 
-  def triage_patient_as_do_not_vaccinate!
-    return unless already_had?
-
-    # If an "already had" vaccination record is created, and the patient needs
-    # triage, we should record a "do not vaccinate" triage to make sure the
-    # patient isn't double vaccinated.
-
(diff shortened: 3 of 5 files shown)


app/models/patient/triage_status.rb
```diff
@@ -47,6 +47,10 @@ class Patient::TriageStatus < ApplicationRecord
        validate: true
 
   def assign_status
+    return self.status = :not_required if patient.vaccination_status(
+      programme:
+    ).vaccinated?
```
thomasleese, line 52: This introduces a dependency between the `TriageStatus` model and the `VaccinationStatus` model, meaning the latter has to be updated before the former. I think this has wider consequences as it means we need to be more careful about the updating of statuses (they can't be done in parallel or independently of each other).

So far, we've avoided this by having each `Status` model generate statuses based purely on the raw data about a patient (consents, triage, vaccination records, etc). I would suggest we stick with that approach to give us the flexibility mentioned above.

In this case, the `vaccination_records` are already available in this model from the `has_many` above, and in fact we already use this to determine the vaccination status in the `vaccination_history_requires_triage?` (by using the `VaccinatedCriteria` class).

Separately, I think we should keep the existing style of this method where we have what is effectively a single `case` statement, so if we can add this conditional to the various `status_should_be` methods, that would be preferred.

alistair-white-horne-tng: All sounds very sensible

app/models/patient/triage_status.rb
```diff
@@ -73,19 +73,27 @@ def vaccination_history_requires_triage?
 
   private
 
+  def vaccinated?
```
thomasleese, line 76: I think we can use this method in `vaccination_history_requires_triage?` to avoid duplication.

app/models/patient/triage_status.rb
```diff
@@ -73,19 +73,27 @@ def vaccination_history_requires_triage?
...
+    VaccinatedCriteria.call(programme:, patient:, vaccination_records:)
```
thomasleese, line 77: Since this method is being called multiple times I think it's worth caching the value.

```suggestion
    @vaccinated ||= VaccinatedCriteria.call(programme:, patient:, vaccination_records:)
```

thomasleese, line 77: Since this method is being called multiple times I think it's worth caching the value.

```suggestion
    @vaccinated ||= VaccinatedCriteria.call(programme:, patient:, vaccination_records:)
```

app/models/patient/triage_status.rb
```diff
@@ -73,19 +73,27 @@ def vaccination_history_requires_triage?
...
+  end
+
   def status_should_be_safe_to_vaccinate?
+    return false if vaccinated?
```
thomasleese, line 81: Is this right? It means that once a patient has been vaccinated, even if a triage was performed we lose track of these patients and nurses can no longer find them in the triage tab.

My understanding of the ticket is that we only wanted to hide patients if they had been vaccinated and were marked as needing triage.

app/models/patient/triage_status.rb
alistair-white-horne-tng: Done

app/models/patient/triage_status.rb
alistair-white-horne-tng: No, I actually think that it should override whatever the last triage outcome was. I certainly think it would be ambiguous if the triage status was "safe to vaccinate" if the child has already had the vaccine.

I would argue that it should behave the same as if the vaccination record were there before the consent came in; arguably the consent should be invalidated too because it doesn't make sense to give consent for a vaccination which shouldn't be administered.

thomasleese: Okay, I think we should discuss this with @paulrobertlloyd as I think this isn't quite how it was designed.

paulrobertlloyd:
@alistair-white-horne-tng Let’s save us having a meeting…

Spoke to @thomasleese, and I agree with your humble opinion on this; they should no longer appear in triage (the prototype also needs to be updated to reflect this).

The triage tab only shows children that need triage, or have had triage (or for which a triage outcome can still be changed). Once a child is vaccinated, they can no longer be triaged, so they shouldn’t appear in the triage tab.

thomasleese (approved):
I think we need to update the scope of the Jira issue to match this change where it's not just about the "Needs triage" status, but all the triage statuses.

alistair-white-horne-tng:
> Do we need to make any changes to this logic
> 
> https://github.com/nhsuk/manage-vaccinations-in-schools/blob/d7b8aab09c049565509ebf834c70f322568bcaf4/app/models/vaccination_record.rb#L178-L203
> 
> ?
> Currently when you mark as patient as already had, they get triaged as do not vaccinate.

@thomasleese, I've simply got rid of it. I'm pretty sure that doesn't make it any more likely that the child will get vaccinated. Do you think the same? Certainly all the unit tests are still green

PR #3548, opened 2025-05-13
Deduplicate health answers with followups

thomasleese:
This updates the deduplication logic introduced to support health answers across multiple vaccines to also support health answers with followup questions, which so far aren't in use but will be used for Flu.

I've introduced a new class `HealthAnswersDeduplicator` which handles this logic and then re-instated the tests we had for Flu which are currently to show that it's working as expected.

I think there are likely some scenarios where the deduplicator doesn't work correctly, most likely if there are questions with the same title but different followups, however we don't know of any where that is the case currently.

[Jira Issue](https://nhsd-jira.digital.nhs.uk/browse/MAV-1176)

### DIFF
--- a/app/lib/health_answers_deduplicator.rb
+++ b/app/lib/health_answers_deduplicator.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+class HealthAnswersDeduplicator
+  def initialize(vaccines:)
+    @vaccines = vaccines
+  end
+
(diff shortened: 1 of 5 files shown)


jhenderson (approved):
I'm wondering if it's worth adding a documentation commen at the top to get a better idea what the code is trying to accomplish given the complexity of the algorithm. Just a thought.

PR #3549, opened 2025-05-14
Use inputs.git-sha to checkout out code for build

misaka:
Otherwise we always build from using the workflow ref, but tag using the
deploy sha.


bogsi17 (approved):

thomasleese (approved):
🚀

PR #3555, opened 2025-05-15
Track user IDs in Sentry

thomasleese:
When sending errors to Sentry this allows us to track which user was logged in at the time of the error, helping us to investigate issues. Only the anonymised user ID will be sent to Sentry rather than any personally identifiable information.


benilovj:
@thomasleese I reckon we merge this in.

PR #3556, opened 2025-05-15
Ensure non-unicode strings aren't normalised

thomasleese:
In rare scenarios it's possible to end up calling `normalise_whitespace` on a non-unicode string. In this case, we shouldn't be removing whitespace characters that are specific to unicode as a `Encoding::CompatibilityError` error ends up being raised.

This has only happened in practice once, and only in the test environment: https://good-machine.sentry.io/issues/6604598285/

In this case, a CSV file was uploaded where the encoding couldn't be detected, which lead to a unhandled exception being raised, rather than a validation error on the file.

### DIFF
--- a/lib/core_ext/string/normalise_whitespace.rb
+++ b/lib/core_ext/string/normalise_whitespace.rb
@@ -4,11 +4,13 @@ class String
   def normalise_whitespace
     result = self
 
-    # \u200D is a zero-width joiner (ZWJ) which is used in the frontend to display the NHS number
-    result = result.tr("\u200D", "")
+    if result.encoding == Encoding::UTF_8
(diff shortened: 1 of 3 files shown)


lib/core_ext/string/normalise_whitespace.rb
alistair-white-horne-tng, line 1: Ah yes, I initially implemented this too, but took it out because I (evidently incorrectly) assumed that encoding would be normalised as the CSV was parsed (before `normalise_whitespace` was executed)

thomasleese: In most cases it is, there is a rare scenario where the encoding on the CSV file couldn't be detected, and therefore is left as a string of bytes.

PR #3558, opened 2025-05-15
Generate consent status from submitted at time

thomasleese:
This changes how we determine the most recent consent response from the `created_at` to the `submitted_at`. In most cases these values will be the same, but it's possible for a consent to be matched with a patient manually which can occur some period of time after the consent was submitted.

In particular, it's possible to manually match multiple consents in a different order to when they were submitted, which can lead to bugs where the patient is given the wrong consent status.

This has been implemented by storing a `submitted_at` column on consent responses, which has been done to ensure we don't hit performance issues when filtering or ordering consents where we need to fetch the related consent form.

[Jira Issue](https://nhsd-jira.digital.nhs.uk/browse/MAV-1196)


benilovj (approved):
LGTM

PR #3560, opened 2025-05-16
Normalise whitespace in NHS number

thomasleese:
When editing the NHS number of a patient we should normalise the whitespace on the NHS number to ensure that it's accepted and is valid. We do this already for spaces, but not for special unicode whitespace characters.

[Jira Issue](https://nhsd-jira.digital.nhs.uk/browse/MAV-504)

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -186,7 +186,10 @@ class Patient < ApplicationRecord
 
   encrypts :address_line_1, :address_line_2, :address_town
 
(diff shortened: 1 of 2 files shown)


alistair-white-horne-tng (approved):
Good spot. I presume you've also checked other normalisations in other record types too?

thomasleese:
> I presume you've also checked other normalisations in other record types too?

Yeah, this was the only one that was relevant.

PR #3565, opened 2025-05-16
Enable filtering children in programme by consent and triage status

jhenderson:
This PR introduces the ability to filter children within a programme based on their consent and triage status, in addition to the existing programme outcome filter.

This enhancement addresses the need for teams to efficiently identify and manage children based on their consent and triage outcomes, improving workflow and data accessibility.

Note: children who don’t need triage (didn’t answer yes to any health questions) don't show a triage status row in the card component unless specific triage filters are applied.


app/views/programmes/patients.html.erb
```diff
@@ -22,6 +22,8 @@
           form: @form,
           url: patients_programme_path(@programme),
           programme_statuses: Patient::VaccinationStatus.statuses.keys,
+          consent_statuses: Patient::ConsentStatus.statuses.keys,
+          triage_statuses: %w[safe_to_vaccinate do_not_vaccinate delay_vaccination required],
```
jhenderson, line 26: I copied this from `app/views/programmes/patients.html.erb`. Was wondering if we could put it somewhere and share it between the two partials. Any ideas or just leave it for now?

app/views/programmes/patients.html.erb
```diff
@@ -30,7 +32,12 @@
     <%= render AppSearchResultsComponent.new(@pagy) do %>
       <% @patients.each do |patient| %>
         <%= render AppPatientSearchResultCardComponent.new(
-              patient, link_to: patient_path(patient), programme: @programme, show_year_group: true,
+              patient,
+              link_to: patient_path(patient),
+              programme: @programme,
+              consent_status: @form.consent_status,
```
jhenderson, line 38: I have some doubt about passing in  `consent_status` and `triage_status` here using the `@form` object - the interface is big enough.

I thought about accessing this data using params inside `AppPatientSearchResultCardComponent`, example `params[:search_form][:consent_status]` but since we already process those params via `SearchForm` I thought maybe using `SearchForm` is better.

What do you think?

app/views/sessions/triage/show.html.erb
```diff
@@ -13,7 +13,7 @@
     <%= render AppSearchComponent.new(
           form: @form,
           url: session_triage_path(@session),
-          triage_statuses: %w[safe_to_vaccinate do_not_vaccinate delay_vaccination required],
+          triage_statuses: %w[required delay_vaccination do_not_vaccinate safe_to_vaccinate],
```
jhenderson, line 16: I copied this from `app/views/programmes/patients.html.erb`. Was wondering if we could put it somewhere and share it between the two partials. Any ideas or just leave it for now?

paulrobertlloyd:
The order of the status summary rows should match the order of the filters: consent, triage, programme.

app/components/app_patient_search_result_card_component.rb
```diff
@@ -68,10 +82,34 @@ def call
...
+             outcome: outcome
            )
   end
+
+  def display_triage_status?
+    status = @patient.triage_status(programme: @programme).status
+
+    return true if @triage_status.present?
+
+    status == "required"
```
thomasleese, line 113:
```suggestion
    @patient.triage_status(programme: @programme).required?
```

jhenderson: That causes the tests to fail but we could do this:

```
return true if @triage_status.present?

@patient.triage_status(programme: @programme).required?
```

PR #3579, opened 2025-05-21
Create ADR for autoscaling decisions

TheOneFromNorway:
- Useful to document in repository why we have made specific decisions


adr/00010-autoscaling-for-ecs.md
```diff
@@ -0,0 +1,36 @@
+# 10. Autoscaling for ECS
+
+Date: 2025-05-21
+
+## Status
+
+In progress
```
thomasleese, line 7: Should this be `Accepted` to say that we've decided to do it?

thomasleese (approved):

PR #3582, opened 2025-05-21
Ensure dates are deserialized with time zone

thomasleese:
There seems to be an issue when dates are stored in the browser session where they will come back with the time zone information correctly, but when stored in a model object that calls `clear_changes_information` the time zone is stripped and we end up with a date time object in the server time. This is a problem in our deployed environments where the server time is UTC, and in daylight saving the users will see a date that is 1 hour out.

We need `clear_changes_information` as we use that in a number of places to render appropriate yellow tags when values have changed, but in this case when the model object is being first instanced, we don't want to display these as changes.

[Jira Issue](https://nhsd-jira.digital.nhs.uk/browse/MAV-1002)

### DIFF
--- a/spec/models/concerns/request_session_persistable_spec.rb
+++ b/spec/models/concerns/request_session_persistable_spec.rb
@@ -9,9 +9,10 @@ describe RequestSessionPersistable do
         "key"
       end
 
-      attribute :value, :string
+      attribute :string, :string
+      attribute :datetime, :datetime
 
-      validates :value, presence: true, on: :update
+      validates :string, presence: true, on: :update
 
       def reset_unused_fields
       end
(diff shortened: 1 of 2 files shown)


spec/models/concerns/request_session_persistable_spec.rb
```diff
@@ -23,6 +24,32 @@ def reset_unused_fields
...
+        expect(model.changed?).to be(false)
+      end
+
+      it "parses and keeps the time zone" do
+        expect(model.datetime).to eq(Time.zone.local(2025, 5, 21, 11, 48, 17))
+        expect(model.datetime).to be_a(ActiveSupport::TimeWithZone)
+        expect(model.datetime.time_zone.name).to eq("London")
+      end
+
+      context "when system timezone is UTC" do
```
jhenderson, line 41: When I removed the `around` block, the example still passed. I guess `ClimateControl.modify` is not affecting the timezone configuration.

thomasleese: Hmm I was worried that might be the case, it seems like the Ruby process needs to restart to pick up the latest `TZ`. In GitHub Actions we run the tests in UTC `TZ` so we are getting coverage, but it makes it tricky to test locally.

PR #3588, opened 2025-05-22
Accept import capitalisation differences during import

alistair-white-horne-tng:
Names which only differ by capitalisation (eg "SMITH" vs "Smith") currently require a manual review. This has caused many (100s) manual reviews per import for C&W. This is made even more confusing by the fact that Mavis capitalises all surnames in the front end, so the "change" looked identical to the users.

This should probably be automated, accepting the incoming capitalisation.

[MAV-1084](https://nhsd-jira.digital.nhs.uk/browse/MAV-1084)

### DIFF
--- a/app/models/concerns/pending_changes_concern.rb
+++ b/app/models/concerns/pending_changes_concern.rb
@@ -6,31 +6,16 @@ module PendingChangesConcern
   included { attribute :pending_changes, :jsonb, default: {} }
 
   def stage_changes(attributes)
-    need_save = false
+    attributes.each do |attr, new_value|
+      current_value = public_send(attr)
 
-    new_pending_changes =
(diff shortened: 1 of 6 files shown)


thomasleese (approved):

app/models/concerns/pending_changes_concern.rb
```diff
@@ -6,15 +6,18 @@ module PendingChangesConcern
...
+        pending_changes.delete(attr.to_s)
+      else
+        pending_changes[attr.to_s] = new_value
       end
-
-    if new_pending_changes.any?
-      self.pending_changes = pending_changes.merge(new_pending_changes)
     end
+
+    save!
```
alistair-white-horne-tng, line 20: @thomasleese, based on the fact @jhenderson and I recently removed the `.update!` from this line, do we now want to get rid of this `save!` too?

PR #3591, opened 2025-05-23
Look up NHS numbers for pending changes

thomasleese:
This ensures that when uploading twins, the second patient will have an NHS number in the pending changes and the users will see clearly that it is two different parents and both need to be kept.

[Jira Ticket - MAV-273](https://nhsd-jira.digital.nhs.uk/browse/MAV-273)


app/lib/update_patients_from_pds.rb
```diff
@@ -21,14 +23,26 @@ def call
           PatientNHSNumberLookupJob.set(
             priority:,
             queue:,
-            wait: index * wait_between_jobs
+            wait: jobs_queued * wait_between_jobs
```
jhenderson, line 26: How large do you think `patients` could be? If the collection was large, later jobs could have very significant delays.

app/jobs/patient_nhs_number_lookup_with_pending_changes_job.rb
```diff
@@ -0,0 +1,36 @@
...
+    patient_with_pending_changes = patient.with_pending_changes
+    patient_with_pending_changes.nhs_number =
+      nil unless patient.nhs_number_changed?
+
+    return unless patient_with_pending_changes.changed?
+
+    if patient_with_pending_changes.nhs_number.present? &&
+         !patient_with_pending_changes.invalidated?
+      return
+    end
```
misaka, line 18: The goal of this bit is really unclear to me. Is it just me? Would it be worth adding a comment just to make it a bit clearer what it is doing? Or maybe put it behind a named method?

app/lib/update_patients_from_pds.rb
misaka: I agree we need to think about this, but maybe we just need to deal with this until we can get Sidekiq in place, which I'd like to start on really soon.

thomasleese: While I agree we have this problem, I don't think this change makes it significantly worse. Most parents _don't_ have pending changes, and therefore there won't be that many additional jobs queued.

app/jobs/patient_nhs_number_lookup_with_pending_changes_job.rb
thomasleese: Yeah agreed. It's basically [doing this logic](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/6c3ffb95cb7fe63d4b4d8bf5010c01496f828920/app/jobs/patient_nhs_number_lookup_job.rb#L9) but for the patient with pending changes, but I will make it clearer.

jhenderson (approved):

thomasleese:
> @thomasleese @misaka @murugapl how does the current PDS work affect this change? Is it worth getting this one over the line e.g. in v3.2.0?

I've been wondering about this, I actually think the PDS work makes this particular PR redundant, as we'll be looking up NHS numbers prior to importing them, so this problem goes away (whereas now the matching is happening before the PDS lookup).

PR #3592, opened 2025-05-23
Don't set session when recording already had vaccinations

thomasleese:
The session of a vaccination record is used to determine the location where the vaccination took place. When recording an "already had" vaccination, we don't know where the vaccination took place so it doesn't make sense to store a location


thomasleese:
#3551 is a better alternative to this, we'll continue to associate the records with the session they were recorded in, but their `location` will be `nil`.

PR #3602, opened 2025-05-27
Add basic API endpoints to be used in testing

thomasleese:
This adds two new API endpoints which for now will be only used by the functional tests:

- `GET /api/locations` - Fetch locations that the service knows about.
- `POST /api/onboard` - Onboard a new organisation.

The idea behind this is to allow the tests to set up a fresh organisation each time they are run (by fetching a number of schools using the `/api/locations` endpoint and then using `/api/onboard` to create the organisation). By using a fresh organisation each time we avoid issues where multiple parallel test runs conflict with each other.

### DIFF
--- a/spec/requests/api/onboard_spec.rb
+++ b/spec/requests/api/onboard_spec.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+describe "/api/onboard" do
+  let(:config) { YAML.safe_load(file_fixture(filename).read) }
+
+  describe "POST" do
+    subject(:request) do
+      post "/api/onboard",
+           params: config.to_json,
+           headers: {
+             "Content-Type" => "application/json"
+           }
+    end
+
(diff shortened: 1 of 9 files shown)


spec/requests/api/onboard_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      before do
+        create(:programme, :hpv)
+        create(:school, :secondary, :open, urn: "123456")
+        create(:school, :secondary, :open, urn: "234567")
+        create(:school, :secondary, :open, urn: "345678")
+        create(:school, :secondary, :open, urn: "456789")
+      end
+
+      it "responds wth created" do
```
jhenderson, line 26: Typo

```suggestion
      it "responds with created" do
```

misaka (approved):

misaka:
🚢

PR #3606, opened 2025-05-27
Display version of Mavis in the footer

murugapl:
This change ensures that the current git ref or branch name used for deployment is set as the APP_VERSION environment variable. The application’s footer will display this value, allowing users to easily identify which version of the code is currently running.

For local development, the current branch name is used as the APP_VERSION.


config/initializers/version.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+path = Rails.root.join("public/ref")
+version = File.exist?(path) ? File.read(path).strip : nil
```
thomasleese, line 4: I wonder if here we should run the code in `bin/setup` if the file doesn't exist?

Usually I run `bin/setup` once when setting up the project, but then I'll run `bin/dev` so the file `public/ref` will never be refreshed.

thomasleese (commented):
Looks good, just one small comment.

thomasleese (approved):
Nice!

PR #3607, opened 2025-05-27
Remove unnecessary update! call in PendingChangesConcern

jhenderson:
Since records using this mixin will be bulk imported later in the process, using `update!` would trigger an immediate database write, validation checks, and callbacks for each record, negatively impacting performance without providing any benefit.

### DIFF
--- a/spec/models/concerns/pending_changes_concern_spec.rb
+++ b/spec/models/concerns/pending_changes_concern_spec.rb
@@ -34,12 +34,6 @@ describe PendingChangesConcern do
       expect(model.pending_changes).to eq({ "family_name" => "Smith" })
     end
 
-    it "updates the pending_changes attribute" do
-      expect { model.stage_changes(given_name: "Jane") }.to change {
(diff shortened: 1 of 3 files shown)


thomasleese:
@simonnorris1 is going to run a performance test of the importing to get a baseline before we merge this change in.

spec/models/concerns/pending_changes_concern_spec.rb
```diff
@@ -34,12 +34,6 @@
       expect(model.pending_changes).to eq({ "family_name" => "Smith" })
     end
 
-    it "updates the pending_changes attribute" do
-      expect { model.stage_changes(given_name: "Jane") }.to change {
-        model.reload.pending_changes
```
thomasleese, line 39: Would removing the call to `.reload` mean we can keep these tests?

thomasleese (approved):

simonnorris1:
image

PR #3608, opened 2025-05-27
Fix: prevent incorrect data merging for twins during patient import

jhenderson:
When importing records for twins or siblings with identical identifying information, the system was incorrectly attributing data from one twin to another when handling duplicate records. This resulted in critical personal identifiers (gender, registration, preferred names) being erroneously merged.

This fix ensures that when pending changes require review, any auto-accepted changes are properly moved to the `pending_changes` queue instead of being directly applied.

[JIRA Ticket](https://nhsd-jira.digital.nhs.uk/browse/MAV-1214)


app/models/patient_import_row.rb
```diff
@@ -222,19 +168,107 @@ def nhs_number_value
...
+    # This ensures all changes are reviewed together rather than having
+    # some changes applied immediately while others await review. This
+    # is particularly critical when handling potential duplicates like twins,
+    # where auto-accepting some changes could lead to data from one twin being
+    # incorrectly applied to another twin's record.
+    if patient.pending_changes.any?
+      auto_accepted_changes.each do |attribute, (old_value, new_value)|
+        patient[attribute] = old_value
+        patient.pending_changes[attribute] = new_value
+      end
```
thomasleese, line 231:
```suggestion
      patient.restore_attributes(auto_accepted_changes)
      patient.pending_changes.merge!(auto_accepted_changes)
```

thomasleese (commented):
Just one small suggestion about the reverting logic (using existing Rails methods), otherwise looks good.

app/models/patient_import_row.rb
jhenderson: Turns out this didn't work. We need to make some adjustments given `patient.changes` returns something like `{"given_name" => ["Fabiola", "Pete"]}`. I'm leaning towards the loop now, but happy either way. 😄 

```
new_values = auto_accepted_changes.transform_values { |values| values.last }
patient.restore_attributes(auto_accepted_changes.keys)
patient.pending_changes.merge!(new_values)
```

thomasleese (approved):

PR #3622, opened 2025-05-30
Add rake task to transfer school data

murugapl:
Added a new rake task schools:roll_over that updates the location and school IDs for sessions, patients, and consent forms when rolling over a school to a new one, in the event the old one has been closed and reopened under a new name and URN for example. The task takes two arguments: old_urn and new_urn, which are the URNs of the old and new schools, respectively.

https://nhsd-jira.digital.nhs.uk/browse/MAV-1229

### DIFF
--- a/lib/tasks/schools.rake
+++ b/lib/tasks/schools.rake
@@ -212,4 +212,30 @@ namespace :schools do
       year_groups: [8, 9, 10, 11]
     )
   end
+
+  desc "Transfer child records from one school to another."
+  task :move_patients, %i[old_urn new_urn] => :environment do |_task, args|
+    old_loc = Location.school.find_by(urn: args[:old_urn])
+    new_loc = Location.school.find_by(urn: args[:new_urn])
+
+    raise "Could not find one or both schools." if old_loc.nil? || new_loc.nil?
+
+    if !new_loc.team_id.nil? && new_loc.team_id != old_loc.team_id
+      raise "#{new_loc.urn} belongs to #{new_loc.team.name}. Could not complete transfer."
+    end
+    new_loc.update!(team: old_loc.team)
+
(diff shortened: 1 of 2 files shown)


thomasleese (approved):

lib/tasks/schools.rake
```diff
@@ -212,4 +212,19 @@ namespace :schools do
...
+
+    raise "Could not find one or both schools." if old_loc.nil? || new_loc.nil?
+
+    Session.where(location_id: old_loc.id).update_all(location_id: new_loc.id)
+    Patient.where(school_id: old_loc.id).update_all(school_id: new_loc.id)
+    ConsentForm.where(location_id: old_loc.id).update_all(
+      location_id: new_loc.id
+    )
+    ConsentForm.where(school_id: old_loc.id).update_all(school_id: new_loc.id)
+  end
```
thomasleese, line 229: I think we may also want to update any outstanding `SchoolMove`s.

spec/lib/tasks/schools_rake_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+describe "schools:move_patients" do
+  def capture_rake_task_output(task_name, *args)
+    stdout = StringIO.new
+    $stdout = stdout
+    Rake::Task[task_name].invoke(*args)
+    $stdout = STDOUT
+    Rake.application[task_name].reenable
+    stdout.string
+  end
```
thomasleese, line 11: It doesn't look like this output is being tested anywhere, so I think we can simplify this and instead call `Rake::Task["schools:move_patients"].invoke` directly, similar to https://github.com/nhsuk/manage-vaccinations-in-schools/blob/24f2fa48d3003e0977c47d37ddb0ae7edbf1222b/spec/lib/tasks/status_update_spec.rb#L5

thomasleese (approved):

PR #3652, opened 2025-06-04
Fix long batch name

MartinVanIJcken:
This PR fixes MAV-1240

Currently long batch names cause the table to look weird. This was not really an issue as we do not have batch names over length 6 on prod. Regardless, a ticket was opened and so here is a fix.

Using the css-style, batch names are now allowed to wrap inside of the name, allowing the table with to remain limited.

### DIFF
--- a/app/views/vaccines/index.html.erb
+++ b/app/views/vaccines/index.html.erb
@@ -27,7 +27,9 @@
           <% batches.each do |batch| %>
             <% body.with_row do |row| %>
               <% row.with_cell(text: batch.name) do %>
-                <%= batch.name %>
+                <span class="nhsuk-u-text-break-word">
+                  <%= batch.name %>
+                </span>
                 <% if batch.id == @todays_batch_id_by_programme[vaccine.programme] %>
                   <br>
                   <span class="nhsuk-caption-m">


MartinVanIJcken:
> I also wonder if the approach should be to restrict batch names to a length of about 10, but that doesn't mean we can't make this change as well.

That would be possible but I fail to see who benefits from that change. 

> I also wondered about using `<code>` or `\<pre\>` to mark these up, but not sure.

I would oppose the use of <code>. Using HTML tags not in accordance with their semantic meaning can mess up the accessibility of a webpage. As for `\<pre\>`, it does not support the word-break css property so again I would be hesistant to do this.

thomasleese:
> I would oppose the use of `<code>`. Using HTML tags not in accordance with their semantic meaning can mess up the accessibility of a webpage. As for `\<pre\>`, it does not support the word-break css property so again I would be hesitant to do this.

In this case I actually imagine the accessibility might be improved by marking the batch number with something other than `<span>`, as it's not a word so it can't be read out properly by screen readers. I was imagining `<code>` might allow it to be read out as a string of letters and numbers, but I could be wrong.

thomasleese (approved):

PR #3653, opened 2025-06-04
Remove ? and _ from NHS number on import

MartinVanIJcken:
These characters maybe unintentionally introduced when copying an NHS number from our app into excel using an encoding that does not support zero-width-joiner characters. They should be removed again so that the NHS number can be properly parsed.

https://nhsd-jira.digital.nhs.uk/browse/MAV-1215

### DIFF
--- a/app/models/immunisation_import_row.rb
+++ b/app/models/immunisation_import_row.rb
@@ -432,7 +432,7 @@ class ImmunisationImportRow
   end
 
   def patient_nhs_number_value
-    patient_nhs_number&.to_s&.gsub(/\s/, "")
+    patient_nhs_number&.to_s&.gsub(/\s|\?|_/, "")
   end
 
   def reason_not_administered_value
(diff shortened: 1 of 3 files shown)


thomasleese (commented):
The change looks good to me. Could we add some unit tests for this scenario? Probably [something like this](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/3fb1667edb448c61330cf46d8fc6a8879e34ba65/spec/models/immunisation_import_row_spec.rb#L1763-L1777) but for NHS number instead of date of birth.

I'm also wondering about whether this is the right place for the code, or whether it should go in the model normalisation which would mean it also covers the scenario of a nurse manually pasting an NHS number with these characters in to the text box when editing a patient?

https://github.com/nhsuk/manage-vaccinations-in-schools/blob/3fb1667edb448c61330cf46d8fc6a8879e34ba65/app/models/patient.rb#L189-L192

PR #3661, opened 2025-06-05
Fix version number display in production

murugapl:
The previous implementation to display the Mavis version number in the footer of the app used the deploy git ref to create a file with the version number which was baked into the image. With the current deploy strategy, the image is only built during deploys to QA and test, before the version tag is even created. So the version number isn't displayed appropriately in production.

This change uses environment variables to set the version number instead. The MAVIS_APP_VERSION environment variable is now set during infrastructure deployment and passed to the ECS containers. For production deployments, the version format is validated to ensure it follows semantic versioning (v1.2.3 format).

This ensures the correct version number is displayed regardless of when the Docker image was built, as the version is determined at deployment time rather than build time.

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -20,6 +20,10 @@ on:
         description: Docker image tag to deploy
         required: false
         type: string
+      git_ref_to_deploy:
+        required: true
+        type: string
+
   workflow_call:
     inputs:
       environment:
@@ -28,6 +32,9 @@ on:
         type: string
       image_tag:
         required: false
+        type: string
+      git_ref_to_deploy:
+        required: true
         type: string
 
 permissions: {}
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy-infrastructure.yml
```diff
@@ -135,12 +142,17 @@ jobs:
         run: sudo snap install --classic aws-cli
       - name: Check if any deployments are running
         run: ../scripts/check-for-running-deployments.sh ${{ inputs.environment }}
+      - name: Set Mavis app version
+        run: |
+          GIT_REF_TO_DEPLOY="${{ inputs.git_ref_to_deploy }}"
+          echo "GIT_REF_TO_DEPLOY=$GIT_REF_TO_DEPLOY" >> $GITHUB_ENV
```
thomasleese, line 148: I don't think we're using this environment variable anywhere, only `MAVIS_APP_VERSION`.

terraform/app/variables.tf
```diff
@@ -202,6 +208,10 @@ locals {
     {
       name  = "MAVIS__SPLUNK__ENABLED"
       value = var.enable_splunk ? "true" : "false"
+    },
+    {
+      name  = "MAVIS_APP_VERSION" # <-- ADD THIS
```
thomasleese, line 213:
```suggestion
      name  = "APP_VERSION" # <-- ADD THIS
```

The only reason we have `MAVIS__` prefixed on the others is due to the way we load in the `Settings`. In general I think we can avoid the `MAVIS` prefix until it's strictly necessary.

thomasleese (commented):
Small comments on the naming, but otherwise looks good.

terraform/app/variables.tf
```diff
@@ -202,6 +208,10 @@ locals {
     {
       name  = "MAVIS__SPLUNK__ENABLED"
       value = var.enable_splunk ? "true" : "false"
+    },
+    {
+      name  = "APP_VERSION" # <-- ADD THIS
```
thomasleese, line 213:
```suggestion
      name  = "APP_VERSION"
```

thomasleese (approved):

PR #3664, opened 2025-06-05
Removing case sensitivity in validation check of clinic name

KarimMestekawy:
This ensures that the validation check on the clinic name is not case sensitive when importing vaccination records for community clinics.

https://nhsd-jira.digital.nhs.uk/browse/MAV-691

### DIFF
--- a/app/models/immunisation_import_row.rb
+++ b/app/models/immunisation_import_row.rb
@@ -525,7 +525,7 @@ class ImmunisationImportRow
           "is greater than #{MAX_FIELD_LENGTH} characters long"
         )
       elsif clinic_name_required &&
-            !organisation.community_clinics.exists?(name: clinic_name.to_s)
+            !organisation.community_clinics.any? { |clinic| clinic.name.downcase == clinic_name.to_s.downcase }
         errors.add(clinic_name.header, "Enter a clinic name")
       end
     elsif clinic_name_required


app/models/immunisation_import_row.rb
```diff
@@ -525,7 +525,7 @@ def validate_clinic_name
           "is greater than #{MAX_FIELD_LENGTH} characters long"
         )
       elsif clinic_name_required &&
-            !organisation.community_clinics.exists?(name: clinic_name.to_s)
+            !organisation.community_clinics.any? { |clinic| clinic.name.downcase == clinic_name.to_s.downcase }
```
thomasleese, line 528:
```suggestion
            !organisation.community_clinics.exists?("LOWER(name) = ?", clinic_name.to_s.downcase)
```

This should have the same behaviour but it's a more optimised query. In the original change, Rails will end up fetching all the community clinic records and creating an instance of them in memory to then call `.any?` on, whereas using `exists?` means we'll execute a single `SELECT EXISTS` SQL query.

thomasleese (commented):
Thanks for the PR, I've just left a comment on the approach taken to implement the fix.

I think it would also be good to add some unit tests for this, perhaps something in this block: https://github.com/nhsuk/manage-vaccinations-in-schools/blob/744c2d57e8d8738c7b4d71f74625c402e7011334/spec/models/immunisation_import_row_spec.rb#L1150

spec/models/immunisation_import_row_spec.rb
```diff
@@ -1352,6 +1352,69 @@
 
         it { should eq("A Clinic") }
       end
+
+      context "with an existing community clinic" do
```
thomasleese, line 1356: These tests look good, but I think they're in the wrong section. Since these are concerned with testing the validations they belong in the `describe "validations"` section above, rather than here which is testing the `location_name` method.

spec/models/immunisation_import_row_spec.rb
```diff
@@ -1352,6 +1352,69 @@
...
+
+        let(:session) { create(:session, organisation:, programmes:) }
+
+        before { create(:community_clinic, name: "A clinic", organisation:) }
+
+        it "is matching" do
+          immunisation_import_row.valid?
+          expect(immunisation_import_row.errors.full_messages).not_to include(
+            a_string_including("Enter a clinic name")
+          )
```
thomasleese, line 1373:
```suggestion
          expect(immunisation_import_row).to be_valid
```

spec/models/immunisation_import_row_spec.rb
```diff
@@ -1352,6 +1352,69 @@
...
+
+        let(:session) { create(:session, organisation:, programmes:) }
+
+        before { create(:community_clinic, name: "A clinic", organisation:) }
+
+        it "is case insensitive" do
+          immunisation_import_row.valid?
+          expect(immunisation_import_row.errors.full_messages).not_to include(
+            a_string_including("Enter a clinic name")
+          )
```
thomasleese, line 1394:
```suggestion
          expect(immunisation_import_row).to be_valid
```

spec/models/immunisation_import_row_spec.rb
```diff
@@ -1352,6 +1352,69 @@
...
+
+        let(:session) { create(:session, organisation:, programmes:) }
+
+        before { create(:community_clinic, name: "A clinic", organisation:) }
+
+        it "has errors" do
+          expect(immunisation_import_row).to be_invalid
+          expect(immunisation_import_row.errors.full_messages).to include(
+            a_string_including("Enter a clinic name")
+          )
```
thomasleese, line 1415: To be sure the error is attached to the right field, I would suggest something like:

```suggestion
          expect(immunisation_import_row.errors["CLINIC_NAME"]).to contain_exactly("Enter a clinic name")
```

spec/models/immunisation_import_row_spec.rb
```diff
@@ -1352,6 +1352,69 @@
...
+          valid_data.merge(
+            "CLINIC_NAME" => "A clinic",
+            "CARE_SETTING" => "2",
+            "SESSION_ID" => session.id.to_s
+          )
+        end
+
+        let(:session) { create(:session, organisation:, programmes:) }
+
+        before { create(:community_clinic, name: "A clinic", organisation:) }
```
thomasleese, line 1367: These two lines are the same across all 3 of these tests, so it can be moved to a higher level outside the context, particular if combined with my comment above about moving these tests in to the validation section.

KarimMestekawy: Simplified it

spec/models/immunisation_import_row_spec.rb
KarimMestekawy: I initially had it like this because there were error messages from other validation checks, but I now have the full correct mock data and simplified it as you stated

spec/models/immunisation_import_row_spec.rb
```diff
@@ -782,6 +782,51 @@
...
+      let(:session) { create(:session, organisation:, programmes:) }
+
+      before { create(:community_clinic, name: "A clinic", organisation:) }
+
+      context "with an existing community clinic" do
+        let(:data) { valid_clinic_data.merge("CLINIC_NAME" => "A clinic") }
+
+        it "is matching" do
+          expect(immunisation_import_row).to be_valid
+        end
```
thomasleese, line 806:
```suggestion
        it { should be_valid }
```

Because the `subject` is set to `immunisation_import_row`, we can simplify this code even further. However, it may not always be appropriate if we want to call out something specific in the naming of the `it`, so I don't mind either way.

spec/models/immunisation_import_row_spec.rb
KarimMestekawy: Moved

spec/models/immunisation_import_row_spec.rb
KarimMestekawy: Removed the subject and still have it more verbose to be consistent with other 'validation' unit tests in the block

thomasleese (approved):

PR #3666, opened 2025-06-05
Update use of variable

TheOneFromNorway:
- Variable was originally inserted on main but was accidnetally removed during merge conflict
- Add a default value to the variable to simplify local development


terraform/app/variables.tf
```diff
@@ -160,6 +160,7 @@ variable "enable_splunk" {
 variable "app_version" {
   type        = string
   description = "The version identifier for the MAVIS application deployment"
+  default     = "NA"
```
thomasleese, line 163:
```suggestion
  default     = "Unknown"
```

Just thinking about what users might see in this case.

misaka (approved):

terraform/app/variables.tf
thomasleese: I think `Not available` is at least better than `NA` in that case.

misaka: I agree with @thomasleese, if this is for user consumption then words are better than just a couple letters.

thomasleese (approved):

PR #3677, opened 2025-06-06
Ensure duplicate patients aren't shown

thomasleese:
When filtering on patients that an organisation is allowed to see, using a left outer join means that each patient will be returned for each school move, session or vaccination record they have. This means duplicate patients are shown in various parts of the UI.

Instead, we can use an `EXISTS` query, which is actually more performant as it can short-circuit the conditionals, and avoids duplicate patients being shown.


jhenderson (approved):
Nice refactor.

PR #3682, opened 2025-06-09
Add Flu vaccination method question to online consent

thomasleese:
When asking parents if they consent to the Flu vaccination, we need to ask which vaccination method they consent to, whether nasal spray or injection. This updates the decision question of the online consent journey to include this, matching the latest designs in the prototype.

[Jira Issue - MAV-1235](https://nhsd-jira.digital.nhs.uk/browse/MAV-1235)

## Screenshot

Screenshot 2025-06-09 at 09 46 10


app/views/parent_interface/consent_forms/start/_hpv.html.erb
```diff
@@ -0,0 +1,18 @@
+<h2 class="nhsuk-heading-m">
+  HPV
+</h2>
```
misaka, line 3: I don't see this heading on the prototype ... I was just checking because I noticed there was no "Flu" heading but there are these heading on the others. Do we need this `h2` on the non-flu start pages?

misaka: Does this mean Flu should have an `h2` then?

thomasleese: I decided not to add it because there isn't one in the prototype, the refactor should keep the other 3 programmes working the same as they do now, while adding the new start page content for Flu.

paulrobertlloyd: For the avoidance of doubt, here are the 3 starts pages:

Screenshot of start page for HPV.

Screenshot of start page for doubles.

Screenshot of start page for flu.

* HPV doesn’t show a `h2` heading
* Doubles shows a `h2` heading above each programme
* Flu includes a `h2` heading within its description

They are basically each different, in their own way, sorry! We might want to review how we show content for each type on the start page, given this variance.

PR #3687, opened 2025-06-09
Add GIAS download command to deployed envs

misaka:
The rake task to download the GIAS file is only available in dev environments because the `mechanize` gem isn't available in the `production` env. This change installs that gem in `production`, allowing the download to run.

It also moves the `schools:download` Rake task to the mavis cli as a simple lift-and-shift, I don't think adding tests or cmdline options adds much, but it's nice to have it in it's own file and available through the mavis cli.

[MAV-1257](https://nhsd-jira.digital.nhs.uk/browse/MAV-1257)


thomasleese (approved):

thomasleese (approved):
Tested locally it seems to work.

PR #3689, opened 2025-06-09
Remove unique index constraint from sessions

jhenderson:
As part of the upcoming flu vaccination feature, we need to support multiple sessions at the same location within the same academic year. This requires removing the unique composite index that previously enforced a one-to-one relationship between locations and sessions.

For query optimisation:
- Added a composite index on `[organisation_id, location_id]` which:
  - Supports queries filtering by both columns together
  - Supports queries filtering by `organisation_id` alone (leftmost prefix rule)
- Added a separate index on `location_id` for queries that filter by location alone (since `location_id` isn't the leftmost column in the composite index)

https://nhsd-jira.digital.nhs.uk/browse/MAV-1276

### DIFF
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -17,7 +17,8 @@
 #
 # Indexes
 #
-#  idx_on_organisation_id_location_id_academic_year_3496b72d0c  (organisation_id,location_id,academic_year) UNIQUE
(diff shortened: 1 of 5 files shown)


thomasleese (approved):
Change looks good, I'll write a Jira ticket for this and attach it so we can include it in a release.

PR #3700, opened 2025-06-10
Removing the addition of a deceased child to clinics and sessions

KarimMestekawy:
This ensures that a child with a date of death is not added to sessions when included in the class list upload

https://nhsd-jira.digital.nhs.uk/browse/MAV-1228


app/models/patient_import_row.rb
```diff
@@ -40,8 +40,12 @@ def to_patient
...
 
   def to_school_move(patient)
-    if patient.new_record? || patient.school != school ||
-         patient.home_educated != home_educated || patient.not_in_organisation?
+    if !patient.deceased? &&
+         (
+           patient.new_record? || patient.school != school ||
+             patient.home_educated != home_educated ||
+             patient.not_in_organisation?
+         )
```
thomasleese, line 48: I'm finding the condition in this `if` statement quite big and harder to read at a glance now, and I wonder if it's worth refactoring to use an early return at the top of the method body, for example:

```rb
return if patient.deceased?
```

We could possibly even do the same with the conditions that we have already.

```rb
return unless patient.new_record? || patient.school != school ||
  patient.home_educated != home_educated || patient.not_in_organisation?
```

What do you think?

thomasleese (approved):

PR #3709, opened 2025-06-11
Add date input to school transfer rake task

murugapl:
The schools:move_patients task now has a change_date to indicate the date after which all records should be amended with the new school. This makes the task more flexible if to be used with schools that have records on Mavis before and after a change to urn.

### DIFF
--- a/lib/tasks/schools.rake
+++ b/lib/tasks/schools.rake
@@ -214,7 +214,8 @@ namespace :schools do
   end
 
   desc "Transfer child records from one school to another."
-  task :move_patients, %i[old_urn new_urn] => :environment do |_task, args|
+  task :move_patients,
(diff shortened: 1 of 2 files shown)


lib/tasks/schools.rake
```diff
@@ -225,12 +226,24 @@ namespace :schools do
     end
     new_loc.update!(team: old_loc.team)
 
+    change_date =
+      begin
+        Date.parse(args[:change_date])
+      rescue StandardError
+        nil
+      end
+    raise "Invalid date format. Use YYYY-MM-DD." if change_date.nil?
```
thomasleese, line 235: Can this go in to the `rescue` block above?

thomasleese:
Closing this as we determined there's more work needed, specifically around how we handle sessions if an school closes in the middle. We can re-open when we're ready to pick this up again.

PR #3720, opened 2025-06-12
Adding school URN alongside school names to the dropdown in the class list import screen

KarimMestekawy:
This ensures that the dropdown list includes the school names and URN to avoid any confusion with duplicate school names when MaVIS expands to more schools and regions

https://nhsd-jira.digital.nhs.uk/browse/MAV-1131


thomasleese:
Closing this as the ticket has been changed to add the address rather than the URNs.

PR #3730, opened 2025-06-16
Fix issues with duplicate records, parent updates and incorrect school moves

jhenderson:
**Issue 1: correct parent contact assignment during twin record resolution**

This commit fixes an issue where parent contacts could be incorrectly assigned to patient records under specific circumstances during duplicate resolution.

The bug occurs when:
1. A patient record has an import issue raised against it (typically for twins/near-duplicates)
2. The second record contains different parent contacts (usually from school class list uploads)
3. The import issue is resolved with "keep both records" or "keep previously uploaded record"
4. Parent contacts differ between the original and duplicate records

**Issue 2: prevent false school moves when importing twins at different schools**

When importing records for twins without NHS numbers who are enrolled at different schools, MAVIS mistakenly generates school moves for both children. Although each twin is uploaded correctly to their respective school, selecting "Keep both records" during duplicate handling results in both being shown at the same school and triggers inaccurate school move entries.


app/forms/import_duplicate_form.rb
```diff
@@ -49,6 +49,26 @@ def discard_pending_changes!
   end
 
   def keep_both_changes!
-    object.apply_pending_changes_to_new_record! if can_keep_both?
+    if can_keep_both?
```
thomasleese, line 52: If the user chooses to accept the pending changes, I think we would also need to apply this logic.

thomasleese (commented):
Overall looks good, just left one comment about when we apply the logic from pending changes.

I've got some minor comments about the names. For `FamilyRelationshipMapping`, although we've named the model "Parent", it's a bit of a misnomer because it can also contain guardians and other non-family relationships, and I wonder if the name `FamilyRelationshipMapping` makes this worse. Would `ParentRelationshipMapper` work instead? (having the same name as the model)

I also wondered about the `Mapper` suffix and think maybe `Factory` might be better suited, but on this one I don't mind either way.

thomasleese:
@jhenderson Looks like this might need rebasing.

PR #3782, opened 2025-06-19
Allow sessions to not have registration

jhenderson:
This PR introduces a new boolean column `requires_registration` to the `Session` model, allowing us to control whether registration is required for a specific session. When registration is not required, the "Register" tab will be hidden, but users can still access all patients under the "Record vaccinations" tab.

JIRA: [MAV-1292](https://nhsd-jira.digital.nhs.uk/browse/MAV-1292)

### DIFF
--- a/app/components/app_patient_session_record_component.rb
+++ b/app/components/app_patient_session_record_component.rb
@@ -21,7 +21,8 @@ class AppPatientSessionRecordComponent < ViewComponent::Base
     patient.consent_given_and_safe_to_vaccinate?(programme:, academic_year:) &&
       (
         patient_session.registration_status&.attending? ||
-          patient_session.registration_status&.completed? || false
+          patient_session.registration_status&.completed? ||
+          !patient_session.session.requires_registration?
       )
   end
 
--- a/app/controllers/sessions/record_controller.rb
+++ b/app/controllers/sessions/record_controller.rb
@@ -17,13 +17,14 @@ class Sessions::RecordController < ApplicationController
 
   def show
     scope =
-      @session
-        .patient_sessions
-        .includes(
-          :latest_note,
-          patient: %i[consent_statuses triage_statuses vaccination_statuses]
-        )
-        .has_registration_status(%w[attending completed])
+      @session.patient_sessions.includes(
+        :latest_note,
+        patient: %i[consent_statuses triage_statuses vaccination_statuses]
+      )
+
+    if @session.requires_registration?
+      scope = scope.has_registration_status(%w[attending completed])
+    end
 
     patient_sessions =
       @form.apply(scope).consent_given_and_ready_to_vaccinate(
--- a/spec/features/hpv_vaccination_administered_spec.rb
+++ b/spec/features/hpv_vaccination_administered_spec.rb
@@ -72,6 +72,21 @@ describe "HPV vaccination" do
     then_i_see_a_success_message
   end
 
+  scenario "Administered without registration" do
+    given_registrations_are_not_required
+    and_i_am_signed_in
+
+    when_i_go_to_a_patient_that_is_ready_to_vaccinate
+    and_i_fill_in_pre_screening_questions
+    and_i_record_that_the_patient_has_been_vaccinated("Other")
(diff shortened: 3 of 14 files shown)


db/migrate/20250618160357_add_requires_registration_field_to_sessions.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddRequiresRegistrationFieldToSessions < ActiveRecord::Migration[8.0]
+  def change
+    add_column :sessions, :requires_registration, :boolean, default: true
```
thomasleese, line 5:
```suggestion
    add_column :sessions, :requires_registration, :boolean, default: true, null: false
```

app/controllers/sessions/record_controller.rb
```diff
@@ -22,7 +22,7 @@ def show
           patient: %i[consent_statuses triage_statuses vaccination_statuses]
         )
         .in_programmes(@session.programmes)
-        .has_registration_status(%w[attending completed])
+        .eligible_for_vaccination_recording
```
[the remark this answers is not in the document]
thomasleese: Yeah not sure about this name. You could have `ready_for_vaccination` but I think we've already got something like that elsewhere.

I'm also wondering if it would be better to have this logic in the controller rather than the model, the scope is only used in that one place and it's very specific to that page so unlikely to be reused.

app/controllers/sessions/record_controller.rb
```diff
@@ -15,14 +15,22 @@ class Sessions::RecordController < ApplicationController
   before_action :set_batches, except: :show
 
   def show
+    base_relation = PatientSession.joins(:session)
```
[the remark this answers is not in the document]
jhenderson: That does work! Good call. Why didn't I think of that! 🤦 

I guess I was trapped in a local minimum so to speak thinking it had to be the one or the other explicitly.

thomasleese (approved):

jhenderson:
> This looks good, I think there are a few other places where we will need to factor in whether or not the session requires registration:
> 
> * https://github.com/nhsuk/manage-vaccinations-in-schools/blob/528af921b8bdc4e9ed674b607bec1776491f9743/app/components/app_session_actions_component.rb#L88
> * https://github.com/nhsuk/manage-vaccinations-in-schools/blob/528af921b8bdc4e9ed674b607bec1776491f9743/app/views/patient_sessions/_header.html.erb#L35
> * https://github.com/nhsuk/manage-vaccinations-in-schools/blob/528af921b8bdc4e9ed674b607bec1776491f9743/app/components/app_patient_session_record_component.rb#L23-L24

Made those changes. Good spot. 👍

spec/features/hpv_vaccination_administered_spec.rb
```diff
@@ -72,6 +72,21 @@
     then_i_see_a_success_message
   end
 
+  scenario "Administered without registration" do
```
jhenderson, line 75: @thomasleese I opted to put the test here so I can reuse a good number of steps. What do you think? Still go ahead and create a separate file? I don't mind either way.

app/components/app_patient_session_record_component.rb
```diff
@@ -21,7 +21,8 @@ def render?
     patient.consent_given_and_safe_to_vaccinate?(programme:) &&
       (
         patient_session.registration_status&.attending? ||
-          patient_session.registration_status&.completed? || false
+          patient_session.registration_status&.completed? ||
+          !patient_session.session.requires_registration?
```
jhenderson, line 25: The logic was wrong initially as I didn't fully understand what the criteria was saying. Basically, if the session doesn't require registration, then the user should still be able to record a vaccination.

I also got rid of the `false`. I don't think it's needed as the guard operated `&` will give `nil` which is falsey anyway.

spec/features/hpv_vaccination_administered_spec.rb
thomasleese: I think this works!

PR #3798, opened 2025-06-20
Fix data replication deployment

bogsi17:
* Now that the DB uses a custom KMS key, the role for deploying data replication resources needs access to this key.
* To keep the restrictions restrictive, any actions that would disable the key are still denied

### DIFF
--- a/.github/workflows/data-replication-pipeline.yml
+++ b/.github/workflows/data-replication-pipeline.yml
@@ -71,12 +71,18 @@ jobs:
           if [ -z "${{ inputs.db_snapshot_arn }}" ]; then
               echo "No snapshot ARN provided, fetching the latest snapshot"
               SNAPSHOT_ARN=$(aws rds describe-db-cluster-snapshots \
-              --query 'DBClusterSnapshots[?contains(DBClusterSnapshotIdentifier, `${{ inputs.environment}}`)].[DBClusterSnapshotArn, SnapshotCreateTime]' \
+              --query "DBClusterSnapshots[?DBClusterIdentifier=='mavis-${{ inputs.environment }}'].[DBClusterSnapshotArn, SnapshotCreateTime]" \
               --output text | sort -k2 -r | head -n 1 | cut -f1)
+          
+              if [ -z "$SNAPSHOT_ARN" ]; then
+                  echo "No snapshots found for mavis-${{ inputs.environment }}"
+                  exit 1
+              fi
           else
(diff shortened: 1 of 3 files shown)


terraform/account/resources/iam_policy_DeployDataReplicationResources.json
```diff
@@ -38,6 +39,16 @@
         }
       }
     },
+    {
+      "Effect": "Deny",
+      "Action": [
+        "kms:Delete*",
+        "kms:ScheduleKeyDeletion",
+        "kms:Revoke*",
+        "kms:Disable*"
```
TheOneFromNorway, line 48: What about:
* kms:CancelKeyDeletion
* kms:CreateGrant
* kms:DisableKey
* kms:PutKeyPolicy
* kms:ReplicateKey
* kms:RetireGrant
* kms:RevokeGrant

.github/workflows/data-replication-pipeline.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           if [ -z "${{ inputs.db_snapshot_arn }}" ]; then
               echo "No snapshot ARN provided, fetching the latest snapshot"
               SNAPSHOT_ARN=$(aws rds describe-db-cluster-snapshots \
-              --query 'DBClusterSnapshots[?contains(DBClusterSnapshotIdentifier, `${{ inputs.environment}}`)].[DBClusterSnapshotArn, SnapshotCreateTime]' \
+              --query 'DBClusterSnapshots[?contains(DBClusterSnapshotIdentifier, `${{ inputs.environment}}`) && !contains(DBClusterSnapshotIdentifier, `data-replication`].[DBClusterSnapshotArn, SnapshotCreateTime]' \
```
TheOneFromNorway, line 74: Is it possible to filter by `DB instance or cluster` or `Instance/Cluster Name` and then pass the actual cluster name instead? That would be more futureproof

terraform/account/resources/iam_policy_DeployDataReplicationResources.json
bogsi17: `kms:CreateGrant` is needed to restore the cluster from a snapshot. I added the rest be denied

.github/workflows/data-replication-pipeline.yml
bogsi17: Yes that's possible. Small caveat: That now matches snapshots taken by the AWS Backup job. Those snapshots are tagged with the environment, so they needed to be excluded from the deny policy

PR #3799, opened 2025-06-22
Draft session dates to prevent save on "back"

MartinVanIJcken:
The "back" button on the session dates edit page was broken. Even when one clicked back, the changes made were still saved. MAV-516 includes a bug report for this.

By adding a new model which is a draft version of the session dates we can allow modification of session dates in the edit page without immediately saving them to the database.

### DIFF
--- a/app/controllers/session_dates_controller.rb
+++ b/app/controllers/session_dates_controller.rb
@@ -2,41 +2,33 @@
 
 class SessionDatesController < ApplicationController
   before_action :set_session
+  before_action :set_draft_session_dates
 
   def show
-    @session.session_dates.build if @session.session_dates.empty?
+    initialize_draft_from_session
+
+    if params[:add_another] ||
+         @draft_session_dates.session_dates_attributes.empty?
+      add_blank_session_date
+    end
   end
 
   def update
-    @session.assign_attributes(remove_invalid_dates(session_params))
-    @session.set_notification_dates
+    @draft_session_dates.session_dates_attributes =
+      fetch_draft_session_dates_from_form
 
-    render :show, status: :unprocessable_content and return if @session.invalid?
-
-    @session.save!
-
-    # If deleting dates, they don't disappear from `session.dates` until
-    # the model has been saved due to how `accepts_nested_attributes_for`
-    # works.
-    if any_destroyed?
-      @session.session_dates.reload
-      @session.set_notification_dates
-      @session.save!
-    end
-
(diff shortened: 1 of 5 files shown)


spec/features/draft_session_dates_e2e_spec.rb
```diff
@@ -0,0 +1,487 @@
...
+  def when_i_add_a_second_date
+    # Find the second list item (after clicking "Add another date")
+    within page.all(".app-add-another__list-item")[1] do
+      fill_in "Day", with: "16"
+      fill_in "Month", with: "3"
+      fill_in "Year", with: "2024"
+    end
+  end
+
+  def when_i_add_a_date
```
misaka, line 232: This method and the one below are never used. Did you mean to keep them or can we remove them?

spec/features/draft_session_dates_e2e_spec.rb
```diff
@@ -0,0 +1,487 @@
...
+describe "Draft session dates end-to-end functionality" do
+  around { |example| travel_to(Time.zone.local(2024, 2, 18)) { example.run } }
+
+  scenario "User can modify session dates without immediate saving" do
+    given_i_have_a_session_with_existing_dates
+    when_i_visit_the_session_edit_page
+    then_i_see_the_existing_session_dates
+
+    when_i_click_change_session_dates
+    then_i_see_the_draft_session_dates_page_with_existing_dates
```
misaka, line 12: Similar, I think we're ok without saying "draft_session_dates" everywhere here, I think we can just say "session_dates".

spec/features/draft_session_dates_e2e_spec.rb
```diff
@@ -0,0 +1,487 @@
+# frozen_string_literal: true
+
+describe "Draft session dates end-to-end functionality" do
```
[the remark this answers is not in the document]
MartinVanIJcken: I have renamed both the file and the tests in tests here

spec/features/draft_session_dates_e2e_spec.rb
MartinVanIJcken: removed them

spec/features/draft_session_dates_e2e_spec.rb
MartinVanIJcken: sure, that seems fine by me

MartinVanIJcken:
This pull request is no longer up to date and would need significant reworking.

PR #3819, opened 2025-06-25
Change triage options text

murugapl:
The triage form now specifies which delivery method in every safe to vaccinate option, for programmes with multiple delivery methods. The hint is also more explicit on what method(s) the parent has consented to.

image

https://nhsd-jira.digital.nhs.uk/browse/MAV-1285
https://nhsd-jira.digital.nhs.uk/browse/MAV-1398

### DIFF
--- a/spec/components/app_triage_form_component_spec.rb
+++ b/spec/components/app_triage_form_component_spec.rb
@@ -1,16 +1,16 @@
 # frozen_string_literal: true
 
 describe AppTriageFormComponent do
-  subject { render_inline(component) }
+  subject(:rendered) { render_inline(component) }
 
   let(:component) { described_class.new(triage_form, url:) }
+  let(:url) { "/triage" }
 
   let(:programme) { create(:programme) }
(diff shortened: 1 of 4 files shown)


spec/components/app_triage_form_component_spec.rb
```diff
@@ -21,4 +21,116 @@
 
     it { should have_css(".app-fieldset__legend--reset") }
   end
+
+  describe "hint text and triage options for consented delivery method(s)" do
+    before do
+      allow(programme).to receive(:has_multiple_delivery_methods?).and_return(
```
jhenderson, line 27: I'm wondering if we can avoid stubbing methods and use FactoryBot instead. The majority of the other component tests using FactoryBot to build state. Not a showstopper, but worth considering.

benilovj:
> @murugapl The ‘or’ divider is missing between yes and no options.

@paulrobertlloyd this one is a quite fiddly one to implement - is it possible to do as a follow-up PR?

PR #3823, opened 2025-06-25
Add FHIR records generation and sending to Immunisations API

misaka:
This adds support for generating a FHIR record for a vaccination record and sending it to the NHS Immunisations API. This is a big PR, but a lot of it is quite similar, mainly the FHIR mapper classes and related specs and changes to the models.

Here's whats changed in broad strokes:
- Added the `FHIRMapper` classes (in `app/lib/fhir_mapper/*`) to convert records in Mavis into FHIR records.
  - `FHIRMapper::VaccinationRecord` (in `app/lib/fhir_mapper/vaccination_record.rb`) is the main bit that generates FHIR, it relies on the other FHIR mappers to generate bits and pieces required from other models.
- Changed models to add delegations of `fhir_*` methods to the related FHIR mapper.
- Added `app/lib/nhs/immunisation_api.rb` to connect to Immunisations API and send a vaccination record. This bit has been tested against the integration environment.


app/models/programme.rb
```diff
@@ -147,4 +149,26 @@ def snomed_procedure_code
...
+    "flu" => "6142004"
+  }.freeze
+
+  def snomed_target_disease_code
+    SNOMED_TARGET_DISEASE_CODES.fetch(type)
+  end
+
+  SNOMED_TARGET_DISEASE_TERMS = {
+    "hpv" => "Human papillomavirus infection (disorder)",
+    "flu" => "Influenza (disorder)"
```
mikejamesthompson, line 164: Should use `Influenza` which is the preferred synonym for [6142004](https://termbrowser.nhs.uk/?perspective=full&conceptId1=6142004&edition=uk-edition&release=v20250604&server=https://termbrowser.nhs.uk/sct-browser-api/snomed&langRefset=999001261000000100,999000691000001104).

mikejamesthompson (changes requested):
Not a full review, just a few comments, mostly on SNOMED terms.

app/lib/fhir_mapper/vaccination_record.rb
```diff
@@ -0,0 +1,155 @@
...
+      else
+        raise ArgumentError, "Unknown outcome: #{outcome}"
+      end
+    end
+
+    def fhir_vaccine_code
+      FHIR::CodeableConcept.new(
+        coding: [
+          FHIR::Coding.new(
+            system: "http://snomed.info/sct",
```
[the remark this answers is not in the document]
misaka: tbh, I'm not sure of the benefit of adding a constant here. This string is almost certainly never to change, and if it does it's not like it's hard to find and change all instances of it in the code. It's also not a value that needs to be used outside of the actual FHIR mapper, so I'm tempted to just leave it.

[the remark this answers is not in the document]
mikejamesthompson: Agreed

mikejamesthompson (approved):
Those changes look good, thanks.

jhenderson (approved):
Nicely done @misaka. A relatively straightforward review as you've separated the concerns into various classes. 👍 

My only concern was using `delegate_missing_to`. I'm all for it if many methods are needed, but it appears some `FHIRMapper` objects only need one sometimes a bit more. What do you think about making those delegate methods explicit? I'm on the fence, so I'll leave that final decision to you.

app/lib/nhs/immunisations_api.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module NHS::ImmunisationsAPI
+  class PatientNotFound < StandardError
+  end
+
+  class << self
+    def record_immunisation(vaccination_record)
```
mikejamesthompson, line 8: Is it worth putting a guard here to make sure we're only sending vaccination records that were recorded directly in Mavis?

mikejamesthompson: And should we have a similar guard to only send records for vaccinations that were given? Or should these checks go somewhere else?

mikejamesthompson: Oh and also somewhere we need to limit this to just specific vaccination types.

mikejamesthompson: Discussed with Misha, and these restrictions will be put in place at the point the job is enqueued.

PR #3836, opened 2025-06-27
Create Instruction model for Patient Specific Directions (PSD)

jhenderson:
For patients requiring triage under the PSD protocol, nurses need an option to attach a PSD instruction, which acts as a permission for a healthcare assistant (HCA) to administer a specific vaccine method at a given date/time.

We chose the name "Instruction" to avoid PSD in the name for better abstraction, as the `protocol` column can have values like PSD, PGD, National or Nurses Only.


db/migrate/20250708131033_create_instructions.rb
```diff
@@ -0,0 +1,23 @@
...
+                     to_table: :users
+                   }
+      t.references :patient, null: false, foreign_key: true
+      t.references :programme, null: false, foreign_key: true
+      t.references :vaccine, null: false, foreign_key: true
+
+      t.string :vaccine_method, null: false
+      t.string :delivery_site, null: false
+      t.boolean :full_dose, null: false
+      t.string :protocol, null: false
```
thomasleese, line 18: I think `vaccine_method` and `delivery_site` should probably be enums (and therefore an integer) and with the same enum values as what we have already for triage status and vaccination records.

I think it would probably also make sense for `protocol` to be an integer-backed enum too.

thomasleese (approved):
Code change looks good, since we have these enums in a few models now I wonder if it's worth pulling them out in to concerns to keep the values in sync, but that's a further improvement.

Regarding the name `Instruction` I remember before I went on holiday we discussed whether it was too generic and we needed something like `PatientSpecificDirection`, but I can see that these now support different protocols (so maybe that's no longer appropriate).

jhenderson:
@thomasleese should we add auditing and soft deletion?

PR #3847, opened 2025-06-30
Add rollback documentation

bogsi17:


docs/releasing.md
```diff
@@ -133,3 +133,9 @@ At this point the histories of the `release` and `main` branches will have
 diverged and it will not be possible to fast-forward the `release` branch when
 releasing. It will have to be reset to the latest release candidate as
 previously described.
+
+## Rollback
+
+A release can be rolled back by deploying the previous release tag using the regular GitHub workflow.
```
TheOneFromNorway, line 139: Maybe add a comment that this can be done on a per-service level or for all services. Also documenting how to stop an ongoing deployment for ECS deploy process could be good

PR #3866, opened 2025-07-02
[DRAFT} Spike: auth sharing with reporting component (MAV-1406)

aldavidson:
This spike is a proof-of-concept, to test the [proposed approach](https://docs.google.com/presentation/d/1pKn8nLklvwyCv22s-aRwwIhahRgR3SU5t0T_c1S4TvU/edit?slide=id.p#slide=id.p) for managing auth tokens between the Rails app and the Python reporting component


app/controllers/concerns/token_authentication_concern.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    def authenticate_user_by_jwt!
+      jwt = jwt_if_given
+      jwt_info = decode_jwt!(jwt)
+      if jwt_info
+        data = jwt_info.first["data"]
+        @current_user =
+          User.find_by(
+            id: data["user"]["id"],
+            session_token: data["user"]["session_token"],
```
mikejamesthompson, line 45: As below, do we need to use `session_token` here if we're using `pwd_auth_session_token`?

app/controllers/one_time_tokens_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  skip_before_action :authenticate_user!
+  before_action :authenticate_app_by_token!
+
+  def verify
+    skip_policy_scope
+    @token = OneTimeToken.find_by!(token: params[:token])
+    @token.delete # <- Tokens are one-time use
+    json_data =
+      @token.as_json.merge({ user: @token.user.as_json, jwt: jwt(@token) })
```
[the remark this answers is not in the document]
aldavidson: We share the session token so that it can be used when we match the user in reporting-to-mavis API calls - that's the only practical way I found of supporting the backchannel logout requirement (ie. when a user removes their smartcard, they have to be automatically logged-out of everything)

Again, happy to chat about this and see if we can find a better way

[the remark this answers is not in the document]
aldavidson: I think you're right, we can probably get rid of this. In the very first version I just had 'user', then added the JWT once I had the basics working.

config/routes.rb
```diff
@@ -318,6 +318,12 @@
     resource :organisations, only: %i[new create]
   end
 
+  # for commissioner reporting app
+  post "/tokens/authorize", controller: :one_time_tokens, action: "authorize"
```
thomasleese, line 322: Should this route also be under `/api`?

app/models/one_time_token.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  validates :user_id, uniqueness: true
+  validates :token, uniqueness: true
+
+  def self.generate!(user_id:, cis2_info: {})
+    create!(user_id: user_id, token: SecureRandom.hex(32), cis2_info: cis2_info)
+  end
+
+  def self.expire_before
+    Settings.mavis_reporting_app.token_ttl_seconds.seconds.ago
```
thomasleese, line 34:
```suggestion
    Settings.one_time_token.ttl_seconds.seconds.ago
```

Since at the moment the `OneTimeToken` model is not specific to the reporting app, I think it would be good to keep it generic in its naming.

app/controllers/concerns/token_authentication_concern.rb
```diff
@@ -0,0 +1,72 @@
...
+module TokenAuthenticationConcern
+  extend ActiveSupport::Concern
+
+  included do
+    private
+
+    def client_id_error!(token)
+      if token.blank?
+        render json: { errors: "invalid_request" }, status: :unauthorized and
+          return
```
thomasleese, line 12: I don't think the `and return` does anything here since nothing comes after it.

app/controllers/reporting_api/totals_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module ReportingAPI
+  class TotalsController < ::ReportingAPI::BaseController
```
thomasleese, line 4:
```suggestion
  class TotalsController < BaseController
```

app/models/one_time_token.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def self.find_or_generate_for!(user_id:, cis2_info: {})
```
thomasleese, line 37:
```suggestion
  def self.find_or_generate_for!(user:, cis2_info: {})
```

I think everywhere this is used we have access to a user object, so I think this would be nicer to avoid needing to do `user_id: user.id` in lots of places.

db/migrate/20250722121803_rename_pwd_auth_session_token_to_reporting_app_session_token.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class RenamePwdAuthSessionTokenToReportingAppSessionToken < ActiveRecord::Migration[
+  8.0
+]
+  def change
+    rename_column :users, :pwd_auth_session_token, :reporting_app_session_token
```
thomasleese, line 7: I think we could probably combine this migration with the previous one to avoid needing to create the column and then rename it.

thomasleese (commented):
Overall approach looks good, just got some comments on the naming and code structure.

config/routes.rb
aldavidson: yes, see the [above comment](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3866#discussion_r2242128162) for details

db/migrate/20250722121803_rename_pwd_auth_session_token_to_reporting_app_session_token.rb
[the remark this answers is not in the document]
aldavidson: we'll keep this as `reporting_app_session_token` for now. The mechanism has been coded in such a way is it could certainly be used by other apps in future, but there would be some other slight modifications needed (like, the current check on `CLIENT_ID` is just against a single env var, for multiple clients you'd probably want that to be in a model, and add a UI for generating/revoking them, etc) so a rename can be done quite simply as part of that point if we ever get there.

config/routes.rb
```diff
@@ -318,6 +318,12 @@
...
+  namespace :reporting_api, path: "/reporting-api/" do
```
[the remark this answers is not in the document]
aldavidson: Done

config/feature_flags.yml
```diff
@@ -1,5 +1,10 @@
 testing_api: Basic API useful for automated testing.
 
+reporting_app: Python app for Commissioner reporting. Enabling this enables the
```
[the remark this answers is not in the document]
thomasleese: I feel like in that case it could make sense to have separate feature flags anyway.

PR #3874, opened 2025-07-04
Add test coverage for patient scope filtering edge cases

jhenderson:
We recently fixed a bug in year group filtering (#3867 ) that also resolved an issue with consent status filtering. Testers identified a discrepancy where patient counts differed between unfiltered and filtered views.

The root cause was patients who appeared in `PatientPolicy::Scope` through school moves or historical vaccination records but lacked programme-specific associations. These patients would be visible in unfiltered results but excluded when consent filters were applied, creating inconsistent counts.

These tests ensure such edge cases are caught in future changes, preventing similar regressions.


spec/forms/search_form_spec.rb
```diff
@@ -507,4 +507,50 @@
       end
     end
   end
+
+  context "when scope considers patients with school moves or vaccination records" do
```
misaka, line 511: Took me a bit to understand the setup of this test here, I wonder if it would be clearer if we said that we're using the `PatientPolicy::Scope`? Would the following be correct?

```suggestion
  context "using PatientPolicy scope which includes patients with school moves or vaccination records" do
```

misaka (approved):

PR #3875, opened 2025-07-04
MAV-1480 incorrect filters on vaccinations tab

MartinVanIJcken:


MartinVanIJcken:
This is a design choice, not a bug

PR #3876, opened 2025-07-04
Identity check default

murugapl:
During the identity check pre-vaccination, the child is now selected by default, reducing the number of clicks most vaccination flows will have. This is because a vast majority of children will be able to  identify themselves.

Also adds feature tests for the identity check flow.

https://nhsd-jira.digital.nhs.uk/browse/MAV-1349

### DIFF
--- a/app/components/app_vaccinate_form_component.html.erb
+++ b/app/components/app_vaccinate_form_component.html.erb
@@ -16,7 +16,7 @@
       <%= f.govuk_radio_buttons_fieldset :identity_check_confirmed_by_patient,
                                          legend: { text: "Has #{patient.given_name} confirmed their identity?", size: nil } do %>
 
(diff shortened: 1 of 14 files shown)


benilovj (approved):
I think this is OK to go in as is.

A few non-blocking comments, maybe for a follow-up PR:
* in other feature specs, we've combined the validation checking in with happy scenarios:
  * you set the scenario up
  * try something invalid
  * see a validation error
  * try something else invalid
  * see another validation error
  * then do the right thing and proceed.
* the advantage of this approach is that it reduces time-consuming set-ups and tear downs that add extra time to the spec running and balloon running times
* we may want to test changing the identifier from the 'check and confirm' screen
* we may want to test editing the identifier on a recorded vaccination

PR #3884, opened 2025-07-07
Sync vaccination records to NHSE Immunisations FHIR API

misaka:
Vaccination records recorded in Mavis need to be sent to the NHSE via the Immunisations FHIR API. Providing it is an HPV or flu vaccination, and the outcome is administered. This includes vaccinations recorded in the UI and via the offline import functionality.

NB: This PR requires that the flu vaccination records get a dose number recorded. Currently that isn't the case, but the PR this one, #3882 is based on adds that change.

[MAV-1482](https://nhsd-jira.digital.nhs.uk/browse/MAV-1482)

### DIFF
--- a/app/models/immunisation_import.rb
+++ b/app/models/immunisation_import.rb
@@ -109,5 +109,13 @@ class ImmunisationImport < ApplicationRecord
 
   def postprocess_rows!
     StatusUpdater.call(patient: patients)
+
+    vaccination_records
+      .recorded_in_service
+      .administered
+      .includes(:programme)
+      .find_each do |vaccination_record|
+        EnqueueSyncVaccinationRecordToNHSE.call(vaccination_record)
+      end
   end
 end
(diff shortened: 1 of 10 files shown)


app/models/immunisation_import.rb
```diff
@@ -109,5 +109,13 @@ def count_column(vaccination_record)
 
   def postprocess_rows!
     StatusUpdater.call(patient: patients)
+
+    vaccination_records
+      .recorded_in_service
+      .administered
+      .includes(:programme)
```
mikejamesthompson, line 116: Is it worth factoring out the logic for whether a record should be synced with the API into a single method on the `VaccinationRecord`? Means that as it changes in the future, it's going to be easier not to make a mistake.

misaka: I've updated `EnqueueVaccinationRecordToNHSE` so all the logic of deciding whether a vaccination record can be sent is in there. I'm just doing some local testing now.

jhenderson (approved):
LGTM 👍

PR #3898, opened 2025-07-08
 Show warning when recording a vaccine delivery method that conflicts with the consent or triage

MartinVanIJcken:
JIRA: [MAV-1461](https://nhsd-jira.digital.nhs.uk/browse/MAV-1461).

### DIFF
--- a/app/models/draft_vaccination_record.rb
+++ b/app/models/draft_vaccination_record.rb
@@ -244,7 +244,22 @@ class DraftVaccinationRecord
       identity_check&.confirmed_by_other_relationship
   end
 
+  def vaccine_method_matches_consent_and_triage?
+    return true if delivery_method.blank? || !administered?
+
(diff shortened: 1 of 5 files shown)


spec/components/app_vaccination_record_summary_component_spec.rb
```diff
@@ -327,4 +327,242 @@
       expect(rendered).to have_css(".app-highlight", text: "1 January 2027")
     end
   end
+
+  describe "#incorrect_vaccine_method?" do
```
benilovj, line 331: We tend to avoid using doubles / mocks for testing view components (see other specs in this file), as they're quite brittle – best use object factories instead.

benilovj (changes requested):

benilovj:
@thomasleese I've rebased and updated this PR with the feedback. I've probably done too much on it to review it...

thomasleese (approved):

PR #3899, opened 2025-07-08
Improve Data replication workflow

bogsi17:
This PR contains some improvements for the data replication workflow

* DB destruction and recreation is done in a single job, that shortens the workflow duration and requires just 1 approval
* The workflow stops if there was an error in the terraform plan job
* Added an option to deploy just a new web app, skipping the DB recreation altogether

### DIFF
--- a/.github/workflows/data-replication-pipeline.yml
+++ b/.github/workflows/data-replication-pipeline.yml
@@ -1,5 +1,5 @@
 name: Data replication pipeline
-run-name: ${{ inputs.action }} data replication resources for ${{ inputs.environment }}
+run-name: ${{ inputs.deployment_type }} for data replication resources for ${{ inputs.environment }}
 
 on:
   workflow_dispatch:
@@ -15,18 +15,17 @@ on:
           - qa
           - sandbox-alpha
           - sandbox-beta
+      deployment_type:
+        description: Deployment type
+        required: true
+        type: choice
+        options:
+          - Deployment with DB recreation
+          - Application only deployment
       image_tag:
         description: Docker image tag to deploy
         required: false
         type: string
-      action:
-        description: Action to perform on data replication env
-        required: true
-        type: choice
-        options:
-          - Destroy
-          - Recreate
-        default: Recreate
       db_snapshot_arn:
         description: ARN of the DB snapshot to use (optional)
         required: false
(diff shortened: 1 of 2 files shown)


.github/workflows/data-replication-pipeline.yml
```diff
@@ -200,12 +157,24 @@ jobs:
...
+            "-var=imported_snapshot=${{ env.SNAPSHOT_ARN }}"
+            "-var-file=env/${{ inputs.environment }}.tfvars"
+            "-var=allowed_egress_cidr_blocks=$CIDR_BLOCKS"
+            "-out=${{ runner.temp }}/tfplan"
+          )
+          
+          if [ "${{ env.REPLACE_DB_CLUSTER }}" = "true" ]; then
+            PLAN_ARGS+=("-replace" "aws_rds_cluster.cluster")
+          fi
+          terraform "${PLAN_ARGS[@]}" | tee ${{ runner.temp }}/tf_stdout
```
TheOneFromNorway, line 177: We never check the return code of terraform plan so this will report success even when the plan fails.  Also since we never use `${{ runner.temp }}/tf_stdout` I suggest instead
```
terraform "${PLAN_ARGS[@]}"
if [[ "$?" -eq 1 ]]; then 
  exit 1
else
  exit 0
fi
```

.github/workflows/data-replication-pipeline.yml
```diff
@@ -200,12 +157,24 @@ jobs:
...
+            "-var=image_digest=${{ env.DOCKER_DIGEST }}"
+            "-var=db_secret_arn=${{ env.DB_SECRET_ARN }}"
+            "-var=imported_snapshot=${{ env.SNAPSHOT_ARN }}"
+            "-var-file=env/${{ inputs.environment }}.tfvars"
+            "-var=allowed_egress_cidr_blocks=$CIDR_BLOCKS"
+            "-out=${{ runner.temp }}/tfplan"
+          )
+          
+          if [ "${{ env.REPLACE_DB_CLUSTER }}" = "true" ]; then
+            PLAN_ARGS+=("-replace" "aws_rds_cluster.cluster")
```
TheOneFromNorway, line 175: Because we use 
`cluster_identifier     = "${local.name_prefix}-rds-${formatdate("hh-mm-ss", timestamp())}"`
the cluster will be replaced every time we run terraform apply, so this parameter is redundant.

PR #3901, opened 2025-07-09
Add progress bar to the CLI command `mavis generate cohort-imports`

alistair-white-horne-tng:
When generating large imports (>1000), the command is slow, and some feedback is appreciated.


thomasleese:
@alistair-white-horne-tng I've changed the base branch to `next`.

Issue #1943, opened 2024-10-08
Handle yearly rollover


thomasleese:
Each academic year we need to perform a number of steps to ensure the system is configured correctly, these are currently manual and we might to think about how to make these automated.

- Create unscheduled sessions for each team and location
  - https://github.com/nhsuk/manage-vaccinations-in-schools/pull/1942
  - `UnscheduledSessionsFactory.new.call`

PR #3904, opened 2025-07-09
Send flu reminders also when vaccination/consent were received last year

MartinVanIJcken:
There was a bug that would prevent consent notifications from being sent when the parent had already consented last year or when the child had already been vaccinated last year. This behavior is incorrect for seasonal vaccines for which consent is requested again every year and which need to be taken again every year too.

This fixes [MAV-1524](https://nhsd-jira.digital.nhs.uk/browse/MAV-1524)

### DIFF
--- a/app/lib/consented_criteria.rb
+++ b/app/lib/consented_criteria.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class ConsentedCriteria
+  def initialize(programme:, patient:)
+    @programme = programme
+    @patient = patient
+    @consents = patient.consents
+  end
+
+  def call
+    consents_for_programme = consents.select { it.programme_id == programme.id }
+
+    if programme.seasonal?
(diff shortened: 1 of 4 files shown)


app/lib/consented_criteria.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class ConsentedCriteria
```
thomasleese, line 3: I think the use of this class for the emails means we need to make some further changes. For example, should the consent status of a patient also use this method to ensure that when the year rolls over the patient is no longer appearing as having received consent for Flu?

I would suggest we keep the scope of this PR specific to vaccinations, and we worry about the consent expiration issue when we deal with rollover (#1943). Fixing the issue for previous vaccinations solves the immediate problem as we won't have any patients with previous consents for Flu for the programme this coming year.

PR #3908, opened 2025-07-09
Add protocol enum to vaccination_records table

jhenderson:
Patient Group Directions (PGDs) are formal written instructions that allow specified registered healthcare professionals (HCPs) to supply and/or administer certain prescription-only medicines (POMs) to groups of patients who meet criteria defined within the PGD. In the case of Mavis, the POMs are vaccines.

The plan is to have a protocol on each vaccination record which can either be pgd, psd (or national even). By default, and what’s been recorded so far in Mavis has been via the PGD protocol.

If there's a `PatientSpecificDirection` record then the vaccination record is being recorded under the Patient Specific Directions (PSD's) protocol. PSD's will only exist in certain situations when a clinician has gone through and created them.


app/models/draft_vaccination_record.rb
```diff
@@ -129,6 +131,10 @@ def already_had?
   # So that a form error matches to a field in this model
   alias_method :administered, :administered?
 
+  def protocol
+    super || :pgd
```
thomasleese, line 135: Do we need the `super` here? I can't see which class it would be getting this method from?

jhenderson: > Do we need the super here?

On second thoughts I don't think so.

thomasleese (approved):

PR #3909, opened 2025-07-09
Pass triage to `GovukNotifyPersonalisation`

thomasleese:
This allows for emails and texts to include personalisation content according to the triage. Specifically we need this for Flu where content will be different according to the vaccine method chosen at the triage stage.

To support this I've also been able to simplify the `TriageMailerConcern`.


tvararu (approved):

thomasleese:
We can do this in an alternative way, using the triage and consent status of the patient, so closing this for now.

PR #3915, opened 2025-07-10
Add flu pre-screening questions

murugapl:
Ask about recent asthma flare ups in the pre-screening form for flu vaccinations, only when nasal spray is being administered. Also ask the medication question for all flu vaccinations

https://nhsd-jira.digital.nhs.uk/browse/MAV-1495

### DIFF
--- a/app/forms/vaccinate_form.rb
+++ b/app/forms/vaccinate_form.rb
@@ -75,10 +75,6 @@ class VaccinateForm
     draft_vaccination_record.save # rubocop:disable Rails/SaveBang
   end
 
-  def ask_not_taking_medication? = programme.doubles?
-
-  def ask_not_pregnant? = programme.td_ipv?
-
   private
 
   delegate :organisation, to: :patient_session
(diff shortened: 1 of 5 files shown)


app/forms/vaccinate_form.rb
```diff
@@ -75,10 +75,12 @@ def save(draft_vaccination_record:)
     draft_vaccination_record.save # rubocop:disable Rails/SaveBang
   end
 
-  def ask_not_taking_medication? = programme.doubles?
+  def ask_not_taking_medication? = programme.doubles? || programme.flu?
 
   def ask_not_pregnant? = programme.td_ipv?
 
+  def ask_asthma_flare_up? = programme.flu?
```
thomasleese, line 82:
```suggestion
  def ask_asthma_flare_up? = programme.flu? && delivery_method == :nasal_spray
```

I would suggest keeping all the logic on whether to ask the question in the single method so it can be re-used elsewhere if necessary.

murugapl: Yeah I wanted to do that but delivery method is defined in the component and these question-specific methods in the vaccinate_form. I've moved those methods into the component now- does that work?

thomasleese (approved):

app/forms/vaccinate_form.rb
thomasleese: Yeah, that looks good!

PR #3936, opened 2025-07-12
Allow default daily batch per vaccine method

thomasleese:
This ensures that when administered vaccinations for a programme like flu where there are multiple vaccine methods (injection vs nasal), the default batch for the day is different for the two vaccine methods.

[Jira Issue - MAV-1431](https://nhsd-jira.digital.nhs.uk/browse/MAV-1431)

### DIFF
--- a/app/components/app_consent_confirmation_component.rb
+++ b/app/components/app_consent_confirmation_component.rb
@@ -66,14 +66,11 @@ class AppConsentConfirmationComponent < ViewComponent::Base
         .map do |consent_form_programme|
           programme = consent_form_programme.programme
 
-          if programme.flu?
-            if consent_form_programme.vaccine_method_nasal?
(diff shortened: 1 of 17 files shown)


benilovj (approved):
Code looks really good - thanks for sorting out the delivery methods vs vaccine methods situation.

I have a couple of non-blocking design-related questions, which we can clarify on Monday and perhaps can be done as a follow-up PR:

### More explicit delivery method in the list of day's batches

Current design:

Image

Potential alternative:

Image

### More explicit batch description when picking the default

Current design:

Image

Potential alternative:

Image

thomasleese:
> I just realised that we ought to update the page for editing the default batch so it

Agreed, I've amended the commit with both of those.

benilovj (approved):

PR #3947, opened 2025-07-14
Triage for flu: UI changes

murugapl:
Make vaccine method explicit in relation to triage for the flu programme and other minor visual changes to the triage form

- Display vaccine method in status tags on session >triage pages, patient session pages and programme>children page
- Add an "or" divider between yes and no options on the triage form (applies to all programmes)
- Add hints to flu triage form to indicate the vaccine method the nurse has triaged safe for
- Add the vaccine method to the "safe to vaccinate" triage entry in the activity log for all flu triages

https://nhsd-jira.digital.nhs.uk/browse/MAV-1413
https://nhsd-jira.digital.nhs.uk/browse/MAV-1500

Triage form
Activity log
Patient session page
Session triage page
Programme children page

### DIFF
--- a/app/components/app_patient_session_triage_component.html.erb
+++ b/app/components/app_patient_session_triage_component.html.erb
@@ -6,10 +6,6 @@
   <% if latest_triage.nil? || latest_triage.needs_follow_up? %>
     <p>You need to decide if it’s safe to vaccinate.</p>
 
-    <% if triage_status&.consent_requires_triage? %>
-      <p>Responses to health questions need triage.</p>
-    <% end %>
-
     <% if triage_status&.vaccination_history_requires_triage? %>
       <p>Incomplete vaccination history for <%= programme.name_in_sentence %>. Check if the child needs another dose.</p>
     <% end %>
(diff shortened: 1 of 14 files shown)


app/components/app_patient_session_triage_component.html.erb
```diff
@@ -22,11 +18,15 @@
     <% end %>
   <% elsif latest_triage %>
     <% if latest_triage.ready_to_vaccinate? %>
-      <p><%= latest_triage.performed_by.full_name %> decided that <%= patient.full_name %> is safe to vaccinate.</p>
+      <% if programme.flu? %>
```
MartinVanIJcken, line 21: Perhaps it would be better to write a helper method like `triage_specifies_delivery_method`

app/helpers/triage_statuses_helper.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module TriageStatusesHelper
+  def triage_status_specifies_delivery_method?(triage_status)
+    return false unless triage_status&.persisted?
+    triage_status =
+      Patient::TriageStatus.includes(:programme).find(triage_status.id)
```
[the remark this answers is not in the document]
murugapl: Since `programme.has_multiple_vaccine_methods?` triggers a lazy load, this raises an error. So I 've moved that check out of the method- does that work?

app/models/patient/triage_status.rb
```diff
@@ -81,6 +81,10 @@ def vaccination_history_requires_triage?
     end && !vaccinated?
   end
 
+  def specifies_delivery_method?
+    safe_to_vaccinate? && vaccine_method.present?
+  end
```
thomasleese, line 86: I find the naming of this method a little confusing, but I'm also wondering if everywhere that we do `specifies_delivery_method?`, we can do `vaccine_method.present?` instead because they'll only be present if the patient is safe to vaccinate.

thomasleese (approved):

PR #3966, opened 2025-07-15
Refactor syncing of vaccination records to immunisations API

misaka:
This changes how we check if a `VaccinationRecord` needs to be synced from relying on `updated_at` to a newly added column `nhs_immunisations_api_sync_pending_at`. This will allows us to trigger when data associated to the vaccination record changes, e.g. the patient's NHS number.

The sync logic is also moved from `SyncVaccinationRecordToNHSJob` into `NHS::ImmunisationsAPI` to keep it all in one place.

[MAV-1475](https://nhsd-jira.digital.nhs.uk/browse/MAV-1475)


app/models/immunisation_import.rb
```diff
@@ -110,6 +110,8 @@ def count_column(vaccination_record)
   def postprocess_rows!
     StatusUpdater.call(patient: patients)
 
-    EnqueueSyncVaccinationRecordToNHS.call(vaccination_records)
+    vaccination_records.syncable_to_nhs_immunisations_api.each(
```
jhenderson, line 113: Could use `find_each` instead of `each`

thomasleese (approved):

PR #3972, opened 2025-07-16
Check flipper before running CLI imms API sync command

alistair-white-horne-tng:
This change adds a check at the beginning of the vaccination records sync command to verify that the `immunisations_fhir_api_integration` feature flag is enabled before initiating the sync operation.

Previously, the command claimed that the sync was successful, even though nothing had happened.

### DIFF
--- a/spec/features/cli_vaccination_records_sync_spec.rb
+++ b/spec/features/cli_vaccination_records_sync_spec.rb
@@ -3,6 +3,10 @@
 require_relative "../../app/lib/mavis_cli"
 
 describe "mavis vaccination-records sync" do
+  before { Flipper.enable :immunisations_fhir_api_integration }
+
+  after { Flipper.disable :immunisations_fhir_api_integration }
+
   context "when the vaccination record exists and has not been synced" do
     it "syncs the vaccination record to the NHS API" do
       given_a_vaccination_record_exists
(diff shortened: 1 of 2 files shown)


spec/features/cli_vaccination_records_sync_spec.rb
```diff
@@ -92,6 +101,7 @@ def when_i_run_the_sync_command
   end
 
   def when_i_run_the_sync_command_with_an_invalid_id
+    Flipper.enable :immunisations_fhir_api_integration
```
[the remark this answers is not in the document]
alistair-white-horne-tng: How about this now? I'm not sure how I feel about the duplication of `given_...` and `and_...`, but this makes it read the nicest in the `context` sections. Are you happy with that?

spec/features/cli_vaccination_records_sync_spec.rb
```diff
@@ -60,6 +72,18 @@ def given_a_synced_vaccination_record_exists
       )
   end
 
+  def given_the_immunisations_fhir_api_integration_is_enabled
+    Flipper.enable :immunisations_fhir_api_integration
+  end
+
+  def and_the_immunisations_fhir_api_integration_is_enabled
```
benilovj, line 79: I think it's slightly more idiomatic to use `alias_method` here:

```
  alias_method :and_the_immunisations_fhir_api_integration_is_enabled,
               :given_the_immunisations_fhir_api_integration_is_enabled
```

spec/features/cli_vaccination_records_sync_spec.rb
```diff
@@ -6,6 +6,7 @@
   context "when the vaccination record exists and has not been synced" do
     it "syncs the vaccination record to the NHS API" do
       given_a_vaccination_record_exists
+      and_the_immunisations_fhir_api_integration_is_enabled
```
benilovj, line 9: Would it be cleaner and simpler to add the following after the `describe` block:

```
before do
  Flipper.enable :immunisations_fhir_api_integration
end

after do
  Flipper.disable :immunisations_fhir_api_integration
end
```

and then override it where you want to explicitly disable it (the innermost `before` block should be executed last):
```
before do
  Flipper.disable :immunisations_fhir_api_integration
end
```
and that way you can skip all the extra steps

spec/features/cli_vaccination_records_sync_spec.rb
benilovj: This becomes moot with the suggestion above.

PR #3984, opened 2025-07-17
Allow custom programme year groups per school

thomasleese:
This introduces a new `Location::ProgrammeYearGroup` model stored on each location and programme that stores the year groups that are administered for a particular programme at a particular location. This is needed because although most schools administer the same set of year groups per programme, there will be exceptions to this.

To support this we will need to run the Rake task `create_default_programme_year_groups` soon after deployment to ensure that existing locations are set up with the correct year groups as they now will be via the onboarding or `schools:add_to_organisation` Rake task.

This also adds two new CLI commands which can be used to add or remove programme year groups from a particular school. This will be needed to be used by the operations team when supporting organisations that administer flu in schools outside the normal year groups.

- `mavis schools add-programme-year-group URN PROGRAMME_TYPE YEAR_GROUPS`
- `mavis schools remove-programme-year-group URN PROGRAMME_TYPE YEAR_GROUPS`

[Jira Issue - MAV-1429](https://nhsd-jira.digital.nhs.uk/browse/MAV-1429)


app/models/organisation.rb
```diff
@@ -82,13 +82,17 @@ def generic_team
...
+        Location
+          .create!(
+            name: "Community clinics",
+            team: generic_team,
+            ods_code:,
+            type: :generic_clinic,
+            year_groups:
+          )
+          .tap { it.create_default_programme_year_groups!(programmes) }
+      end
```
misaka, line 101: Just checking this is meant to be here, since it doesn't match the commit message. It looks right to me, just probably was missed from an earlier commit.

app/lib/govuk_notify_personalisation.rb
```diff
@@ -400,6 +392,10 @@ def vaccine_side_effects
 
   private
 
+  def administered_year_groups
```
misaka, line 395: Minor nit, but it was blocking my understanding of the code. Why "administered"? Is it not simply "year_groups"?

benilovj: Perhaps something like `year_groups_in_programme_scope` or `eligible_year_groups` (one for the future maybe...)

PR #3996, opened 2025-07-18
Include academic year in status models

thomasleese:
This updates the three patient status models (consent, triage and vaccination) to consider the academic year when generating the status. Specifically, we now create distinct instances of these models per academic year, allowing for various views in the service to be tailored to a specific academic year. See individual commits for more details.

In terms of test coverage, most of this has been implemented by adding `academic_year` as a required argument in a number of places, and therefore the tests passing indicates that this is being taken in to account, however it's worth adding more test coverage in follow up PRs as we implement the various features of rollback that will be built upon this work.

We should be careful about any potential performance implications from this change, specifically the number of consent status, triage status and vaccination statuses will double each academic year. The migration which adds the new column also updates the index, and this is likely going to block the tables while it's running, so I would recommend that we run this out of hours.

[Jira Issue - MAV-267](https://nhsd-jira.digital.nhs.uk/browse/MAV-267)

### DIFF
--- a/app/components/app_patient_search_result_card_component.rb
+++ b/app/components/app_patient_search_result_card_component.rb
@@ -5,6 +5,7 @@ class AppPatientSearchResultCardComponent < ViewComponent::Base
     patient,
     link_to:,
     programme: nil,
+    academic_year: nil,
     triage_status: nil,
     show_parents: false,
     show_postcode: false,
@@ -16,6 +17,7 @@ class AppPatientSearchResultCardComponent < ViewComponent::Base
     @patient = patient
     @link_to = link_to
     @programme = programme
+    @academic_year = academic_year
     @triage_status = triage_status
 
     @show_parents = show_parents
@@ -61,7 +63,7 @@ class AppPatientSearchResultCardComponent < ViewComponent::Base
             row.with_value { helpers.patient_parents(@patient) }
           end
         end
-        if @programme
+        if @programme && @academic_year
           summary_list.with_row do |row|
             row.with_key { "Consent status" }
             row.with_value { consent_status_tag }
@@ -97,7 +99,11 @@ class AppPatientSearchResultCardComponent < ViewComponent::Base
 
   def render_status_tag(status_type, outcome)
     status_model =
-      @patient.public_send("#{status_type}_status", programme: @programme)
+      @patient.public_send(
+        "#{status_type}_status",
+        programme: @programme,
+        academic_year: @academic_year
+      )
 
     status_key =
       if status_type == :triage && status_model.vaccine_method.present? &&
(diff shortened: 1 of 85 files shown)


misaka (approved):
Big one ... at this size, if it works, it ships? 😅 🚀

MartinVanIJcken:
Is there a reason you chose to put the academic_year filtering in the ConsentGrouper rather than using the academic year scope directly on the consents? 

E.g.

```
ConsentGrouper
        .call(
          @patient.reload.consents,
          programme_id: @programme.id,
          academic_year: @academic_year
        )
        .each { send_triage_confirmation(@patient_session, @programme, it) }
```
instead of

```
ConsentGrouper
        .call(
          @patient.reload.consents.for_academic_year(@academic_year),
          programme_id: @programme.id
        )
        .each { send_triage_confirmation(@patient_session, @programme, it) }
```

MartinVanIJcken:
Do you think the changes form https://github.com/nhsuk/manage-vaccinations-in-schools/pull/4053/commits/7859be133bda2cadf9cfaed7fe92e5a16bbaeb1f and https://github.com/nhsuk/manage-vaccinations-in-schools/pull/4053/commits/60d0071d859b411c09bdb9346e636b27606922e0

would be better off included in this PR?

thomasleese:
> Do you think the changes from [7859be1](https://github.com/nhsuk/manage-vaccinations-in-schools/commit/7859be133bda2cadf9cfaed7fe92e5a16bbaeb1f) and [60d0071](https://github.com/nhsuk/manage-vaccinations-in-schools/commit/60d0071d859b411c09bdb9346e636b27606922e0) would be better off included in this PR?

I think it's okay to keep them separate to this PR. I've tried to keep the scope of this PR specific to only the three patient status classes and anything beyond that I think it fine to stay in #4053.

MartinVanIJcken:
I noticed VaccinationStatus has had an academic_year column added but was not given any scope to use that information. Was it a mistake to add the academic_year column on this model?

thomasleese:
> I noticed VaccinationStatus has had an academic_year column added but was not given any scope to use that information. Was it a mistake to add the academic_year column on this model?

`VaccinationStatus` should have an academic year, because the patient's vaccination status can be different per year (for example with flu where you're given a vaccination each year).

It's used when rendering the vaccination status of a patient in various places, as the [`Patient#vaccination_status` method now requires an `academic_year` to be passed in](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3996/files#diff-94dcc014f62f0d9b6084ad98f82f74cd40c86b029ef1691914bdbc960c2f42dcR320) where the [right record for that academic year is then selected](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/3996/files#diff-94dcc014f62f0d9b6084ad98f82f74cd40c86b029ef1691914bdbc960c2f42dcR463).

PR #3998, opened 2025-07-18
Readonly container filesystem

bogsi17:
Making the ECS container filesystem readonly prevents unlimited write access in case a running container gets compromised.
Specific write-access is allowed to certain directories that are added as volumes in the Dockerfile. In addition to the rails directories, the SSM systems manager that gives CLI access on running containers also needs write access to certain directories


misaka (approved):
Interesting. Having this read-write has helped with operational issues in the past, although on balance I'd say we really should be making them read-only.

PR #4010, opened 2025-07-21
Use aqua-green for registration attending status colour

paulrobertlloyd:
Changing the registration attending status colour to aqua-green means that nurses can more easily see if a child is ready to vaccinate as all the status colours align: consent outcome, (triage outcome), (registration outcome) and soon (PSD outcome). Nurses often like to think that “green means good to go”.


tvararu:
💅

PR #4011, opened 2025-07-21
Set ECR lifecycle policy

bogsi17:
With this lifecycle policy, images will be stored only for three months in ECR before they get deleted

Jira-Issue: MAV-1582

### DIFF
--- a/terraform/account/main.tf
+++ b/terraform/account/main.tf
@@ -101,6 +101,27 @@ resource "aws_ecr_repository" "mavis" {
   image_tag_mutability = "MUTABLE"
 }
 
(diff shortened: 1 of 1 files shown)


terraform/account/main.tf
```diff
@@ -101,6 +101,27 @@ resource "aws_ecr_repository" "mavis" {
   image_tag_mutability = "MUTABLE"
 }
 
+resource "aws_ecr_lifecycle_policy" "mavis" {
+  repository = aws_ecr_repository.mavis.name
+  policy = jsonencode({
```
TheOneFromNorway, line 106: Should we make use of the aws_ecr_lifecycle_policy_document data object like we have with IAM policies?

PR #4036, opened 2025-07-22
Rename `Team` to `Subteam`

thomasleese:
This renames the team model, and all associated models and foreign keys to subteam. The functionality of the service before and after should be the same.

We're doing this because we're going to be restructuring organisations and teams in to three levels. What we currently call teams becomes subteams, organisations becomes teams and we'll be introducing a new top-level organisation to group teams together.

To support this change I've also added three new personalisation variables to use in GOV.UK Notify: `subteam_name`, `subteam_email` and `subteam_phone`. These are equivalent to the existing variables prefixed with `team_` and once the templates have been updated we can remove them.

[Jira Issue - MAV-1280](https://nhsd-jira.digital.nhs.uk/browse/MAV-1280)

### DIFF
--- a/docs/managing-teams.md
+++ b/docs/managing-teams.md
@@ -21,18 +21,18 @@ organisation:
 
 programmes: [] # A list of programmes (flu, hpv, menacwy, td_ipv)
 
-teams:
-  team1: # Identifier to link team with school and links below, not used in app
+subteams:
+  subteam1: # Identifier to link team with school and links below, not used in app
     name: # Name of the team
     email: # Contact email address
     phone: # Contact phone number
     reply_to_id: # Optional GOV.UK Notify Reply-To UUID
 
 schools:
-  team1: [] # URNs managed by a particular team
+  subteam1: [] # URNs managed by a particular team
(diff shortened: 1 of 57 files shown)


docs/managing-teams.md
```diff
@@ -21,18 +21,18 @@ organisation:
 
 programmes: [] # A list of programmes (flu, hpv, menacwy, td_ipv)
 
-teams:
-  team1: # Identifier to link team with school and links below, not used in app
+subteams:
+  subteam1: # Identifier to link team with school and links below, not used in app
```
benilovj, line 25: Small nit; we're using `team` and `subteam` here and it's not 100% clear whether that's the same concept or different ones

benilovj (approved):

docs/managing-teams.md
benilovj: oh right, having seen https://github.com/nhsuk/manage-vaccinations-in-schools/pull/4038, I'm with you now.

misaka (approved):

PR #4038, opened 2025-07-22
Rename `Organisation` to `Team`

thomasleese:
This renames the organisation model, and all associated models and foreign keys to team. The functionality of the service before and after should be the same.

We're doing this because we're going to be restructuring organisations and teams in to three levels. What we currently call teams becomes subteams, organisations becomes teams and we'll be introducing a new top-level organisation to group teams together.

To support this change I've also added two new personalisation variables to use in GOV.UK Notify: `team_privacy_policy_url` and `team_privacy_notice_url`. These are equivalent to the existing variables prefixed with `organisation_` and once the templates have been updated we can remove them.

Depends on #4036 

[Jira Issue - MAV-1280](https://nhsd-jira.digital.nhs.uk/browse/MAV-1280)


misaka (approved):
yOlO!!!

PR #4048, opened 2025-07-23
Prevent changing the vaccination record administration date to a different academic year

murugapl:
This change adds a validation that the vaccination event administration date falls within the same academic year as the session it is attached to. This validation prevents editing vaccination records to be outside the given academic year.

https://nhsd-jira.digital.nhs.uk/browse/MAV-1567

image

### DIFF
--- a/app/models/draft_vaccination_record.rb
+++ b/app/models/draft_vaccination_record.rb
@@ -54,11 +54,8 @@ class DraftVaccinationRecord
   end
 
   on_wizard_step :date_and_time, exact: true do
-    validates :performed_at,
-              presence: true,
-              comparison: {
-                less_than_or_equal_to: -> { Time.current }
-              }
+    validates :performed_at, presence: true
+    validate :performed_at_within_range
   end
 
   on_wizard_step :outcome, exact: true do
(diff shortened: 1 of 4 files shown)


app/models/draft_vaccination_record.rb
```diff
@@ -154,6 +151,31 @@ def patient=(value)
...
+        :performed_at,
+        "Vaccination cannot take place before #{earliest_possible_value.to_fs(:long)}"
+      )
+    elsif performed_at > latest_possible_value
+      errors.add(
+        :performed_at,
+        "Vaccination cannot take place after #{latest_possible_value.to_fs(:long)}"
+      )
+    end
+  end
```
thomasleese, line 177: I would suggest moving these methods in to the `private` section below as they shouldn't be called outside this class.

app/models/draft_vaccination_record.rb
```diff
@@ -296,6 +293,31 @@ def reset_unused_fields
...
+
+  def latest_possible_value
+    [
+      session.academic_year.to_academic_year_date_range.last.end_of_day,
+      Time.current
+    ].min
+  end
+
+  def performed_at_within_range
+    if performed_at < earliest_possible_value
```
thomasleese, line 308: I think it's possible for `performed_at` to be `nil` if the user has cleared the text boxes. They would still see a validation error from `validates :performed_at, presence: true` but I think this validation would fail due to the comparison of `nil`. Perhaps we can skip this check if it's `nil`?

app/models/draft_vaccination_record.rb
```diff
@@ -296,6 +293,31 @@ def reset_unused_fields
     end
   end
 
+  def earliest_possible_value
+    session.academic_year.to_academic_year_date_range.first.beginning_of_day
```
thomasleese, line 297: I think it's possible for `session` to be `nil` if a historical record is being edited, in which case we can skip the validation.

PR #4056, opened 2025-07-23
Show Imms API sync status on vaccination records

alistair-white-horne-tng:
This change adds a status for the imms API sync. These are the available statuses:

- Not synced: record isn't synced with the API, used for records which are:
    - not administered (eg unwell)
    - historic (either upload, or eventually flowed from the API)
    - marked as already vaccinated
- Pending: the record has been created or updated but not yet synced with the API.
- Synced: the imms API record was successfully updated after the local record was updated.
- Failed: a day has passed since the local record was updated with no successful update to the remote record.
- Cannot sync: the associated patient is missing an NHS number

Also adds a line or two of help text underneath the tag.

image

[MAV-1411](https://nhsd-jira.digital.nhs.uk/browse/MAV-1411)


app/components/app_vaccination_record_summary_component.rb
```diff
@@ -236,6 +236,15 @@ def call
           end
         end
       end
+
+      if @vaccination_record.respond_to?(:sync_status)
```
thomasleese, line 240: Would this ever be `false`?

app/helpers/vaccination_records_helper.rb
```diff
@@ -27,4 +27,70 @@ def vaccination_record_status_tag(vaccination_record)
...
+      when :cannot_sync
+        "orange"
+      else
+        "grey"
+      end
+
+    govuk_tag(text:, colour:)
+  end
+
+  def format_sync_status_with_info(vaccination_record)
```
thomasleese, line 52: This feels like it has quite a lot of logic and complexity to be a helper method to me. I would suggest it might be better placed as its own component, and then because it's a class we can have methods for all the various parts to this.

app/components/app_vaccination_record_summary_component.rb
alistair-white-horne-tng: Yes, it seems that `@vaccination_record` can sometimes be of type `DraftVaccinationRecord`, which doesn't have the concept of syncing to the API

app/models/concerns/vaccination_record_sync_to_nhs_immunisations_api_concern.rb
```diff
@@ -12,6 +12,44 @@ def syncable_to_nhs_immunisations_api?
     recorded_in_service?
   end
 
+  def sync_status
```
misaka, line 15: I think the way we're using "synced" is a little ambiguous below. We seem to be using to indicate if a vaccination record has been sent to the Imms API, but technically "in-sync" would also mean that a record that shouldn't be in the  Imms API isn't there. The opposite can happen, for example, when a patient's NHS number is invalidated/removed., which will result in `cannot_sync` below, even while we're waiting for it to be deleted.

Also, I think we might be able to reuse `next_sync_action` for some of the logic here. Let's have a chat about this?

app/components/app_vaccination_record_api_sync_status_component.rb
```diff
@@ -0,0 +1,67 @@
...
+    cannot_sync: "orange",
+    not_synced: "grey"
+  }.freeze
+
+  def initialize(vaccination_record)
+    super
+    @vaccination_record = vaccination_record
+  end
+
+  def call
```
thomasleese, line 17: I would suggest this method can be broken up in to various private methods on the class to improve readability. 

Just to give an example of the kind of thing I mean:

```rb
attr_reader :vaccination_record

delegate :nhs_immunisations_api_synced_at, :sync_status, to: :vaccination_record

def secondary_text(text) = tag.span(text, class: "nhsuk-u-secondary-text-color")

def last_synced_at
  return unless %i[not_synced cannot_sync].include?(status)

  secondary_text(
    "Last synced: #{
      if nhs_immunisations_api_synced_at?
        nhs_immunisations_api_synced_at.to_fs(:long)
      else
        "Never"
      end
    }"
  )
end

def additional_information_text
  ...
end

def call
  safe_join([sync_status_tag, additional_information_text, last_sync_at].compact, tag.br)
end
```

app/models/concerns/vaccination_record_sync_to_nhs_immunisations_api_concern.rb
```diff
@@ -12,6 +12,44 @@ def syncable_to_nhs_immunisations_api?
...
+    end
+
+    if nhs_immunisations_api_synced_at.present? &&
+         (
+           !nhs_immunisations_api_sync_pending_at? ||
+             nhs_immunisations_api_synced_at >
+               nhs_immunisations_api_sync_pending_at
+         )
+      return :synced
+    end
```
misaka, line 34:
```suggestion
    if nhs_immunisations_api_synced_at.present?
      sync_pending_at = nhs_immunisations_api_sync_pending_at
      synced_at = nhs_immunisations_api_synced_at
      if !sync_pending_at.nil? || synced_at > sync_pending_at
        return :synced
      end
    end
```

PR #4064, opened 2025-07-24
Fix back link on vaccination recording flow

murugapl:
The back link always takes you to the page you came from when recording a vaccination from the UI. The user no longer has to click through questions they haven't seen before to return to the patient session page.

https://nhsd-jira.digital.nhs.uk/browse/MAV-249


app/models/concerns/editable_wrapper.rb
```diff
@@ -21,7 +21,7 @@ def read_from!(instance)
     self.editing_id = instance.id
 
     readable_attribute_names
-      .excluding("editing_id")
+      .excluding("editing_id", "first_active_wizard_step")
```
thomasleese, line 24: I think this would be better placed in the `DraftVaccinationRecord#readable_attribute_names` method. The reason being that the `EditableWrapper` concern is used by a number different classes, of which only `DraftVaccinationRecord` has this attribute.

app/models/concerns/editable_wrapper.rb
```diff
@@ -37,7 +37,7 @@ def write_to!(instance)
     end
 
     writable_attribute_names
-      .excluding("editing_id")
+      .excluding("editing_id", "first_active_wizard_step")
```
thomasleese, line 40: Similar here about putting it in `DraftVaccinationRecord#writable_attribute_names`.

app/models/draft_vaccination_record.rb
```diff
@@ -277,6 +278,7 @@ def writable_attribute_names
       programme_id
       session_id
       vaccine_id
+      first_active_wizard_step
```
thomasleese, line 281: Ah just spotted this, I think we should remove this from `writable_attribute_names` because this a list of attributes which can be written to the `VaccinationRecord` model.

thomasleese (commented):
Looks good, I did wonder if we have this problem with any of the other `Draft` models, and whether it could be fixed in a generic way?

app/models/concerns/editable_wrapper.rb
murugapl: I realised I don't actually need it in either method, so removed!

thomasleese:
> > Looks good, I did wonder if we have this problem with any of the other `Draft` models, and whether it could be fixed in a generic way?
> 
> I can't reproduce it on the consent journey, probably because all the wizard steps are shown to the user by default. Otherwise, did you have anything in mind for making it more generic? Another thought was storing all the steps shown to the user initially, not just the first step, so clicking back would always only take you to those steps that you have seen before (if we replace `previous_wizard_path` with `active_steps[index-1]` or equivalent, but this is only needed if we anticipate more complex skip logic in the future/ can be done as an enhancement later?

If it doesn't occur in any other places then I think it's absolutely fine like this.

Something else I'd like to address at some point is making sure the back link is consistent when when you click on a "Change" link on any "Check and confirm" page, as I think it's a bit inconsistent at the moment where the user goes.

PR #4084, opened 2025-07-28
Clarify `in_programmes` scope on patients and patient sessions

thomasleese:
This makes a number of improvements to the `in_programmes` scopes on the `Patient` and `PatientSession` models:

- The `PatientSession` scope no longer needs an academic year. This is because the academic year can come from the associated session. This prevents the scope being used for a different academic year to the session which could cause unusual bugs.
- The logic of the `PatientSession` scope has been fixed to ensure that patients who are eligible for a programme based on their year group, but not eligible in a particular session (perhaps because the year groups aren't administered at the location), don't appear in the query. This fixes the possibility of a patient appearing incorrectly if they go to a school, but aren't eligible for any of the programmes administered at the school. Specifically, this fixes an issue where now that multiple sessions can exist for the same location, if a patient was eligible in one of the sessions they were appearing in all of them regardless of programme.
- Both scopes have been updated to ensure that patients who aren't in any sessions for any particular academic year are excluded from the results, ensuring they don't appear in the wrong place. For example, at the moment, if a patient was eligible for a programme one year, they'll continue to show in all subsequent academic years even if they're no longer eligible or have moved out of the cohort.

The scope has also been renamed `appear_in_programmes` to make it clearer what this scope represents. Initially I was going to go with `eligible_in_programmes` but I wanted to avoid the word eligible as it's not quite correct.

For example, a patient may be eligible for a programme based on their year group, but only exist in sessions where that programme isn't administered at their year group, therefore they remain eligible, but wouldn't show in those sessions for that programme.

The method `Session#programmes_for` already worked like the scopes do after this change, but a mismatch between the scope and this method meant it was possible for a patient to appear in a session, but with no eligible programmes, leading to bugs such as this one:

https://good-machine.sentry.io/issues/6770236206/

[Jira Issue - MAV-1293](https://nhsd-jira.digital.nhs.uk/browse/MAV-1293)
[Jira Issue - MAV-1429](https://nhsd-jira.digital.nhs.uk/browse/MAV-1429)

### DIFF
--- a/app/models/patient_session.rb
+++ b/app/models/patient_session.rb
@@ -21,6 +21,22 @@
 #  fk_rails_...  (session_id => sessions.id)
 #
 
+# The patient session model represents a patient who goes to the school or
+# clinic location represented by the session. This doesn't necessarily mean
+# that the patient is eligible for any of the programmes administered in any
+# particular session they belong to.
+#
+# This is designed to support programmes being dynamically added or removed to
+# sessions without needing to also create or destroy patient session
+# instances. While also adding support for patients becoming eligible if year
+# groups are added or removed to locations, again without needing to also
+# create or destroy patient session instances.
+#
+# It also supports the scenario where a patient belongs to a session but is
+# only eligible for one of many programmes administered in the session. In
+# that case, the list of programmes they appear in won't be the same as the
+# complete list of programmes administered in the session.
+
 class PatientSession < ApplicationRecord
   audited associated_with: :patient
   has_associated_audits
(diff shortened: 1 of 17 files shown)


tvararu (approved):

app/models/session.rb
```diff
@@ -182,8 +182,8 @@ def vaccine_methods
     programmes.flat_map(&:vaccine_methods).uniq.sort
   end
 
-  def programmes_for(patient: nil, year_group: nil)
-    year_group ||= patient.year_group
+  def programmes_for(year_group: nil, patient: nil, academic_year: nil)
+    year_group ||= patient.year_group(academic_year:)
```
misaka, line 186: I don't see any tests for this method. Do you reckon the existing testing covers it? Or could there be some edge-cases, especially with the addition of `academic_year`, that haven't been covered by other tests?

app/models/patient_session.rb
```diff
@@ -72,9 +88,24 @@ class PatientSession < ApplicationRecord
...
-          joins(:patient).merge(
-            Patient.appear_in_programmes(programmes, academic_year:)
+        ->(programmes) do
+          # Is the patient eligible for any of those programmes by year group?
+          location_programme_year_groups =
+            Location::ProgrammeYearGroup
+              .where("programme_id = session_programmes.programme_id")
+              .where("location_id = sessions.location_id")
+              .where(
+                "year_group = sessions.academic_year - patients.birth_academic_year - 5"
```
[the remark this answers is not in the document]
thomasleese: Yes, that's right. I can extract it in to a variable.

app/models/session.rb
thomasleese: I think it's covered by existing tests for the current academic year, tests for upcoming academic years will be added in #2712.

thomasleese: I linked to the wrong PR originally, have fixed the reference to one that isn't already merged in 😆

app/models/patient_session.rb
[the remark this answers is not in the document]
misaka: Thanks. Unfortunate that it turns it into a multi-liner though. 😞

PR #4105, opened 2025-07-30
Remove academic year default assignment in CreateSession CLI command

murugapl:
Move academic year default from argument definition to call method to ensure Rails is loaded before accessing AcademicYear model

### DIFF
--- a/app/lib/mavis_cli/organisations/create_sessions.rb
+++ b/app/lib/mavis_cli/organisations/create_sessions.rb
@@ -11,10 +11,9 @@ module MavisCLI
 
       option :academic_year,
              type: :integer,
-             desc: "The academic year to create the sessions for",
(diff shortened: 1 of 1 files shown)


app/lib/mavis_cli/organisations/create_sessions.rb
```diff
@@ -24,6 +23,8 @@ def call(ods_code:, academic_year:)
           return
         end
 
+        academic_year = AcademicYear.pending if academic_year.nil?
```
thomasleese, line 26:
```suggestion
        academic_year ||= AcademicYear.pending
```

I think this looks slightly neater in my opinion.

thomasleese (approved):

thomasleese (approved):

PR #4108, opened 2025-07-30
Add URN/postcode to school options drop down

murugapl:
When schools have the same name, the autocomplete controller only finds the first match with the name and renders all other options with the same name as a duplicate of the first. This means the user will not be able to see or select the other options. Making the option names uniqe (e.g. by adding the URN) ensures all the options render and can be selected. We only add the URN for SAIS view and postcode for parent view when schools have the same name, not for all schools.

Screenshot 2025-07-30 111942
Screenshot 2025-07-30 112243

https://nhsd-jira.digital.nhs.uk/browse/MAV-1131

### DIFF
--- a/app/helpers/locations_helper.rb
+++ b/app/helpers/locations_helper.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module LocationsHelper
+  def location_display_name(location, show_postcode: false, show_urn: false)
+    return location.name unless show_postcode || show_urn
+
+    identifier =
+      show_postcode ? location.address_postcode : "URN: #{location.urn}"
+    "#{location.name} (#{identifier})"
+  end
+end
--- a/app/views/draft_class_imports/session.html.erb
+++ b/app/views/draft_class_imports/session.html.erb
@@ -13,7 +13,8 @@
     <%= tag.option "", value: "" %>
 
     <% @session_options.each do |session| %>
-      <%= tag.option session.location.name,
+      <% is_duplicate = @session_options.count { |s| s.location.name == session.location.name } > 1 %>
+      <%= tag.option location_display_name(session.location, show_urn: is_duplicate),
                      value: session.id,
                      selected: session.id == @draft_class_import.session_id,
                      data: { hint: format_address_single_line(session.location) } %>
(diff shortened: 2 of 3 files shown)


app/helpers/schools_helper.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module SchoolsHelper
+  def school_display_name(school, duplicate_schools = [], show_postcode: false)
```
thomasleese, line 4:
```suggestion
  def location_display_name(school, duplicate_schools = [], show_postcode: false)
```

Since I don't think there's anything specific about schools here, and in theory when we have community clinics they could also have duplicate names?

For community clinic we would show ODS code instead of URN.

app/views/draft_class_imports/session.html.erb
```diff
@@ -13,7 +13,7 @@
     <%= tag.option "", value: "" %>
 
     <% @session_options.each do |session| %>
-      <%= tag.option session.location.name,
+      <%= tag.option school_display_name(session.location, @duplicate_school_names),
```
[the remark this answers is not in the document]
murugapl: Ah I like that!

app/views/draft_class_imports/session.html.erb
```diff
@@ -13,7 +13,8 @@
     <%= tag.option "", value: "" %>
 
     <% @session_options.each do |session| %>
-      <%= tag.option session.location.name,
+      <% is_duplicate = @session_options.count { |s| s.location.name == session.location.name } > 1 %>
+      <%= tag.option location_display_name(session.location, show_postcode: false, show_urn: is_duplicate),
```
thomasleese, line 17:
```suggestion
      <%= tag.option location_display_name(session.location, show_urn: is_duplicate),
```

Since it defaults to `false`.

app/helpers/locations_helper.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module LocationsHelper
+  def location_display_name(school, show_postcode: false, show_urn: false)
```
thomasleese, line 4:
```suggestion
  def location_display_name(location, show_postcode: false, show_urn: false)
```

I would suggest changing this too, but the method name looks good now.

thomasleese (approved):

PR #4110, opened 2025-07-30
Add `theme-color` meta tag, adapt based on hosting environment

paulrobertlloyd:
If I’ve got this right, this means that we can correctly set a `theme-color` of NHS brand blue (`#005eb8`) on the production environment, otherwise the `theme-color` will share that of the banner at the top of the page.

### DIFF
--- a/app/lib/hosting_environment.rb
+++ b/app/lib/hosting_environment.rb
@@ -1,9 +1,35 @@
 # frozen_string_literal: true
 
 module HostingEnvironment
+  ENVIRONMENT_COLOUR = {
+    production: "blue",
+    development: "white",
+    review: "purple",
+    test: "red",
+    qa: "orange",
+    preview: "yellow"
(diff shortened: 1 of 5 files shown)


app/lib/hosting_environment.rb
```diff
@@ -1,11 +1,37 @@
...
+    preview: "#fff59d"
+  }.freeze
+
   class << self
     def name
       pull_request ? "review" : ENV.fetch("SENTRY_ENVIRONMENT", "development")
     end
 
+    def colour
+      ENVIRONMENT_COLOUR[name.to_sym]
```
thomasleese, line 28:
```suggestion
      ENVIRONMENT_COLOUR.fetch(name.to_sym, "white")
```

I suspect you'll need this for the tests to pass.

thomasleese (approved):

PR #4135, opened 2025-07-31
Add error message when picking an invalid organisation

thomasleese:
This adds a new form which handles the select an organisation page, ensuring that we validate that the user has selected a valid organisation and showing an error message if not.

This fixes an issue where if you selected an invalid organisation or didn't select an organisation, then the user would be taken to an error page.

Users don't see this page yet as it's not enabled when using CIS2, but they will once users can be part of multiple teams.

## Screenshot

Screenshot 2025-07-31 at 19 24 36

### DIFF
--- a/app/forms/select_organisation_form.rb
+++ b/app/forms/select_organisation_form.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class SelectOrganisationForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attr_accessor :current_user, :request_session
+
+  attribute :organisation_id, :integer
+
+  validates :organisation_id, inclusion: { in: :organisation_id_values }
+
+  def save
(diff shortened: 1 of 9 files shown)


app/forms/select_organisation_form.rb
```diff
@@ -0,0 +1,35 @@
...
+    return false if invalid?
+
+    request_session["cis2_info"] = {
+      "selected_org" => {
+        "name" => organisation.name,
+        "code" => organisation.ods_code
+      },
+      "selected_role" => {
+        "code" => User::CIS2_NURSE_ROLE,
+        "workgroups" => ["schoolagedimmunisations"]
```
benilovj, line 23:
```suggestion
        "workgroups" => [AuthenticationConcern::CIS2_WORKGROUP]
```

benilovj (approved):

PR #4145, opened 2025-08-01
Add `notify_parents`, and cache as a column on `VaccinationRecord`

alistair-white-horne-tng:
`notify_parents?` is a function which decides whether a patient's parents should be notified about a given vaccination record. This will be useful in [MAV-1596](https://nhsd-jira.digital.nhs.uk/browse/MAV-1596) and [MAV-1674](https://nhsd-jira.digital.nhs.uk/browse/MAV-1674).

This value is then saved to a new column on `VaccinationRecord`. This is useful for two reasons:
1) It freezes this value to what it was at the time that the vaccination record was created. This is an accurate representation of what this value should hold; "at the time of vaccination, the patient didn't want their parents to be notified"
2) Patients who fall into this category will soon need to be exposed in the "Important notices" tab [MAV-1674](https://nhsd-jira.digital.nhs.uk/browse/MAV-1674), which will otherwise require a very heavy database query


app/models/vaccination_record.rb
```diff
@@ -166,6 +167,8 @@ class VaccinationRecord < ApplicationRecord
               less_than_or_equal_to: -> { Time.current }
             }
 
+  before_create :set_notify_parents_value
```
thomasleese, line 170: Generally we try to avoid using callbacks, it can lead to unexpected side-effects, and they're also not run when importing records in bulk. It also means that it's now not possible to create a vaccination record without this callback being run.

I would suggest instead that we run the equivalent of this [when the vaccination records are first created](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/de3864d08401930a30d17dfb19f6fa5833a8aa68/app/controllers/draft_vaccination_records_controller.rb#L104), which also means we already have access to the consents at the time, the academic year, etc.

db/migrate/20250731154006_add_notify_parents_to_vaccination_records.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class AddNotifyParentsToVaccinationRecords < ActiveRecord::Migration[8.0]
+  def up
+    add_column :vaccination_records, :notify_parents, :boolean
```
thomasleese, line 5: Should we default this to `false`? I'm wondering if there's any benefit to it being able to be both `nil` and `false`?

benilovj: From the domain point of view, a default of `true` is more appropriate (as **not** notifying the parents is an exception).

thomasleese: I was wondering about that too. I suggested `false` because sending a confirmation to parents who's patient explicitly asked them not to receive is perhaps worse than accidentally not sending a confirmation in normal circumstances?

app/controllers/concerns/vaccination_mailer_concern.rb
```diff
@@ -59,8 +59,10 @@ def parents_for_vaccination_mailer(vaccination_record)
       ConsentGrouper.call(patient.consents, programme_id:, academic_year:)
 
     parents =
-      if consents.any?(&:via_self_consent?)
-        consents.any?(&:notify_parents_on_vaccination) ? patient.parents : []
+      if !vaccination_record.notify_parents
+        []
```
thomasleese, line 63: I was thinking even above `ConsentGrouper.call`, since we're not using `consents` in this case. Perhaps something like:

```rb
return [] unless patient.send_notifications? && vaccination_record.notify_parents?
```

app/lib/vaccination_notification_criteria.rb
```diff
@@ -0,0 +1,26 @@
...
+    @vaccination_record = vaccination_record
+  end
+
+  def call
+    patient
+      .consents
+      .where(programme:)
+      .not_invalidated
+      .not_withdrawn
+      .all?(&:notify_parents_on_vaccination)
```
thomasleese, line 14: I think for consent from a parent this value will be `nil`.

thomasleese: Just to add to this, I think the logic you had before was correct:

```rb
.all? { it.notify_parents_on_vaccination || it.notify_parents_on_vaccination.nil? }
```

Although I think it might read better like:

```rb
.none? { it.notify_parents_on_vaccination == false }
```

thomasleese (approved):

PR #4160, opened 2025-08-04
Improve message for errors caused by directly visiting draft vaccination record confirmation screen

samcoy3:
When a user logs in to Mavis and visits `/draft-vaccination-record/confirm` without a draft vaccination record in progress, they will now see a more informative error message. Previously a generic page appeared, indicating that there was an error but suggesting that trying again later would fix the issue.

[Jira issue - MAV-1592](https://nhsd-jira.digital.nhs.uk/browse/MAV-1592)

# Screenshots
The new error page looks like this:
image

### DIFF
--- a/app/controllers/draft_vaccination_records_controller.rb
+++ b/app/controllers/draft_vaccination_records_controller.rb
@@ -14,6 +14,7 @@ class DraftVaccinationRecordsController < ApplicationController
 
   include WizardControllerConcern
 
+  before_action :validate_patient_id, only: :show
   before_action :validate_params, only: :update
   before_action :set_batches, if: -> { current_step == :batch }
   before_action :set_locations, if: -> { current_step == :location }
@@ -56,6 +57,12 @@ class DraftVaccinationRecordsController < ApplicationController
   end
 
   private
+
+  def validate_patient_id
+    if current_step == :confirm && @draft_vaccination_record.patient_id.nil?
+      render 'errors/no_patient_with_draft_vaccination_record', status: :unprocessable_entity
+    end
+  end
 
   def validate_params
     if current_step == :date_and_time
--- a/app/views/errors/no_patient_with_draft_vaccination_record.html.erb
+++ b/app/views/errors/no_patient_with_draft_vaccination_record.html.erb
@@ -1,0 +1,22 @@
+<%= h1 "Error: page not available", size: "xl" %>
+
+<p class="nhsuk-body">
+  You do not have a vaccination record open, or it has timed out.
+</p>
+
+<p class="nhsuk-body">
+  To see a vaccination record, you can search for it in Mavis in either:
+</p>
+
+<ul>
+  <li> the Sessions area (through the relevant session)</li>
+  <li> the Children area</li>
+</ul>
+
+<p class="nhsuk-body">
+  You will need to know the child’s name.
+</p>
+
+<p class="nhsuk-body">
+  <a href="/dashboard">Go to the Mavis dashboard</a>
+</p>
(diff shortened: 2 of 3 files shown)


app/views/errors/no_patient_with_draft_vaccination_record.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <li> the Sessions area (through the relevant session)</li>
+  <li> the Children area</li>
+</ul>
+
+<p class="nhsuk-body">
+  You will need to know the child’s name.
+</p>
+
+<p class="nhsuk-body">
+  <a href="/dashboard">Go to the Mavis dashboard</a>
```
thomasleese, line 21:
```suggestion
  <% link_to "Go to the Mavis dashboard", dashboard_path %>
```

app/controllers/draft_vaccination_records_controller.rb
```diff
@@ -57,6 +58,12 @@ def update
 
   private
 
+  def validate_patient_id
+    if current_step == :confirm && @draft_vaccination_record.patient_id.nil?
```
thomasleese, line 62: I think we could try and come up a more generic approach here. Although we've only seen the case in Sentry where a user navigates to `/draft-vaccination-record/confirm`, there's no reason why they couldn't navigate to `/draft-vaccination-record/batch` without a patient or a session.

app/views/errors/no_patient_with_draft_vaccination_record.html.erb
```diff
@@ -0,0 +1,22 @@
+<%= h1 "Error: page not available", size: "xl" %>
+
+<p class="nhsuk-body">
+  You do not have a vaccination record open, or it has timed out.
+</p>
+
+<p class="nhsuk-body">
+  To see a vaccination record, you can search for it in Mavis in either:
```
thomasleese, line 8: We don't typically use the service name ("Mavis") in the body content. I think it would be good to get some input from the designs on this page.

I'm also wondering if we need that much content on this page and if one of our standard error pages will do?

app/controllers/draft_vaccination_records_controller.rb
thomasleese: Thinking about this some more, it's not specific to the `draft-vaccination-record` route either, any of the `draft-` routes could encounter this problem.

spec/controllers/no_draft_page_open_spec.rb
thomasleese, line 1: Feature tests would go in the `spec/features` directory, and in there you don't need to specify `type: :feature` as [it's inferred automatically](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/8b814db86a61f49259c9a2ec99f5ac64f9bef3f0/spec/spec_helper.rb#L188).

spec/controllers/no_draft_page_open_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+feature "Draft vaccination controller errors", type: :feature do
+  [
+    "/draft-vaccination-record/batch",
+    "/draft-vaccination-record/date_and_time",
```
thomasleese, line 6: This path isn't valid, the `date_and_time` step [is mapped to `date-and-time`](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/8b814db86a61f49259c9a2ec99f5ac64f9bef3f0/config/locales/en.yml#L714).

But this got me thinking, do we want to make this check somehow lower priority, so users see the normal 404 error page if the URL itself doesn't refer to a real step?

app/views/errors/no_draft_page_open.erb
```diff
@@ -0,0 +1,9 @@
+<%= h1 "Error: page not available", size: "xl" %>
+
+<p class="nhsuk-body">
+  You do not have a draft page open, or it has timed out.
```
thomasleese, line 4: I'm not sure about this wording, it would be the first time we make any reference to users to any concept of "drafts", from the perspective of users I'm not sure they'd understand what that means.

PR #4163, opened 2025-08-04
refactor: Change `Triage` table from `triage` to `triages`

MartinVanIJcken:
This is the default naming convention assumed by ActiveRecord. When using a concern that is loaded before the self.table_name is ran, using a non-standard table name breaks access to the table.


thomasleese (approved):
Code change looks good, can we raise a ticket for this so we can assign it to a release?

PR #4176, opened 2025-08-05
Show gillick-no-notify patients in important notices

alistair-white-horne-tng:
This change creates another important notice type for patients who gave consent under Gillick competence, and have requested that their parents aren't notified.

This triggered a refactor of the notification text.

image

This is also exposed as a warning on the patient's page:
image

[MAV-1674](https://nhsd-jira.digital.nhs.uk/browse/MAV-1674)


spec/helpers/patient_notifications_helper_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+          programme: other_programme,
+          notify_parents: true
+        )
+      end
+      let(:no_notify_record) do
+        create(:vaccination_record, patient:, programme:, notify_parents: false)
+      end
+
+      before do
+        notify_record
```
thomasleese, line 135: You can use `let!` to avoid the need to do this, e.g. `let!(:notify_record)`. However, if the values aren't being used in the test it's okay to put them in a `before` block without a name.

mikejamesthompson:
I'm not convinced it makes sense to show the important notice on the top of the child record in this instance because it relates to an individual vaccination record rather than the child as a whole. Perhaps @paulrobertlloyd could review this aspect of the design?

app/models/patient.rb
```diff
@@ -357,6 +372,10 @@ def restricted?
     restricted_at != nil
   end
 
+  def has_vaccination_records_dont_notify_parents?
```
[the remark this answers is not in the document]
alistair-white-horne-tng: I see what you're saying about the database queries, but I quite liked that it can be in the same format as the other notices; simply having:

`if patient.deceased?`
`if patient.invalidated?`
`if patient.restricted?`
and then
`if patient.has_vaccination_records_dont_notify_parents?`

However, also happy to go with your idea if you still think it's worth it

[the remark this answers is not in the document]
thomasleese: I'm concerned about it's use on the "Important notices" page as the current approach means a database query for each patient. I think we could refactor the model and the helper overall though to pre-load the vaccination records for each patient before rendering the page using `includes(:vaccination_records)`.

alistair-white-horne-tng:
@mikejamesthompson , to complete the audit trail, your concern has been worked through [here](https://teamgoodmachine.slack.com/archives/C042JB168BU/p1754469118977409)

app/components/app_notices_table_component.rb
```diff
@@ -4,55 +4,39 @@ class AppNoticesTableComponent < ViewComponent::Base
...
+          @has_vaccination_records_dont_notify_parents_patients
+      ).uniq
+
+    notices =
+      all_patients.flat_map do |patient|
+        helpers
+          .patient_important_notices(patient)
+          .map { |notification| notification.merge(patient:) }
+      end
+    notices.sort_by { _1[:date_time] }.reverse
```
thomasleese, line 40:
```suggestion
    notices.sort_by { it[:date_time] }.reverse
```

app/helpers/patients_helper.rb
```diff
@@ -52,4 +52,54 @@ def patient_year_group(patient, academic_year:)
...
+        date_time: patient.restricted_at,
+        message: "Record flagged as sensitive"
+      }
+    end
+
+    if patient.has_vaccination_records_dont_notify_parents?
+      vaccination_records =
+        patient.vaccination_records.select { it.notify_parents == false }
+
+      if vaccination_records.any?
```
thomasleese, line 84: I think logically this is doing the same as `patient.has_vaccination_records_dont_notify_parents?`, do we need to do the check twice, can we remove the call to `has_vaccination_records_dont_notify_parents?`?

thomasleese (approved):

PR #4178, opened 2025-08-05
Show expired flu consents in activity log

MartinVanIJcken:
This implements [MAV-1711](https://nhsd-jira.digital.nhs.uk/browse/MAV-1711) and follows up on [PR-4053]( https://github.com/nhsuk/manage-vaccinations-in-schools/pull/4053)

### DIFF
--- a/app/components/app_activity_log_component.rb
+++ b/app/components/app_activity_log_component.rb
@@ -318,59 +318,87 @@ class AppActivityLogComponent < ViewComponent::Base
   def decision_expiration_events
     all_programmes = Programme.all.to_a
 
-    AcademicYear.all.filter_map do |academic_year|
-      next if academic_year >= AcademicYear.current
+    AcademicYear.all.flat_map do |academic_year|
+      next [] if academic_year >= AcademicYear.current
 
-      vaccinated_programmes =
-        all_programmes.select do |programme|
+      not_vaccinated_programmes =
+        all_programmes.reject do |programme|
           patient.vaccination_status(programme:, academic_year:).vaccinated?
         end
 
-      programmes_with_expired_consents =
-        consents
-          .select { it.academic_year == academic_year }
-          .flat_map { programmes_for(it) }
-          .reject { vaccinated_programmes.include?(it) }
+      vaccinated_but_seasonal_programmes =
+        all_programmes.select do |programme|
+          patient.vaccination_status(programme:, academic_year:).vaccinated? &&
+            programme.seasonal?
+        end
 
-      programmes_with_expired_triages =
-        triages
-          .select { it.academic_year == academic_year }
-          .flat_map { programmes_for(it) }
-          .reject { vaccinated_programmes.include?(it) }
+      expired_items =
+        {
+          vaccinated_but_seasonal: vaccinated_but_seasonal_programmes,
+          not_vaccinated: not_vaccinated_programmes
+        }.transform_values do |programmes|
+          get_expired_items(academic_year:, programmes:)
+        end
 
-      programmes_with_expired_psds =
-        patient_specific_directions
-          .select { it.academic_year == academic_year }
-          .flat_map { programmes_for(it) }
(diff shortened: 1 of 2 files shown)


app/components/app_activity_log_component.rb
```diff
@@ -302,60 +302,105 @@ def programmes_by_id
 
   def decision_expiration_events
     all_programmes = Programme.all.to_a
+    events = []
 
-    AcademicYear.all.filter_map do |academic_year|
+    AcademicYear.find_each do |academic_year|
```
thomasleese, line 307: You could use `flat_map` here, so you don't need to append to an array.

```rb
AcademicYear.all.flat_map do |academic_year|
  # ...
end
```

It works like a normally `map`, but the return values from the block are concatenated together.

app/components/app_activity_log_component.rb
```diff
@@ -302,60 +302,105 @@ def programmes_by_id
...
+            programmes: not_vaccinated_programmes
+          )
+      }
 
-      programmes_with_expired_items = [
-        programmes_with_expired_consents,
-        programmes_with_expired_triages,
-        programmes_with_expired_psds
-      ].flatten.uniq
+      %i[not_vaccinated vaccinated_but_seasonal].each do |category|
```
thomasleese, line 334: Rather than needing to iterate over the keys and then look them up later, I think we could do something like:

```rb
expired_items.each do |category, category_expired_items|
  # ...
end
```

app/components/app_activity_log_component.rb
```diff
@@ -302,60 +302,105 @@ def programmes_by_id
...
+        }
+      end
     end
+
+    events
+  end
+
+  private
+
+  def get_expired_items(academic_year:, programmes:)
```
thomasleese, line 381: This is a nice refactor, but I think we could go further given that each one has the same three lines. Something like:

```rb
def filter_expired(objects, academic_year:, programmes:)
  objects
    .select { it.academic_year == academic_year }
    .flat_map { programmes_for(it) }
    .select { programmes.include?(it) }
end

def expired_objects(academic_year:, programmes:)
  {
    consents: filter_expired(consents, academic_year:, programmes:),
    triages: filter_expired(triages, academic_year:, programmes:),
    # etc
  }
end
```

app/lib/academic_year.rb
```diff
@@ -15,6 +15,10 @@ def first = [2024, current].min
 
     def all = (first..last).to_a.reverse
 
+    def find_each(&block)
```
thomasleese, line 18: I don't mind adding this method, but I think it would make sense to add it in its own commit (since it's not specifically related to the UI changes) and adding some basic tests for it.

thomasleese (commented):
Overall looks good, I've got some suggestions just to tidy up the code a bit.

app/lib/academic_year.rb
thomasleese: Although I do wonder if `each` is a better name. `find_each` means something specific in Rails [where it will iterate over objects in batches](https://api.rubyonrails.org/v8.0.2/classes/ActiveRecord/Batches.html) re-using the same in-memory objects, which isn't the case here.

MartinVanIJcken: given your flat_map suggestion, I will get rid of this function

app/components/app_activity_log_component.rb
```diff
@@ -318,59 +318,93 @@ def programmes_by_id
...
+                1.second,
+            programmes:
+          }
+        end
       end
-      expired_items << "triage outcome" if programmes_with_expired_triages.any?
-      expired_items << "PSD status" if programmes_with_expired_psds.any?
 
-      next if expired_items.empty?
+    result.compact
```
thomasleese, line 389: If we do `next []` inside the block, we don't need to worry about calling `compact`, it also means the linter won't complain about calling `.compact` and `end` 😆

PR #4184, opened 2025-08-05
Fix incorrect reference to organisations

thomasleese:
These have now been renamed to `Team` and so we need to do the same in this temporary Rake task.


mikejamesthompson (approved):
👍

PR #4209, opened 2025-08-07
Show notice in session if any patients are missing NHS number

MartinVanIJcken:
This is [MAV-1604](https://nhsd-jira.digital.nhs.uk/browse/MAV-1604). This warning is desired now as without NHS numbers, we cannot use the imms api to automatically check that the child has not already been vaccinated. The PR also resolves [MAV-1748](https://nhsd-jira.digital.nhs.uk/browse/MAV-1748) by updating the session overview page to be like in the prototype.

image

image

image

### DIFF
--- a/app/components/app_session_actions_component.rb
+++ b/app/components/app_session_actions_component.rb
@@ -29,12 +29,20 @@ class AppSessionActionsComponent < ViewComponent::Base
 
   def rows
     @rows ||= [
+      no_nhs_number_row,
       no_consent_response_row,
       conflicting_consent_row,
       triage_required_row,
       (register_attendance_row if session.requires_registration?),
       ready_for_vaccinator_row
     ].compact
+  end
+
+  def no_nhs_number_row
+    count = patient_sessions.merge(Patient.without_nhs_number).count
+    href = session_consent_path(session, missing_nhs_number: true)
+
+    generate_row(:children_without_nhs_number, count:, href:)
   end
 
   def no_consent_response_row
--- a/app/components/app_session_actions_component.rb
+++ b/app/components/app_session_actions_component.rb
@@ -38,75 +38,38 @@ class AppSessionActionsComponent < ViewComponent::Base
   end
 
   def no_consent_response_row
-    consent_row("No consent response", status: "no_response")
+    status = "no_response"
+    count =
+      patient_sessions.has_consent_status(status, programme: programmes).count
+    href = session_consent_path(session, consent_statuses: [status])
+
+    generate_row(:children_with_no_consent_response, count:, href:)
   end
 
   def conflicting_consent_row
-    consent_row("Conflicting consent", status: "conflicts")
-  end
-
-  def consent_row(text, status:)
+    status = "conflicts"
     count =
       patient_sessions.has_consent_status(status, programme: programmes).count
-
-    return nil if count.zero?
-
     href = session_consent_path(session, consent_statuses: [status])
 
-    {
-      key: {
-        text: text
-      },
-      value: {
-        text: I18n.t("children", count:)
-      },
-      actions: [{ text: "Review", visually_hidden_text: text.downcase, href: }]
-    }
+    generate_row(:children_with_conflicting_consent_response, count:, href:)
   end
 
   def triage_required_row
     status = "required"
-
     count =
       patient_sessions.has_triage_status(status, programme: programmes).count
-
-    return nil if count.zero?
-
     href = session_triage_path(session, triage_status: status)
 
-    {
-      key: {
-        text: "Triage needed"
-      },
-      value: {
-        text: I18n.t("children", count:)
-      },
-      actions: [
-        { text: "Review", visually_hidden_text: "triage needed", href: }
-      ]
-    }
+    generate_row(:children_requiring_triage, count:, href:)
   end
 
   def register_attendance_row
     status = "unknown"
-
     count = patient_sessions.has_registration_status(status).count
-
-    return nil if count.zero?
-
     href = session_register_path(session, register_status: status)
 
-    {
-      key: {
-        text: "Register attendance"
-      },
-      value: {
-        text: I18n.t("children", count:)
-      },
-      actions: [
-        { text: "Review", visually_hidden_text: "register attendance", href: }
-      ]
-    }
+    generate_row(:children_to_register, count:, href:)
   end
 
   def ready_for_vaccinator_row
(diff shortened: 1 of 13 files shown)


app/components/app_session_actions_component.rb
```diff
@@ -37,15 +42,52 @@ def rows
...
+        text:
+          helpers.link_to(
+            I18n.t(:children_without_nhs_number, count:),
+            href
+          ).html_safe
+      }
+    }
+  end
   def no_consent_response_row
-    consent_row("No consent response", status: "no_response")
```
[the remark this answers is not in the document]
MartinVanIJcken: We could but the refactor does not make semantic sense to me. The repetition really is between each of the rows having the same structure, not so much between these two rows having a similar json they need to generate by the end. The fact that these two rows happen to compute their count and links in a similar way doesn't seem to be very meaningful.

I'd consider a refactor like this

```
  def generate_row(key, count, href: nil, actions: nil)
    {
      key: {
        text: I18n.t("#{key}.title_text")
      },
      value: {
        text:
          if href
            helpers.link_to(I18n.t(key, count:), href).html_safe
          else
            text
          end
      },
      actions:
    }
  end
```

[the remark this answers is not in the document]
thomasleese: Yep your suggestion works too!

app/models/session.rb
```diff
@@ -286,6 +286,10 @@ def open_for_consent?
     close_consent_at&.today? || close_consent_at&.future? || false
   end
 
+  def needs_review?
```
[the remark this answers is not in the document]
thomasleese: If it belongs in all the tabs then I would suggest extracting it out in its own component (again without the method on the model since that's presentational logic) and then rendering the component in each of the views.

app/components/SessionNeedsReviewWarningComponent.rb
```diff
@@ -0,0 +1,30 @@
+class SessionNeedsReviewWarningComponent < ViewComponent::Base
```
thomasleese, line 1: We prefix components with `App` to distinguish them from the standard `Govuk` components we use.

app/components/SessionNeedsReviewWarningComponent.rb
```diff
@@ -0,0 +1,30 @@
...
+    </div>
+  ERB
+
+  def initialize(session:)
+    super
+    @session = session
+  end
+
+  def render?
+    @session.patients.without_nhs_number.exists?
```
thomasleese, line 28: Similar to my other comment, we need to make sure we're counting these correctly and excluding patients who are in the session because they go to the school, but not in the right year group for any of the programmes.

app/components/app_session_needs_review_warning_component.rb
```diff
@@ -0,0 +1,41 @@
...
+    <div class="nhsuk-warning-callout">
+      <h3 class="nhsuk-warning-callout__label">
+        <span role="text">
+          <span class="nhsuk-u-visually-hidden">Important: </span>
+          Needs review
+        </span>
+      </h3>
+
+      <ul>
+        <% if @session.patients.without_nhs_number.count > 0 %>
```
thomasleese, line 14:
```suggestion
        <% if patient_sessions.merge(Patient.without_nhs_number).exists? %>
```

Although do we need this line with the `render?` method?

app/components/app_session_needs_review_warning_component.rb
```diff
@@ -0,0 +1,41 @@
...
+          <li>
+            <%= link_to t(:children_without_nhs_number, count: @session.patients.without_nhs_number.count),
```
thomasleese, line 16:
```suggestion
            <%= link_to t(:children_without_nhs_number, count: patient_session.merge(Patient.without_nhs_number).count),
```

It feels like we need this value in 3 places so it might be worth memoizing it in a method.

```rb 
def without_nhs_number_count
  @without_nhs_number_count ||= ...
```

spec/components/app_session_needs_review_warning_component_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    context "when session has patients without NHS number" do
+      before { create(:patient, nhs_number: nil, session:) }
+
+      it "renders the component" do
+        expect(component.render?).to be(true)
+      end
+
+      it "renders the warning callout" do
```
thomasleese, line 27: I wouldn't expect this test to be in the `describe "#render?"` block as it's not testing that method.

spec/components/app_session_needs_review_warning_component_spec.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+describe AppSessionNeedsReviewWarningComponent do
+  subject(:rendered) { render_inline(component) }
+
+  let(:component) { described_class.new(session:) }
+  let(:session) { create(:session) }
+
+  describe "#render?" do
```
thomasleese, line 9: Inside here we can define this as the subject of the tests:

```rb 
subject { component.render? }
```

And then shorten the amount code needed below by using the short-hand syntax:

```rb
it { should be(true) }
```

spec/components/app_session_needs_review_warning_component_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        expect(rendered).to have_text(
+          I18n.t(:children_without_nhs_number, count: 1)
+        )
+      end
+    end
+
+    context "when session has multiple patients without NHS number" do
+      before do
+        # Create 3 patients without NHS numbers and associate them with the session
+        3.times { create(:patient, nhs_number: nil, session:) }
```
thomasleese, line 62:
```suggestion
        create_list(:patient, 3, nhs_number: nil, session:)
```

spec/components/app_session_needs_review_warning_component_spec.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+describe AppSessionNeedsReviewWarningComponent do
+  subject(:rendered) { render_inline(component) }
```
thomasleese, line 4: It feels like this subject belongs in the `describe "rendered content"` block as that's the subject of the tests in there.

spec/features/filter_state_persistence_spec.rb
```diff
@@ -101,27 +101,34 @@ def and_i_click_the_vaccinated_link
...
   end
 
   def and_i_click_the_register_attendance_link
-    click_on "Review register attendance"
+    click_on "1 child to register"
   end
 
   def and_i_click_the_ready_for_vaccinator_link
-    click_on "Review ready for vaccinator"
+    click_on sprintf "1 child for %s", @programme.name
```
thomasleese, line 120:
```suggestion
    click_on "1 child for #{@programme.name}"
```

It's not very typical to use `sprintf`.

app/components/app_session_needs_review_warning_component.rb
MartinVanIJcken: we don't need it

app/components/app_session_needs_review_warning_component.rb
MartinVanIJcken: We now cache a warnings object that contains this information

spec/components/app_session_needs_review_warning_component_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        expect(rendered).to have_text(
+          I18n.t(:children_without_nhs_number, count: 1)
+        )
+      end
+    end
+
+    context "when session has multiple patients without NHS number" do
+      before do
+        # Create 3 patients without NHS numbers and associate them with the session
+        create(:patient, 3, nhs_number: nil, session:)
```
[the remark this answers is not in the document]
MartinVanIJcken: agreed

PR #4275, opened 2025-08-14
Add /api/reporting controllers, config and tests

aldavidson:
These controllers are disabled by default, through the `:reporting_api` feature flag.
As an extra safeguard against accidental enablement, for now they are also only routable on non-production environments, regardless of the feature flag settings.

**Do Not Merge Yet** - as this adds API endpoints, we need this to be tested and allocated to a release before merging.
 
Jira-Issue: MAV-1784


aldavidson:
Changing this to "Ready" so it can be reviewed

PR #4280, opened 2025-08-14
PDS extravaganza test

benilovj:
A spec coverage gap spun out from #4249. I wanted to check that this test also passes in the production branch and it's independent of the refactoring in #4249.


thomasleese (approved):
🥳

thomasleese (approved):

PR #4345, opened 2025-08-20
Add inference of patient local authority from postcode

aldavidson:
Allow postcodes to be mapped to LocalAuthority records

Adds:
* LocalAuthorityPostcode model and DB migration
* has_one :local_authority_from_postcode relation to Location
* LocalAuthority.for_postcode(..) method, to find a matching LA for a given postcode
* download/import Mavis CLI tasks for the postcode -> LA mappings.  

Note that the postcode mappings file is large - c450MB when unzipped - which meant
that the simplistic approach of reading the whole CSV file into RAM and parsing it to 
an array was unsuitable in this case. Doing it this way, parsing line-by-line, keeps the memory usage to a minimum -

Screenshot 2025-08-19 at 16 24 52

- at the cost of making it not practical to count the number of lines ahead of time. As a compromise, this progress bar is based on the current stream cursor position in bytes, vs the total stream size.


app/lib/mavis_cli/local_authorities/import_postcode_mappings.rb
```diff
@@ -0,0 +1,72 @@
+#frozen_string_literal: true
+
+module MavisCLI
+  module LocalAuthorities
+    class ImportPostcodeMappings < Dry::CLI::Command
+      desc "Import LocalAuthorities data"
```
thomasleese, line 6: I think this is not quite right!

app/models/local_authority.rb
```diff
@@ -32,6 +32,8 @@ class LocalAuthority < ApplicationRecord
   validates :gias_code, uniqueness: true, allow_nil: true
   validates :gss_code, uniqueness: true, allow_nil: true
 
+  has_many :local_authority_postcodes, foreign_key: :gss_code, primary_key: :gss_code
```
thomasleese, line 35:
```suggestion
  has_many :postcodes, foreign_key: :gss_code, primary_key: :gss_code, class_name: "LocalAuthorityPostcode"
```

Just to make using this association slightly nicer for developers 😄

app/models/local_authority.rb
```diff
@@ -73,4 +75,8 @@ def self.from_my_society_import_row(data)
       end_date: data["end-date"]
     )
   end
+
+  def self.for_postcode(postcode)
+    joins(:local_authority_postcodes).where(local_authority_postcodes: {postcode: LocalAuthorityPostcode.normalize_postcode(postcode)}).first
```
thomasleese, line 80: I think if you do something like this you shouldn't need to call `normalize_postcode` as it'll be done automatically by `normalizes`.

```rb
joins(:local_authority_postcodes).merge(LocationAuthorityPostcode.where(postcode:))
```

db/migrate/20250819132430_add_local_authority_postcodes.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class AddLocalAuthorityPostcodes < ActiveRecord::Migration[8.0]
+  def change
+    create_table :local_authority_postcodes,
+                 id: false,
+                 primary_key: %i[gss_code postcode] do |t|
+      t.string :gss_code, null: false
+      t.string :postcode, null: false
```
thomasleese, line 9: I'm wondering if `value` is a better name here to avoid code like `postcode.postcode`. It's [similar to what we do for `SessionDates`](https://github.com/nhsuk/manage-vaccinations-in-schools/blob/906e02eb24a9a3b04af6b25ef2d5e4d6cd33438b/app/models/session_date.rb#L8).

db/migrate/20250819132430_add_local_authority_postcodes.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.timestamps
+
+      t.index %i[gss_code postcode], unique: true
+      t.index %i[postcode gss_code], unique: true
```
thomasleese, line 13: You shouldn't need to add both indexes as one will cover the other.

app/models/local_authority_postcode.rb
```diff
@@ -0,0 +1,32 @@
...
+#  postcode   :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+# Indexes
+#
+#  index_local_authority_postcodes_on_gss_code_and_postcode  (gss_code,postcode) UNIQUE
+#  index_local_authority_postcodes_on_postcode_and_gss_code  (postcode,gss_code) UNIQUE
+#
+class LocalAuthorityPostcode < ApplicationRecord
```
thomasleese, line 17:
```suggestion
LocalAuthority::Postcode
```

It could be nice to make this a nested class, since it's so closely related to the `LocalAuthority` model and helps to keep `app/models` well organised.

aldavidson: That makes sense - I did find the name LocalAuthorityPostcode a bit of a mouthful, but couldn't think of a better one, so I will go with this 👍

db/migrate/20250819132430_add_local_authority_postcodes.rb
thomasleese: Ah yep, that makes sense!

aldavidson:
rebased to latest `next`

spec/models/local_authority/postcode_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+# Indexes
+#
+#  index_local_authority_postcodes_on_gss_code  (gss_code)
+#  index_local_authority_postcodes_on_value     (value) UNIQUE
+#
+describe LocalAuthority::Postcode, type: :model do
+  it { should normalize(:value).from(" Sw111AA ").to("SW11 1AA") }
+  it { should normalize(:value).from(" Sw11.   1aA ").to("SW11 1AA") }
+
+  describe "normalize_postcode" do
```
thomasleese, line 21: Do we need to test this function on its own, it feels like to me like it should be private (or even just directly in the block) and we can test it using `should normalize(:value)` like the ones above?

aldavidson: Yeah, I originally made it a standalone method so that it could be called from the `LocalAuthority.for_postcode(..)` method - but now that external call is not needed (thanks to the merged scope) it can go back inside the block

spec/models/local_authority/postcode_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+#  value      :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+# Indexes
+#
+#  index_local_authority_postcodes_on_gss_code  (gss_code)
+#  index_local_authority_postcodes_on_value     (value) UNIQUE
+#
+describe LocalAuthority::Postcode, type: :model do
```
thomasleese, line 17:
```suggestion
describe LocalAuthority::Postcode do
```

This is not needed as it's inferred by RSpec based on the directory.

thomasleese (approved):

PR #4347, opened 2025-08-20
Prescribers can create PSD instructions during triage

jhenderson:
### Description

Implements the ability for prescribers (e.g. qualified nurses) to add Patient Specific Directions (PSDs) during the triage process when a child is marked as safe to vaccinate with nasal spray. This feature supports delegation workflows where Healthcare Assistants (HCAs) can administer vaccinations under PSD authority.

[Jira Ticket - MAV-1362](https://nhsd-jira.digital.nhs.uk/browse/MAV-1362)

### Changes proposed
- Added conditional PSD option when selecting "safe to vaccinate with nasal spray" in triage page
- Only appears when:
	- Delegation feature flag is enabled
	- User has permission to create PSDs
	- PSD functionality is enabled for the session
	- Vaccination method is nasal spray
- "PSD added" tag appears on patient page if they have a PSD record

### Screenshots

| Choosing to add PSD | Patient page after PSD added |
|------------------|---------------------------|
| Screenshot 2025-08-20 at 14 13 21  | Screenshot 2025-08-20 at 12 24 16 |

### DIFF
--- a/app/components/app_triage_form_component.html.erb
+++ b/app/components/app_triage_form_component.html.erb
@@ -3,7 +3,15 @@
 
     <%= f.govuk_radio_buttons_fieldset :status_and_vaccine_method, **fieldset_options do %>
       <% triage_form.safe_to_vaccinate_options.each do |option| %>
-        <%= f.govuk_radio_button :status_and_vaccine_method, option %>
+        <%= f.govuk_radio_button :status_and_vaccine_method, option do %>
+        <% if show_psd_options?(option) %>
+          <%= f.govuk_radio_buttons_fieldset :psd_action,
+                                             legend: { text: "Do you want to add a PSD?", size: "s" } do %>
+            <%= f.govuk_radio_button :add_psd, "true", label: { text: "Yes" } %>
+            <%= f.govuk_radio_button :add_psd, "false", label: { text: "No" } %>
+          <% end %>
+        <% end %>
+      <% end %>
       <% end %>
       <%= f.govuk_radio_divider %>
       <% triage_form.other_options.each do |option| %>
--- a/app/models/patient_specific_direction.rb
+++ b/app/models/patient_specific_direction.rb
@@ -7,7 +7,6 @@
 #  id                 :bigint           not null, primary key
 #  academic_year      :integer          not null
 #  delivery_site      :integer          not null
-#  full_dose          :boolean          not null
 #  vaccine_method     :integer          not null
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null
@@ -38,8 +37,6 @@ class PatientSpecificDirection < ApplicationRecord
   belongs_to :patient
   belongs_to :programme
   belongs_to :vaccine
-
-  validates :full_dose, inclusion: { in: [true, false] }
 
   enum :delivery_site,
        {
--- a/app/components/app_triage_form_component.rb
+++ b/app/components/app_triage_form_component.rb
@@ -26,6 +26,13 @@ class AppTriageFormComponent < ViewComponent::Base
 
   def builder = GOVUKDesignSystemFormBuilder::FormBuilder
 
+  def show_psd_options?(option)
+    Flipper.enabled?(:delegation) &&
+      helpers.policy(PatientSpecificDirection).create? &&
+      patient_session.session.psd_enabled? &&
+      option == "safe_to_vaccinate_nasal"
+  end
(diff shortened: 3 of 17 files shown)


app/components/app_triage_form_component.html.erb
```diff
@@ -3,7 +3,15 @@
 
     <%= f.govuk_radio_buttons_fieldset :status_and_vaccine_method, **fieldset_options do %>
       <% triage_form.safe_to_vaccinate_options.each do |option| %>
-        <%= f.govuk_radio_button :status_and_vaccine_method, option %>
+        <%= f.govuk_radio_button :status_and_vaccine_method, option do %>
+        <% if helpers.policy(PatientSpecificDirection).create? && psd_enabled? && option == "safe_to_vaccinate_nasal" %>
+          <%= f.govuk_radio_buttons_fieldset :psd_action,
+                                             legend: { text: "Do you want to add a PSD?", size: "s" } do %>
+            <%= f.govuk_radio_button :add_psd, "true", label: { text: "Yes" } %>
```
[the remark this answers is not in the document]
thomasleese: You can [use `inline` to make them appear side by side](https://govuk-form-builder.netlify.app/form-elements/radios/#inline-radios).

app/models/patient_specific_direction.rb
```diff
@@ -39,7 +39,11 @@ class PatientSpecificDirection < ApplicationRecord
   belongs_to :programme
   belongs_to :vaccine
 
-  validates :full_dose, inclusion: { in: [true, false] }
+  validates :full_dose,
```
thomasleese, line 42: I'm wondering if we even need this column on the PSD? The result will be recorded on the vaccination record rather than the PSD I would imagine.

db/migrate/20250819160135_add_psd_enabled_to_sessions.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPsdEnabledToSessions < ActiveRecord::Migration[8.0]
+  def change
+    add_column :sessions, :psd_enabled, :boolean, default: true
```
thomasleese, line 5: Can this be `not null`?

app/models/cis2_info.rb
```diff
@@ -44,6 +44,10 @@ def can_prescribe_pgd?
     role_code == NURSE_ROLE
   end
 
+  def can_prescribe_psd?
```
thomasleese, line 47: I'm wondering actually if we need to rename `can_prescribe_pgd?` to `can_supply_pgd?` since that covers both the prescribing and the supplying of the vaccine (i.e. what nurses do).

thomasleese, line 47: If we're adding a new permission, we should also add it to https://github.com/nhsuk/manage-vaccinations-in-schools/blob/62ad5d3e976b274360ed4b5ea1d7a13bdc8e119e/app/controllers/concerns/authentication_concern.rb#L51 to make sure that a user who only has permission to do that and nothing else will continue to be able to log in.

db/migrate/20250820145646_add_psd_enabled_to_sessions.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPsdEnabledToSessions < ActiveRecord::Migration[8.0]
+  def change
+    add_column :sessions, :psd_enabled, :boolean, default: true, null: false
```
thomasleese, line 5:
```suggestion
    add_column :sessions, :psd_enabled, :boolean, default: false, null: false
```

Actually I'm wondering if it makes sense to start with this disabled, to avoid accidentally enabling PSD when we shouldn't.

thomasleese (commented):
Just got one comment, otherwise it looks good!

thomasleese (approved):

app/models/user.rb
```diff
@@ -139,6 +139,10 @@ def can_supply_using_pgd?
     cis2_enabled? ? cis2_info.can_supply_using_pgd? : fallback_role_nurse?
   end
 
+  def can_prescribe_psd?
```
jhenderson, line 142: @thomasleese On further reflection, would `can_add_psd`? be a better description because prescribing is more to do with medicines. A PSD is just permission for a HCA to carry out the vaccination.

thomasleese (approved):

PR #4362, opened 2025-08-21
Add `can_administer_without_prescription?`

thomasleese:
This adds a method to `User` and `CIS2Info` which can be used to determine with a particular user has permission to administer a vaccination without a prescription, meaning the vaccine was supplied by a prescriber or nurse.

[Jira Issue - MAV-1365](https://nhsd-jira.digital.nhs.uk/browse/MAV-1365)


app/models/cis2_info.rb
```diff
@@ -44,6 +46,10 @@ def can_supply_using_pgd?
     role_code == NURSE_ROLE
   end
 
+  def can_administer_without_prescription?
```
jhenderson, line 49: Is this related to PSD? Would we use this instead of `can_prescribe_psd?`

thomasleese: My understanding is we would use this once a PSD is in place to determine whether the user is allowed to perform the vaccination.

I wasn't entirely sure about the name, but I based it on the description of the activity code [in the ticket](https://nhsd-jira.digital.nhs.uk/browse/MAV-1365).

> Allows user to provide medication to a patient without issuing a prescription.

jhenderson (approved):

PR #4368, opened 2025-08-21
Deduplicate parents and relationships before incrementing import counts

murugapl:
Parents could appear multiple times in memory during import, causing `parents.any?(&:changed?)` to incorrectly return true even when no data had actually changed. This was due to duplicate Parent instances with the same ID being included in the batch.

Ensure we call `.uniq` on parents and relationships before checking for changes to avoid false positives.

https://nhsd-jira.digital.nhs.uk/browse/MAV-1825

### DIFF
--- a/app/jobs/commit_patient_changesets_job.rb
+++ b/app/jobs/commit_patient_changesets_job.rb
@@ -138,8 +138,8 @@ class CommitPatientChangesetsJob < ApplicationJob
       count_column_to_increment =
         import.count_column(
           changeset.patient,
-          changeset.parents,
-          changeset.parent_relationships
+          changeset.parents.uniq,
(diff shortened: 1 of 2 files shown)


app/jobs/commit_patient_changesets_job.rb
```diff
@@ -33,7 +33,7 @@ def perform(import)
   end
 
   def import_patients_and_parents(changesets, import)
-    patients = changesets.map(&:patient)
+    patients = changesets.map(&:patient).uniq
```
thomasleese, line 36: I think `deduplicate_patients!` is designed to do the same thing, although I'm not sure what that does which is different to `uniq`.

PR #4373, opened 2025-08-21
Validate child names on import

misaka:
PDS restricts the allowable characters in a patient's name to be alphanumeric, including letters used in European alphabets, and the extra characters '-', '.' and "'". Especially now that we perform PDS updates as a part of the import process, we need to ensure that we follow these same guidelines that PDS uses.

This change adds another validation to the import to ensure we only accept the valid characters PDS accepts. The error message, when triggered, looks like:

image

It also adds automatic conversion of "fancy quotes" into "plain quotes" as accepted by PDS. Again, see ticket for more details.

See comments on the Jira ticket for more details.

Jira-Issue: MAV-1807

### DIFF
--- a/app/models/patient_import_row.rb
+++ b/app/models/patient_import_row.rb
@@ -12,6 +12,9 @@ class PatientImportRow
       \u00F8-\u017F  # Latin 1 Supplement & Latin Extended A letters
       \u0020         # Space
       \u0027         # Apostrophe
+      \u0060         # Grave accent, will be normalised to apostrophe
+      \u2019         # Preferred Unicode apostrophe, will be normalised to apostrophe
+      \u02BC         # Modifier apostrophe, will be normalised to apostrophe
       \u002E         # Full stop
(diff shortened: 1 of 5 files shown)


app/models/patient_import_row.rb
```diff
@@ -405,6 +405,8 @@ def validate_first_name
         first_name.header,
         "is greater than #{MAX_FIELD_LENGTH} characters long"
       )
+    elsif !first_name.to_s.match?(PDS::Patient::VALID_NAME_REGEX)
+      errors.add(first_name.header, "includes invalid character(s).")
```
thomasleese, line 409: Minor content thing, I think we've normally not put a full stop at the end of these errors, but maybe we should add it consistently across all of them?

thomasleese (approved):

app/models/patient_import_row.rb
misaka: Let's make it consistent for now. Can ask the designers after

thomasleese (approved):

PR #4378, opened 2025-08-22
Track which users should be shown in list of suppliers

thomasleese:
This adds a new column to the users table which we can use to track which users should appear in the list of suppliers that healthcare assistances can choose from when they administer a nasal flu vaccine.

This is effectively the same as whether or not the user is a nurse, but I've decided to not name this column `is_nurse` for two reasons:

- To avoid confusing with the `nurse` `fallback_role` which is only used in non-CIS2 environments.
- To allow for flexibility in the future in case a new role type is added that can also supply a PGD vaccine to healthcare assistants.

Our users log in using CIS2, and it's only at the point of logging in do we know what permissions a user has, so at that point we will update the this value for them. For non-CIS2 environments, we could in theory rely on the `fallback_role` column instead, but I think to make the code consistent it makes sense to use same column in both cases.

[Jira Issue - MAV-1353](https://nhsd-jira.digital.nhs.uk/browse/MAV-1353)

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -24,6 +24,9 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     else
       @user = User.find_or_create_from_cis2_oidc(user_cis2_info, valid_teams)
 
+      # Track which users have authorisation to supply using the PGD protocol.
+      @user.update!(show_in_suppliers: cis2_info.is_nurse?)
+
       # give them a session token for the reporting app also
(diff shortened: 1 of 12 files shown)


misaka:
I feel like we could use some kind of abstraction for this kind of thing, perhaps roles or similar. Happy to get this through for now until we find a better way.

misaka (approved):

thomasleese:
> I feel like we could use some kind of abstraction for this kind of thing, perhaps roles or similar. Happy to get this through for now until we find a better way.

Yeah, I wasn't 100% happy with this approach, but I agree I think it gets us where we need for now.

PR #4387, opened 2025-08-22
Check birth century

MartinVanIJcken:
Anyone born before 2000 is clearly not going to be in secondary education anymore.

This isn't a perfect cut-off point but it's low enough that it will never be a problem for anyone who still is going to school, but it will prevent uploads of children born in the wrong century, which we sometimes see when the CHILD_YEAR_GROUP parameter is set while the date of birth for example does not have a century set. (e.g. 01/01/0015 may show as 01/01/15 in excel)

This is [MAV-1833](https://nhsd-jira.digital.nhs.uk/browse/MAV-1833)

### DIFF
--- a/spec/models/cohort_import_row_spec.rb
+++ b/spec/models/cohort_import_row_spec.rb
@@ -66,7 +66,7 @@ describe CohortImportRow do
     it { should be_valid }
 
     context "when date of birth is outside the programme year group" do
-      let(:data) { valid_data.merge("CHILD_DATE_OF_BIRTH" => "1990-01-01") }
+      let(:data) { valid_data.merge("CHILD_DATE_OF_BIRTH" => "2000-01-01") }
 
       it "is invalid" do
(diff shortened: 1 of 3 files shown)


thomasleese (commented):
Could we add a test to the class import row or cohort import row specs that covers this?

spec/models/cohort_import_row_spec.rb
```diff
@@ -88,6 +88,23 @@
...
+    context "when date of birth is in the previous century" do
+      let(:data) do
+        valid_data.merge(
+          { "CHILD_DATE_OF_BIRTH" => "1911-01-01", "CHILD_YEAR_GROUP" => "9" }
+        )
+      end
+
+      it "is invalid" do
+        expect(cohort_import_row).to be_invalid
+        puts cohort_import_row.errors.full_messages
```
thomasleese, line 100: I'm not sure you meant to keep this line in. 😄

PR #4393, opened 2025-08-26
Allow healthcare assistants to administer nasal spray via PGD supply

thomasleese:


app/components/app_vaccinate_form_component.html.erb
```diff
@@ -54,6 +54,20 @@
       <% end %>
 
       <%= f.govuk_text_area :pre_screening_notes, label: { text: "Pre-screening notes (optional)" }, rows: 3 %>
+
+      <% if form.requires_supplied_by_user_id? %>
```
jhenderson, line 58: Should I see this field as a nurse? I thought it was for HCAs only.

jhenderson (approved):
LGTM 👍

PR #4401, opened 2025-08-26
Persist all PDS search results for patient changesets

murugapl:
Previously, only the final PDS search result was persisted for a patient changeset when a slow import was processed. In-memory updates to `patient_changeset.search_results` will be lost between job executions if not explicitly saved in these cases.

### DIFF
--- a/app/jobs/process_patient_changesets_job.rb
+++ b/app/jobs/process_patient_changesets_job.rb
@@ -20,15 +20,12 @@ class ProcessPatientChangesetsJob < ApplicationJob
         step: step_name,
         result: result,
         nhs_number: pds_patient&.nhs_number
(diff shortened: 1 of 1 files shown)


app/jobs/process_patient_changesets_job.rb
```diff
@@ -22,6 +22,8 @@ def perform(patient_changeset, step_name = nil)
         nhs_number: pds_patient&.nhs_number
       }
 
+      patient_changeset.save!
+
```
misaka, line 26: Can we have this as an alternative (i.e. `else` section) to the `if` statement below? Just to prevent extra DB hits.

thomasleese (approved):

PR #4405, opened 2025-08-26
Prevent deletion of vaccination records sourced from NHS immunisations API

samcoy3:
This PR prevents all users from deleting vaccination records where `source = nhs_immunisations_api`.

Users are already unable to edit such records, because only records with an associated session can be edited.

Jira-Issue: MAV-1772


app/policies/vaccination_record_policy.rb
```diff
@@ -19,7 +19,7 @@ def update?
   end
 
   def destroy?
-    user.is_superuser?
+    user.is_superuser? && record.source != "nhs_immunisations_api"
```
thomasleese, line 22:
```suggestion
    user.is_superuser? && !record.nhs_immunisations_api?
```

It might be worth adding a `prefix` to the enum (perhaps `sourced_from`), then we could do something like:

```rb
record.sourced_from_nhs_immunisations_api?
```

PR #4415, opened 2025-08-27
Make `VaccinationRecord.protocol` optional from imports and API

alistair-white-horne-tng:
This allows historic records, and records received from the API to leave `protocol = nil`.

Also set imported records' protocol to `nil`.


thomasleese:
> @thomasleese What do you want to do about the existing records which have been imported in the past? Do we want to `nil` out `protocol` for these? I wonder if it would be better to do that after #4346 goes in, because then we can be sure that we're editing the correct records?

Agreed, I think we can handle those later, in fact I don't imagine there are many in production because we only added the `protocol` column recently.

thomasleese (approved):

PR #4418, opened 2025-08-27
Persist `created_at` on PDS search results

murugapl:
Ensures search timestamps are preserved during the import process, rather than defaulting to current time in the final commit job.

### DIFF
--- a/spec/models/pds_search_result_spec.rb
+++ b/spec/models/pds_search_result_spec.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: pds_search_results
+#
+#  id          :bigint           not null, primary key
+#  import_type :string
+#  nhs_number  :string
+#  result      :integer          not null
+#  step        :integer          not null
+#  created_at  :datetime         not null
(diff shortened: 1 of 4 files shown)


spec/models/pds_search_result_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        described_class.new(
+          patient:,
+          import:,
+          step: :no_fuzzy_with_history,
+          result: :one_match,
+          nhs_number: "1234567890",
+          created_at: timestamp
+        )
+
+      described_class.import([record])
```
misaka, line 50: I'm a bit confused by this bit ... we test that if the record is imported it sets the `created_at`?

misaka: I'm just not clear on what it's testing ... it would be nice if we could test that `ProcessPatientChangesetsJob` and `CommitPatientChangesetsJob` adds the `created_at`. Normally we'd rely on it showing up in the UI in a feature test, but since the use-case here is simply that we can check the dates these records are created at as part of ops tasks, maybe it's ok to go straight to the db and check them, in a feature test. What do you think?

PR #4421, opened 2025-08-27
HCAs can record vaccinations for PSD patients

jhenderson:
This trait is not needed anymore as the full_dose column has


app/models/draft_vaccination_record.rb
```diff
@@ -131,7 +131,11 @@ def already_had?
   alias_method :administered, :administered?
 
   def protocol
-    :pgd
+    if session.psd_enabled? && @current_user.is_healthcare_assistant?
```
jhenderson, line 134: Are these two conditions sufficient? We might not need to check if PSD is added here because the HCA wouldn't have been allowed to see the record vaccination section on the patient page if there's no PSD added.

app/forms/vaccinate_form.rb
```diff
@@ -71,7 +71,9 @@ def supplied_by_users
     current_user.selected_team.users.show_in_suppliers
   end
 
-  def requires_supplied_by_user_id? = !current_user.show_in_suppliers
+  def requires_supplied_by_user_id?
+    !patient_session.psd_added?(programme:) && !current_user.show_in_suppliers
```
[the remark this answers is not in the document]
thomasleese: I think we should also check `psd_enabled?`, because if that gets turned off I imagine we would want it to implicitly invalidate any existing PSDs.

app/models/draft_vaccination_record.rb
thomasleese: I think we need to do a more thorough check here, particular in relation to the national protocol (#4409) because I there could be a scenario where a patient has a PSD for nasal flu, the national protocol is enabled, and therefore the HCA decides to vaccinate using the injection, which would mean the protocol is `national`, not `psd`.

jhenderson: Updated the logic including a feature scenario to cover this edge case. Is the logic ok?

jhenderson: If both were enabled, which protocol gets used?

app/forms/vaccinate_form.rb
```diff
@@ -77,7 +77,24 @@ def supplied_by_users
...
+    # Under national protocol, healthcare providers normally need to provide supplier
+    # details when recording a vaccination. However, supplier details are not required
+    # for Healthcare Assistants (HCAs) operating under Patient Specific Direction (PSD)
+    # protocol.
+    operating_under_national_protocol? && user_not_designated_as_supplier?
+  end
+
+  def operating_under_national_protocol?
+    session.national_protocol_enabled? && !session.psd_enabled? &&
+      !patient_session.psd_added?(programme:)
```
jhenderson, line 92: After rebasing off `next`, `patient_session` is no longer available. I'm wondering now if we still need to check if a PSD exists, because even if it did, and the session is changed from PSD to national protocol, we wouldn't want o prevent this method from return `true`. What do you think?

app/components/app_vaccinate_form_component.html.erb
```diff
@@ -104,6 +104,20 @@
                                                      size: "s",
                                                    } %>
             <% end %>
+
+            <% if form.current_user.is_healthcare_assistant? && form.session.national_protocol_enabled? %>
```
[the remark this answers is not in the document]
jhenderson: Good point. Need to factor that in.

app/models/draft_vaccination_record.rb
```diff
@@ -28,7 +28,6 @@ class DraftVaccinationRecord
   attribute :protocol, :string
   attribute :performed_ods_code, :string
   attribute :programme_id, :integer
-  attribute :protocol, :string
```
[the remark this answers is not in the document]
thomasleese: Ahhhh yes!

PR #4430, opened 2025-08-28
Create CLI tool to send `search` request

alistair-white-horne-tng:
This command can be invoked using `bin/mavis vaccination-records search PATIENT_ID PROGRAMMES [DATE_FROM] [DATE_TO] [OUTPUT_FILE]`. It is primarily designed to help the testing of the search functionality, but will also be useful for ops.

This command will only make the search request; it won't do anything with the result (apart from display/save into a file).

[MAV-1261](https://nhsd-jira.digital.nhs.uk/browse/MAV-1261)


app/lib/mavis_cli/vaccination_records/search.rb
```diff
@@ -0,0 +1,103 @@
...
+        Date.parse(value)
+      rescue ArgumentError
+        puts "Error: Invalid #{name} format '#{value}'. Expected YYYY-MM-DD."
+        :error
+      end
+    end
+  end
+
+  register "vaccination-records" do |prefix|
+    prefix.register "search", VaccinationRecords::Search
```
misaka, line 101: I feel like this should be more specific to the imms api, maybe either "vaccination-records search_imms_api" or maybe "imms-api search"?

misaka: Or since we provide a patient id it could be an action on patients "patients search-imms-api"

app/lib/mavis_cli/vaccination_records/search.rb
```diff
@@ -0,0 +1,103 @@
...
+
+          if bundle.nil?
+            puts "No result returned"
+            return
+          end
+
+          json = bundle.to_json
+          pretty = JSON.pretty_generate(JSON.parse(json))
+
+          puts pretty
```
misaka, line 77: I think if we provide an output file to write the JSON to, it shouldn't print it out. That's how the output option works normally, it's an either-or.

thomasleese (approved):

PR #4432, opened 2025-08-28
Separate app and infra deployments

TheOneFromNorway:
Separate deployment of application and infrastructure changes

This enables separating out infrastructure components into a separate repository. Some cleanup is required post deployment, this is not done at this stage to ensure backwards compatibility of the infrastructure.

- Replace the task definition creation with a template version
  - The template version will be picked up and modified during the ECS deployment workflow
- Remove redundant resources/outputs
  - We no longer need outputs/s3 buckets since we will be using AWS provided github actions for deployment
- Move application-specific variables into a config file (to be used in deployment)
  - A new SSM resources is introduced that will allow tweaking
    parameters without redeployment
- Create ECS deployment specific role in AWS
  - Limits permission scope for production deployments
  - Adheres to the principle of least privilege
  - Separates permissions between application deployments and infrastructure changes

Ticket: [MAV-1657](https://nhsd-jira.digital.nhs.uk/browse/MAV-1657)


terraform/account/variables.tf
```diff
@@ -32,4 +32,9 @@ locals {
   monitoring_policies = merge(local.base_policies, {
     monitoring_deploy = aws_iam_policy.monitoring_deploy.arn
   })
+
+  ecs_deploy_policies = {
+    ecr_read           = "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly"
+    deploy_ecs_service = aws_iam_policy.deploy_ecs_service.arn
+  }
```
bogsi17, line 39: This could also be inlined in the deployment_permissions.tf

.github/workflows/data-replication-pipeline.yml
```diff
@@ -15,205 +15,153 @@ on:
...
-        description: CIDR blocks to allow egress traffic.
-        type: string
-        required: true
-        default: "[]"
+
+permissions: {}
 
 env:
-  aws_role: ${{ inputs.environment == 'production'
+  aws-role: ${{ inputs.environment == 'production'
```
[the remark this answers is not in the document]
thomasleese: The convention for GitHub Actions is to use dashes, so I would advocate for us to migrate towards that.

.github/workflows/deploy.yml
```diff
@@ -138,21 +138,21 @@ jobs:
...
       id-token: write
     needs: [deploy-infrastructure, determine-git-sha]
-    if: always() && inputs.server_types != 'none' && needs.deploy-infrastructure.result == 'success'
+    if: ${{ !cancelled() && inputs.server_types != 'none' && needs.deploy-infrastructure.result == 'success' }}
     uses: ./.github/workflows/deploy-application.yml
     with:
       environment: ${{ inputs.environment }}
       server_types: ${{ inputs.server_types }}
       git_sha_to_deploy: ${{ needs.determine-git-sha.outputs.git-sha }}
+      app_version: ${{ inputs.git_ref_to_deploy }}
```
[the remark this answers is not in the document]
TheOneFromNorway: It does not will remove

terraform/app/variables.tf
```diff
@@ -120,90 +104,44 @@ variable "rails_master_key_path" {
...
+    })
   })
-  parameter_store_config_list = [for key, value in local.parameter_store_variables : {
-    name      = key
-    valueFrom = aws_ssm_parameter.environment_config[key].arn
-  }]
-  parameter_store_arns = [for key, value in local.parameter_store_variables : aws_ssm_parameter.environment_config[key].arn]
+  non_empty_parameter_groups = toset([
+    for key, value in local.parameter_store_variables : key if length(concat(local.secret_arns[key], local.parameter_store_paths[key])) > 0
+  ])
```
bogsi17, line 129: The `non_empty_parameter_groups` is just used to dynamically create dedicated iam roles and policies for "CORE" and "REPORTING" right? I would prefer to create the iam resources explicitly and avoid this complex construction

terraform/app/variables.tf
```diff
@@ -120,90 +104,44 @@ variable "rails_master_key_path" {
...
-variable "app_version" {
-  type        = string
-  description = "The version identifier for the MAVIS application deployment"
-  default     = "Unknown"
-  nullable    = false
-}
-
 locals {
   is_production = var.environment == "production"
   parameter_store_variables = tomap({
```
bogsi17, line 116: I think the explicit casts to map are not necessary, also in the other places further down

terraform/app/variables.tf
```diff
@@ -120,90 +104,44 @@ variable "rails_master_key_path" {
...
-    MAVIS__ACADEMIC_YEAR_TODAY_OVERRIDE             = var.academic_year_today_override
-    MAVIS__ACADEMIC_YEAR_NUMBER_OF_PREPARATION_DAYS = var.academic_year_number_of_preparation_days
-    MAVIS__PDS__ENQUEUE_BULK_UPDATES                = var.enable_pds_enqueue_bulk_updates ? "true" : "false"
-    MAVIS__PDS__RATE_LIMIT_PER_SECOND               = var.pds_rate_limit_per_second
-    GOOD_JOB_MAX_THREADS                            = 5
```
bogsi17, line 199: How are these values now defined for production?

terraform/app/variables.tf
```diff
@@ -214,59 +152,60 @@ locals {
...
-    {
-      name      = "DB_CREDENTIALS"
-      valueFrom = aws_rds_cluster.core.master_user_secret[0].secret_arn
-    },
-    {
-      name      = "RAILS_MASTER_KEY"
-      valueFrom = var.rails_master_key_path
-    }
-  ], local.parameter_store_config_list)
+  task_secrets = tomap({
```
bogsi17, line 191: Why is the `RAILS_MASTER_KEY` not added as task secret anymore?

terraform/app/variables.tf
TheOneFromNorway: The issue is that you cannot create a policy with empty permissions, so we cannot always create them if no permissions exist. I do agree that the construction is a bit complex so I am open to alternative solutions.

terraform/app/variables.tf
TheOneFromNorway: Ahh this fell out in refactoring, will add it back in

bogsi17 (approved):

PR #4451, opened 2025-09-01
Warning notifications for unrecognised contact details

murugapl:
Adds an extra safety net to check that the consent form submitter's contact details match those which are already on the child's record. If they don't match, the existing contact details will be alerted. This only happens for consent forms that are matched to patients automatically by Mavis and have not been reviewed by a nurse.

We still need to add the notify template ids for email and SMS, and corresponding personalisation.

https://nhsd-jira.digital.nhs.uk/browse/MAV-1717

### DIFF
--- a/app/controllers/concerns/consent_form_mailer_concern.rb
+++ b/app/controllers/concerns/consent_form_mailer_concern.rb
@@ -58,4 +58,37 @@ module ConsentFormMailerConcern
       SMSDeliveryJob.perform_later(:consent_confirmation_refused, **params)
     end
   end
+
+  def send_parental_contact_warning_if_needed(patient, consent_form, user)
+    return if patient.parents.empty? || user.present?
+
+    submitted_email = consent_form.parent_email
+    submitted_phone = consent_form.parent_phone
+
+    match_found =
+      patient.parents.any? do |parent|
+        email_match = parent.email.present? && submitted_email == parent.email
+        phone_match = parent.phone.present? && submitted_phone == parent.phone
+        email_match || phone_match
+      end
+
+    return if match_found
+
+    patient.parents.each do |parent|
+      if parent.email.present?
+        EmailDeliveryJob.perform_later(
+          :parental_consent_warning,
+          parent: parent,
+          patient: patient
+        )
+      end
+
+      next unless parent.phone.present? && parent.phone_receive_updates
+      SMSDeliveryJob.perform_later(
+        :parental_consent_warning,
+        parent: parent,
+        patient: patient
+      )
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/consent_form_mailer_concern.rb
```diff
@@ -58,4 +58,39 @@ def send_consent_form_confirmation_refused(consent_form, programmes: nil)
       SMSDeliveryJob.perform_later(:consent_confirmation_refused, **params)
     end
   end
+
+  def send_parental_contact_warning_if_needed(patient, consent_form, user)
+    return if patient.parents.empty? || user.present?
```
thomasleese, line 63: I think it's slightly confusing to have the check on the `user` being present here, especially as it's no then used in the method. I would suggest changing the way this method is called:

```rb
if current_user.nil?
  send_parental_contact_warning_if_needed(patient, self)
end
```

app/models/consent_form.rb
```diff
@@ -436,6 +437,8 @@ def parent_relationship
   end
 
   def match_with_patient!(patient, current_user:)
+    send_parental_contact_warning_if_needed(patient, self, current_user)
```
thomasleese, line 440: Actually I'm not sure if this is the right place to put this call. The concern is a controller concern, so already it feels wrong to include it in the model. I think it would be better to call this wherever `match_with_patient!` is called, and in fact I think that fits better with checking for the user as then we only need to call it from the automatic matching job.

config/initializers/govuk_notify.rb
```diff
@@ -43,6 +45,8 @@
   consent_confirmation_refused: "eb34f3ab-0c58-4e56-b6b1-2c179270dfc3",
   consent_school_reminder: "ee3d36b1-4682-4eb0-a74a-7e0f6c9d0598",
   consent_school_request: "c7bd8150-d09e-4607-817d-db75c9a6a966",
+  unknown_consent_contact_details_warning:
+    "1fd4620d-1c96-4af1-b047-ed13a90b0f44",
```
thomasleese, line 49: I think it would be good to keep this list in alphabetical order, and I wonder if we could name this with the `consent_` prefix so it goes with the others, something like:

```
consent_unknown_contact_details_warning
```

spec/features/parental_consent_inexact_auto_match_spec.rb
```diff
@@ -91,6 +94,7 @@ def when_a_parent_gives_consent_with_three_of_four_fields_matching_the_cohort
 
   def and_the_nurse_checks_the_consent_responses
     perform_enqueued_jobs
+    perform_enqueued_jobs
```
thomasleese, line 97:
```suggestion
    2.times { perform_enqueued_jobs }
```

Might be worth a comment just explaining why it has to be called twice.

thomasleese (approved):

PR #4460, opened 2025-09-01
Add PDS match rate validation for patient imports

murugapl:
Implement validation to ensure data quality during patient imports by checking that at least 70% of changesets successfully match with PDS records. This prevents processing of imports with poor data quality that could lead to duplicate or incorrect patient records. This check is skipped if less than 10 records are being imported, to avoid blocking small imports unnecessarily.

https://nhsd-jira.digital.nhs.uk/browse/MAV-1022

Screenshot 2025-09-01 at 15 42 12


app/models/patient_changeset.rb
```diff
@@ -54,6 +54,16 @@ class PatientChangeset < ApplicationRecord
...
 
+  scope :with_pds_match, -> { where.not(pds_nhs_number: nil) }
+
+  scope :with_pds_search_attempted,
+        -> do
+          where("pending_changes -> 'search_results' != '[]'::jsonb").where.not(
+            "jsonb_array_length(pending_changes -> 'search_results') = 1 AND 
+            (pending_changes -> 'search_results' -> 0 ->> 'result') = 'no_postcode'"
+          )
+        end
```
thomasleese, line 65: This query has the potential to be quite slow, as it requires reading from the JSON information, I think it's okay for now but one to keep an eye on if it becomes problematic.

app/models/patient_import.rb
```diff
@@ -17,6 +20,21 @@ def count_column(patient, parents, parent_relationships)
     end
   end
 
+  def validate_pds_match_rate!
+    return if valid_pds_match_rate? || changesets.count < CHANGESET_THRESHOLD
+
+    update_column(:status, :low_pds_match_rate)
```
thomasleese, line 26:
```suggestion
    update!(status: :low_pds_match_rate)
```

Should we use `update!` to ensure `updated_at` gets set?

thomasleese (approved):

thomasleese (approved):

PR #4462, opened 2025-09-01
PDS lookup timeline

murugapl:
In the event a user finds an NHS number has been added to the patient incorrectly, they might want to investigate where this NHS number has come from. We surface the PDS search history that has brought in the current NHS number on a patient.

https://nhsd-jira.digital.nhs.uk/browse/MAV-1815

Screenshot 2025-09-01 at 21 55 24
Screenshot 2025-09-01 at 17 43 54


app/components/app_child_summary_component.rb
```diff
@@ -32,6 +34,12 @@ def call
             href:,
             visually_hidden_text: "NHS number"
           )
+        elsif (href = pds_search_history_link)
+          row.with_action(
+            text: "PDS history",
+            href:,
+            visually_hidden_text: "PDS search history"
```
paulrobertlloyd, line 41: I _think_ this text gets appended to the visible text, meaning this link would be read aloud as ‘PDS history PDS search history’. Probably best to not show any visually hidden text for this link,

thomasleese (approved):

PR #4471, opened 2025-09-02
Adjust vaccination source backfilling logic

samcoy3:
This is a small change to the migration which adds source as a column to the VaccinationRecord table.

For historically uploaded vaccination records with the outcome `already_had` created before the migration, we now backfill the source value as `historical_upload` rather than `consent_refusal`. This is consistent with how those records are treated after the migration.

Jira-Issue: MAV-1603


db/migrate/20250829172505_add_source_to_vaccination_record.rb
```diff
@@ -4,11 +4,12 @@ class AddSourceToVaccinationRecord < ActiveRecord::Migration[8.0]
...
 
-    # Backfill values based on presence of session_id
     VaccinationRecord.update_all(source: "historical_upload")
-    VaccinationRecord.where(outcome: "already_had").update_all(
-      source: "consent_refusal"
-    )
+    # This handles records which were created when a parent refused consent because the patient was already vaccinated
+    VaccinationRecord.where(
+      outcome: "already_had",
+      immunisation_imports: []
```
thomasleese, line 11: I'm not sure this works as intended as Rails won't turn it in to the right query, I just tried in `test` and got no results:

```rb
manage-vaccinations(staging)> VaccinationRecord.where(immunisation_imports: []).count
=> 0
```

I think you'll probably want something that pre-loads and then filters using Ruby:

```rb
manage-vaccinations(staging)> VaccinationRecord.includes(:immunisation_imports).count { it.immunisation_imports.present? }
=> 900
```

samcoy3: Interesting. It worked for me locally, I tried migrating the DB with and without this change using an import which contained an `already_had`, and it worked as expected. I'll try to use the approach you suggest.

PR #4475, opened 2025-09-02
Ensure reliable NHS number retrieval from search results

murugapl:
Previously, `nhs_number` within `patient_changeset.search_results` was accessed using symbol keys (e.g., `sr[:nhs_number]`). However, when `search_results` (stored in a `jsonb` column) are deserialized from the database, they become standard Ruby hash objects with string keys, causing symbol access to fail.

Updated the `unique_nhs_numbers` method to use string keys (`sr["nhs_number"]`) for accessing the NHS number.

https://nhsd-jira.digital.nhs.uk/browse/MAV-1887

### DIFF
--- a/app/jobs/process_patient_changesets_job.rb
+++ b/app/jobs/process_patient_changesets_job.rb
@@ -51,7 +51,7 @@ class ProcessPatientChangesetsJob < ApplicationJob
   private
 
   def unique_nhs_numbers(patient_changeset)
-    patient_changeset.search_results.pluck(:nhs_number).compact.uniq
+    patient_changeset.search_results.pluck("nhs_number").compact.uniq
   end
 
   def get_unique_nhs_number(patient_changeset)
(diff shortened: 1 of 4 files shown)


app/jobs/process_patient_changesets_job.rb
```diff
@@ -51,7 +51,7 @@ def perform(patient_changeset, step_name = nil)
   private
 
   def unique_nhs_numbers(patient_changeset)
-    patient_changeset.search_results.pluck(:nhs_number).compact.uniq
+    patient_changeset.search_results.map { |sr| sr["nhs_number"] }.compact.uniq
```
thomasleese, line 54:
```suggestion
    patient_changeset.search_results.map { it["nhs_number"] }.compact.uniq
```

I don't think `sr` is any clearer than `it` 😄

thomasleese (approved):

PR #4478, opened 2025-09-02
Return user nav info in reporting token

aldavidson:
This will be used to populate the header on the reporting side.

It also adds a `GET /logout` page. This will not be linked to from Mavis, but from the Reporting app, which can't `POST /logout` as it can't generate CSRF tokens.

## After (Reporting app, not Mavis)

Screenshot 2025-09-04 at 13 06 11

## Logout page

Screenshot 2025-09-04 at 13 28 09

### DIFF
--- a/app/controllers/api/reporting/one_time_tokens_controller.rb
+++ b/app/controllers/api/reporting/one_time_tokens_controller.rb
@@ -20,7 +20,7 @@ class API::Reporting::OneTimeTokensController < API::Reporting::BaseController
       user_nav: {
         items: [
           { text: display_name, icon: true },
-          { href: "/logout", text: "Log out" }
+          { href: logout_path, text: "Log out" }
         ]
       }
     }
--- a/app/controllers/api/reporting/one_time_tokens_controller.rb
+++ b/app/controllers/api/reporting/one_time_tokens_controller.rb
@@ -9,7 +9,21 @@ class API::Reporting::OneTimeTokensController < API::Reporting::BaseController
   def authorize
     @token = ReportingAPI::OneTimeToken.find_by!(token: params[:code])
     @token.delete # <- Tokens are one-time use
-    json_data = { jwt: @token.to_jwt }
+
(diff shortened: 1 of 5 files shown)


app/views/layouts/_user_nav.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    </li>
+    <% if Settings.cis2.enabled %>
+    <li class="nhsuk-header__account-item">
+        <%= button_to "Change role", user_cis2_omniauth_authorize_path, class: "nhsuk-header__account-button", params: { change_role: true } %>
+    </li>
+    <% end %>
+    <li class="nhsuk-header__account-item">
+    <%= button_to "Log out", logout_path, class: "nhsuk-header__account-button", method: :delete %>
+    </li>
+</ul>
```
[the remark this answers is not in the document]
aldavidson: done, thanks

app/controllers/api/reporting/one_time_tokens_controller.rb
```diff
@@ -9,7 +9,21 @@ class API::Reporting::OneTimeTokensController < API::Reporting::BaseController
...
+    display_name = user.full_name
+    display_name +=
+      " (#{user.role_description})" if user.role_description.present?
+
+    json_data = {
+      jwt: @token.to_jwt,
+      user_nav: {
+        items: [
+          { text: display_name, icon: true },
+          { href: "/logout", text: "Log out" }
```
thomasleese, line 23:
```suggestion
          { href: logout_path, text: "Log out" }
```

app/controllers/logout_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class LogoutController < ApplicationController
```
thomasleese, line 3:
```suggestion
class Users::LogoutController < ApplicationController
```

Could we put this in the users module?

PR #4482, opened 2025-09-03
Ensure PGD supply is disabled when PSD is enabled

thomasleese:
This refactors various parts of the code related to delegation to ensure that when PSD is enabled, PGD supply is disabled. Specifically, this fixes a number of code paths which weren't checking if `pgd_supply_enabled?` was `true` (and instead assumed that it was always true) and also updates the filtering on the record tab to consider the case when PSD and national protocol is enabled.

I had to get rid of the `vaccine_method_for` method as it doesn't work in the case of PSDs, where the vaccine method can be different per patient.

I've also fixed a bug where HCAs are able to record injection vaccines if PGD supply is enabled and national protocol is not.

[Jira Issue - MAV-1893](https://nhsd-jira.digital.nhs.uk/browse/MAV-1893)
[Jira Issue - MAV-1895](https://nhsd-jira.digital.nhs.uk/browse/MAV-1895)


app/policies/vaccination_record_policy.rb
```diff
@@ -2,11 +2,31 @@
...
   def create?
-    user.is_nurse? || user.is_prescriber? ||
+    return true if user.is_nurse? || user.is_prescriber?
+    return false unless user.is_healthcare_assistant?
+
+    approved_vaccine_methods =
+      patient.approved_vaccine_methods(programme:, academic_year:)
+
+    # rubocop:disable Style/BlockDelimiters
+    if session.psd_enabled? &&
```
jhenderson, line 12: I know you mentioned that this needs further refactoring later but I was wondering if it's worth making this logic easier to parse. Something like:

```
can_create_with_psd?(approved_methods) ||
  can_create_with_national_protocol?(approved_methods) ||
  can_create_with_pgd_supply?(approved_methods)


def can_create_with_psd?(approved_methods)
  session.psd_enabled? && 
    approved_methods.any? { |method| 
      patient.has_patient_specific_direction?(programme:, academic_year:, vaccine_method: method) 
    }
end

def can_create_with_national_protocol?(approved_methods)
  session.national_protocol_enabled? && approved_methods.include?("injection")
end

def can_create_with_pgd_supply?(approved_methods)
  session.pgd_supply_enabled? && approved_methods.include?("nasal")
end
```

thomasleese: Yep, that makes sense, I've amended the commit with these suggested changes.

app/forms/vaccinate_form.rb
```diff
@@ -74,58 +74,35 @@ def delivery_site
   end
 
   def protocol
-    if current_user.is_nurse? || healthcare_assistant_using_pgd? ||
-         special_nasal_case?
-      "pgd"
-    elsif healthcare_assistant_using_psd?
+    if healthcare_assistant_using_psd?
```
jhenderson, line 77: The protocol calculation is easier to understand now. 👍

jhenderson (approved):
Looks good, especially all the factorings. 🚀

PR #4493, opened 2025-09-03
Verify that HCA's can't see triaged injected patients when national protocol is off

jhenderson:
This test verifies the existing behavior where Healthcare Assistants (HCAs) cannot see patients in the record vaccination tab when:
- National protocol is turned off
- Patients triaged for injection (consented for nasal, with injection as a second option)

This behavior is working as intended - HCAs can only record vaccinations under specific protocols, so hiding patients they cannot treat prevents confusion and maintains proper clinical governance.

### DIFF
--- a/spec/factories/patients.rb
+++ b/spec/factories/patients.rb
@@ -546,6 +546,44 @@ FactoryBot.define do
       end
     end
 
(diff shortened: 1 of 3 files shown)


thomasleese (approved):
Nice!

PR #4568, opened 2025-09-09
Enforce SSL encrypted DB connections

bogsi17:
* By setting the `rds.force_ssl` DB parameter, the DB will only accept SSL encrypted connections.
* On the client side, the `verify-full` parameter requires SSL connections and also enforces a certificate verification. For that reason, the public root CA certificate needs to exist on the app

Jira-Issue: MAV-1600


config/database.yml
```diff
@@ -12,6 +12,10 @@ test:
 staging:
   <<: *default
   database: manage_vaccinations_staging
+  sslmode: verify-full
```
TheOneFromNorway, line 15: Should we just use `require` instead of `verify-full`, this would just trust the certificate instead of validating it but means we would not need to add/store/update the certificate which is still a big improvement

PR #4571, opened 2025-09-09
Send pseudo-postcode to Imms API when patient postcode nil

mikejamesthompson:
Postcode is a mandatory field when creating a new record in the Imms FHIR API. We need to handle the case where a patient in Mavis has a `nil` postcode, by sending a pseudo-postcode, similar to what we already do for location codes.

The postcode value was chosen from the ones described here: https://digital.nhs.uk/developer/api-catalogue/personal-demographics-service-fhir/get-and-update-contact-details-in-pds

Fixes [MAV-1942](https://nhsd-jira.digital.nhs.uk/browse/MAV-1942).

### DIFF
--- a/app/lib/fhir_mapper/patient.rb
+++ b/app/lib/fhir_mapper/patient.rb
@@ -22,7 +22,7 @@ module FHIRMapper
         name: [FHIR::HumanName.new(family: family_name, given: given_name)],
         birthDate: date_of_birth&.strftime("%Y-%m-%d"),
         gender: gender_fhir_value,
-        address: [FHIR::Address.new(postalCode: address_postcode)]
+        address: [FHIR::Address.new(postalCode: address_postcode || "ZZ99 3CZ")]
       )
     end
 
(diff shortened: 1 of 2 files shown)


app/lib/fhir_mapper/patient.rb
```diff
@@ -22,7 +22,7 @@ def fhir_record(reference_id: nil)
         name: [FHIR::HumanName.new(family: family_name, given: given_name)],
         birthDate: date_of_birth&.strftime("%Y-%m-%d"),
         gender: gender_fhir_value,
-        address: [FHIR::Address.new(postalCode: address_postcode)]
+        address: [FHIR::Address.new(postalCode: address_postcode || "ZZ99 3CZ")]
```
thomasleese, line 25: I think it would be worth extracting the `"ZZ99 3CZ"` literal to a constant with a name explaining what it is.

thomasleese (approved):

PR #4586, opened 2025-09-10
Update reporting app configuration

tvararu:
Instead of storing the CLIENT_ID and SECRET in the secret manager in AWS, we can store them in the encrypted secrets in the repository.

This also sets up defaults for development that align with the new defaults in the reporting app.


tvararu:
@thomasleese I had to update the `secret` in the test credentials to be something other than `""` to fix the tests, done in the last commit:

```sh
$ rails credentials:show --environment=test | tail -n 3
reporting_api:
  client_id: "testing_client_id"
  secret: "testing_client_secret"
```

PR #4634, opened 2025-09-16
Create clinic sessions per programme

thomasleese:
This builds on top of the refactoring work done in https://github.com/nhsuk/manage-vaccinations-in-schools/pull/4526 which unlocked the ability for clinic sessions to work in a similar way to school sessions where there exists a session per group of programmes (HPV, Flu, and Doubles).

This is more of a proof of concept for now, we don't know operationally whether teams are more likely to use clinics per programme, or clinics for all the programmes ("make every contact count").


thomasleese:
Closing this for now, but we can resurrect it if we need to in the future.

PR #4650, opened 2025-09-17
Disallow offline spreadsheets from editing records received from the Imms API

alistair-white-horne-tng:
Vaccination records which are received from the Imms API are considered
immutable in Mavis, because we are not the vaccination record's primary
source. As such, users should not be allowed to edit the records by any
means, including an offline spreadsheet upload.

We (Thomas, Mike, Alistair) considered exposing this as a validation
error on the import, but that approach was discarded. This is because we
would have needed to duplicate a lot of the import logic to decide
whether it was appropriate to throw a validation error. This led to
circular dependencies in the validation logic.

Instead this approach was taken, where an adjusted import issue page is
shown, forcing the user to discard the incoming changes.

Off the back of this change, another issue needed to be resolved;
unedited vaccination records from the API were being flagged for review.
This required a bit of a rewrite of the import logic, as well as the offline
spreadsheet logic.

[MAV-1781](https://nhsd-jira.digital.nhs.uk/browse/MAV-1781)

image

### DIFF
--- a/app/forms/import_duplicate_form.rb
+++ b/app/forms/import_duplicate_form.rb
@@ -35,10 +35,23 @@ class ImportDuplicateForm
   end
 
   def apply_changes_options
-    can_keep_both? ? %w[apply discard keep_both] : %w[apply discard]
+    if can_apply?
+      can_keep_both? ? %w[apply discard keep_both] : %w[apply discard]
+    else
+      %w[discard]
+    end
+  end
+
+  def can_apply?
+    !(
+      object.is_a?(VaccinationRecord) &&
+        object.sourced_from_nhs_immunisations_api?
+    )
   end
 
   def apply_pending_changes!
+    return unless can_apply?
+
     object.patient.apply_pending_changes! if object.respond_to?(:patient)
 
     object.apply_pending_changes!
@@ -51,7 +64,7 @@ class ImportDuplicateForm
   end
 
   def keep_both_changes!
-    object.apply_pending_changes_to_new_record! if can_keep_both?
+    object.apply_pending_changes_to_new_record! if can_keep_both? && can_apply?
   end
 
   def reset_count!
--- a/app/models/immunisation_import_row.rb
+++ b/app/models/immunisation_import_row.rb
@@ -105,8 +105,6 @@ class ImmunisationImportRow
     attributes = {
       dose_sequence: dose_sequence_value,
       full_dose: true,
-      location:,
-      location_name:,
       outcome:,
       patient_id: patient.id,
       performed_at:,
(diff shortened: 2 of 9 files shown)


app/forms/import_duplicate_form.rb
```diff
@@ -36,6 +44,11 @@ def apply_changes_options
     can_keep_both? ? %w[apply discard keep_both] : %w[apply discard]
   end
 
+  def imms_record?
```
thomasleese, line 47:
```suggestion
  def can_apply?
```

This matches the naming convention of `can_keep_both?` and keeps the method name generic (this class doesn’t need to know about the Imms API specifically), making it easier to expand in the future if we needed it.

Should we also update `apply_changes_options` to not return `apply` if `can_apply?` is `false`? That way the view could be more generic and in the case of there being only one option (in this case `discard`) display the content specific to the one option.

thomasleese (commented):
The functionality looks good, I think we should try and make it more generic, so fewer classes need to know about the Immunisations API and the logic around it. I think if we can keep that limited to `ImportDuplicateForm`, and then the views and controllers use methods from the form that would be ideal.

app/forms/import_duplicate_form.rb
alistair-white-horne-tng: Good idea

alistair-white-horne-tng:
@thomasleese I've done as you described in [your comment](https://github.com/nhsuk/manage-vaccinations-in-schools/pull/4650#pullrequestreview-3235027848). I don't think it's reasonable to keep this logic out of the `ImmunisationImportRow` and the offline spreadsheet generation code. What do you think?

alistair-white-horne-tng:
@misaka I just changed the logic to make it much more generic, like we discussed this morning.

cc @thomasleese hopefully this makes you happy

app/models/immunisation_import_row.rb
```diff
@@ -380,6 +386,13 @@ def programmes_by_name
 
   def offline_recording? = session_id.present?
 
+  def imms_api_record?
+    uuid.present? &&
+      VaccinationRecord.find_by!(
+        uuid: uuid.to_s
+      ).sourced_from_nhs_immunisations_api?
```
thomasleese, line 393:
```suggestion
      VaccinationRecord.sourced_from_nhs_immunisations_api.exists?(
        uuid: uuid.to_s
      )
```

I think we should try and stick with `exists?` in this class to improve performance. `EXISTS` will be much faster as Rails won't need to select all the columns from the table and create a `VaccinationRecord` instance for it, instead we'll get a boolean straight from the database. Especially since this method is now going to be called for every row in the file.

thomasleese (approved):

PR #4690, opened 2025-09-22
Send notification on vaccs discovery

alistair-white-horne-tng:
If a vaccination record is created for a patient whose parent has already given consent, then the parent should be notified that their child will no longer be vaccinated.

## Screenshots
image

image

## Post-release tasks

- Change the name of the templates to be prefixed with `[live]`:
  - [Email](https://www.notifications.service.gov.uk/services/629645d8-feb9-4fc1-93f4-759c93ac0e21/templates/e37fe0a2-7584-4c25-983a-8f5a11c818a1)
  - [SMS](https://www.notifications.service.gov.uk/services/629645d8-feb9-4fc1-93f4-759c93ac0e21/templates/fab1e355-bde1-47d5-835c-103bfd232b93)


app/controllers/concerns/vaccination_mailer_concern.rb
```diff
@@ -48,6 +48,60 @@ def send_vaccination_deletion(vaccination_record)
...
+    consents =
+      ConsentGrouper.call(
+        consents,
+        programme_id: vaccination_record.programme_id,
+        academic_year: vaccination_record.academic_year
+      )
+
+    consents = consents.select(&:response_given?).reject(&:withdrawn_at)
+
+    parents = consents.map(&:parent).uniq
```
thomasleese, line 82: I think we should try and avoid the duplicate logic between here and `parents_for_vaccination_mailer`, and it might be worth refactoring `parents_for_vaccination_mailer` in to two methods, one which takes the vaccination record and another which gets the parents from a list of consents.

app/lib/govuk_notify_personalisation.rb
```diff
@@ -314,7 +314,7 @@ def today_or_date_of_vaccination
     if vaccination_record.performed_at.today?
       "today"
     else
-      vaccination_record.performed_at.to_date.to_fs(:long)
+      "on #{vaccination_record.performed_at.to_date.to_fs(:long)}"
```
thomasleese, line 317: Could this change be extracted in to a separate commit?

app/models/immunisation_import.rb
```diff
@@ -32,6 +32,7 @@
 #
 class ImmunisationImport < ApplicationRecord
   include CSVImportable
+  include VaccinationMailerConcern
```
thomasleese, line 35: It doesn't seem right to include a controller concern in to a model. If we need to share this logic between models and controllers then a concern might not be the right place for it and instead we could have a service class.

db/migrate/20250923072038_add_vaccination_record_discovered_sent_at_to_consents.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddVaccinationRecordDiscoveredSentAtToConsents < ActiveRecord::Migration[
+  8.0
+]
+  def change
+    add_column :consents,
+               :vaccination_record_discovered_sent_at,
```
thomasleese, line 8: Just wondering about this naming, because it's not just that a vaccination record was discovered, it's specifically that one was found which means the patient is actually already vaccinated. I'm wondering if a name like `patient_already_vaccinated_notification_sent_at` is clearer?

misaka (approved):

PR #4785, opened 2025-10-02
Extract PDS cascading search logic into separate job

murugapl:
Move PDS cascading search logic from ProcessPatientChangesetsJob into a new PDSCascadingSearchJob to enable its reuse across different contexts.

Job flow:  ProcessImportJob -> PDSCascadingSearchJob (self-chaining, rate-limited) -> ProcessPatientChangesetJob (async) -> CommitPatientChangesetsJob

Jira: [MAV-2170](https://nhsd-jira.digital.nhs.uk/browse/MAV-2170)


spec/jobs/pds_cascading_search_job_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+
+      it "skips wildcard name steps and completes search" do
+        described_class.perform_now(
+          searchable_id: patient_changeset.id,
+          searchable_type: "PatientChangeset"
+        )
+        
+        while enqueued_jobs.any? { |job| job["job_class"] == described_class.name }
+          perform_enqueued_jobs(only: described_class)
+        end
```
thomasleese, line 143: Since we have code like this in a few places now I'm wondering if it's worth extracting it in to a common method that we can use in various parts of the tests.

thomasleese (approved):

PR #4820, opened 2025-10-07
Grafana alerting

bogsi17:
This PR contains the infrastructure to manage alerts in Grafana

- Configure folders for alerts and a notification point
- Send Slack notifications via a webhook that's stored in the repository secrets
- Script to convert alert export to valid terraform
- Outline a process to recreate alerts from the terraform config


terraform/monitoring/grafana/main.tf
```diff
@@ -29,3 +28,70 @@ resource "grafana_data_source" "cloudwatch" {
   })
   uid = "cloudwatch"
 }
+
+resource "grafana_data_source" "postgresql" {
```
TheOneFromNorway, line 32: This is already part of the introduce read only user correct?

.github/workflows/deploy-monitoring.yml
```diff
@@ -147,5 +147,5 @@ jobs:
           sudo apt-get install -y jq uuid-runtime
       - name: Deploy Grafana using tf_grafana.sh
         run: |
-          ./tf_grafana.sh "$environment" plan --plan-file ${{ runner.temp }}/out
+          ./tf_grafana.sh "$environment" plan --plan-file ${{ runner.temp }}/out --slack-webhook-url ${{ inputs.environment == 'production' && secrets.GRAFANA_ALERTS_SLACK_WEBHOOK || secrets.GRAFANA_TEST_ALERTS_SLACK_WEBHOOK }}
```
TheOneFromNorway, line 150: Should we simply do this with environment variables instead of functional arguments?

terraform/monitoring/grafana/main.tf
```diff
@@ -29,3 +28,70 @@ resource "grafana_data_source" "cloudwatch" {
...
+resource "grafana_folder" "load_balancer" {
+  title = "Load Balancer"
+  uid   = "load-balancer-folder"
+}
+
+resource "grafana_contact_point" "slack" {
+  name = "Slack"
+
+  slack {
+    url = var.slack_webhook_url
```
TheOneFromNorway, line 79: We could add ignore changes here since the url will be pretty stactic and then it is not needed to be passed on every apply. If we needed to change it we could always manually taint the resource

terraform/monitoring/grafana/main.tf
```diff
@@ -29,3 +28,70 @@ resource "grafana_data_source" "cloudwatch" {
...
+    ignore_changes = [
+      secure_json_data_encoded,
+      url,
+    ]
+  }
+}
+
+resource "grafana_folder" "ecs" {
+  title = "ECS"
+  uid   = "ecs-folder"
```
TheOneFromNorway, line 62: I like us adding all these folders, do we also want to update the dashboards to live in these new folders?

terraform/monitoring/grafana/main.tf
bogsi17: Right, I think I branched off from that branch initially that's why it shows up here

.github/workflows/deploy-monitoring.yml
bogsi17: Not sure what you mean - the values are secrets, so we have to make sure that they don't show up anywhere. The alternative would be to store them in a secretsmanager secret. But I would prefer if you don't need AWS access in case you want to change the slack channel for the alerts

terraform/monitoring/grafana/main.tf
bogsi17: Yes good idea

.github/workflows/deploy-monitoring.yml
TheOneFromNorway: I would hope that github has some protection against populating secrets as plaintext. AFAIK they should be masked, so we could se the variable `TF_VAR_slack_webhook_url=...` in the workflow which terraform will pick up on plan.

PR #4822, opened 2025-10-07
Active Load balancer target groups

bogsi17:
This change is a follow-up to #4781. It is needed because a deployment happened for the 5.0.0 release, which switched the active target groups again.


TheOneFromNorway (approved):
Double checked and it is now correct

PR #4827, opened 2025-10-08
Redesign global child page to support global triage

thomasleese:
This updates the design of the children page to use the patient search result cards and more filtering to allow the page to meet the need of performing triage outside of a particular session.

[Jira Issue - MAV-2143](https://nhsd-jira.digital.nhs.uk/browse/MAV-2143)


paulrobertlloyd:
@mikejamesthompson I think the statuses as shown on the cards work correctly; they only show when the respective filters are active.

I’ll have another play to check that this is right…

mikejamesthompson:
> Let’s add the year group filters, but I think the rest is working correctly.
> 
> One slight oddity is that if you select a consent or triage filter without selecting a programme, you get no results. I think this might work better if it returned all children with those statuses across all programmes.

Sounds good, thanks.

